PR #96, opened 2014-08-19
Add cbufp_map_at() interface, fix and refactor cbufp, vas_mgr, and valloc components

gedare:
* Fixes a bug in the shared ring buffer size calculation.
* Refactors some of the memory-mapping code used in cbufp.
* Relocates the typedefs for cbuf_t and cbufp_t to the interface headers instead of the include/cbuf.h. This is necessary so that cbufp_t is visible within cbufp.c.
* Makes a small whitespace fix in cbuf_c.c that I encountered along the way.
* Refactors the vas_mgr_expand in preparation of sharing some of the logic with a new function.
* Refactors valloc_free to simplify the logic.
* Moves the calculation of ext_size in valloc_alloc out of a loop.
* Adds the cbufp_map_at and cbufp_unmap_at interface functions to support mapping a cbuf directly from one component into another's VAS.
* Adds valloc_alloc_at() to support allocating a region of virtual memory within a component at a specific virtual address.
* Adds vas_mgr_take() to support claiming a contiguous region of the VAS at a specific virtual address.


gedare:
Did I miss something? This wasn't merged but it was closed.

gparmer:
I messed up not merging back when you originally sent it.  I wasn't sure at
this point if it was too stale.  Can you resubmit the pull request if it is
not stale?

g

gparmer:
Sorry for the confusion on this.  Pulled now.  Thanks for this!

PR #109, opened 2015-01-26
cbuf client side

ryuxin:
Have passed some micro tests. But I do not want to merge it.


gedare:
Can you please fix merge conflicts, and tell us why you don't want it merged yet?

gedare:
Also, this single-commit is way too big. You should provide separate commits and commit messages for "atomic" (in the transactional manner) commits. Smaller commits are easier to review, and atomic commits that can be applied separately, compile correctly, and "work" are nice-to-have, and when debugging they are need-to-have to be able to use git-bisect correctly.

PR #110, opened 2015-01-30
cbuf client side code, split into separate small commits  

ryuxin:


gedare:
High-level: Thanks for making the changes to create smaller commits. It is MUCH easier to read the commits to figure out what is changing. Please close your other request (#109), and I have made some comments on your code. If you have any questions feel free to ask!

src/components/implementation/cbuf_mid/mid_level/cbuf_mid.c
```diff
@@ -38,8 +38,8 @@ void cbuf_call(char c)
 	for (i = 0; i < iter ; i++){
 		int tmem;
 		cbt[i] = cbuf_null();
-		mt[i] = cbuf_alloc(SZ, &cbt[i]);
-		cbuf_unpack(cbt[i], &id, &tmem);
+		mt[i] = cbuf_alloc(SZ, &cbt[i], 1);
```
gparmer: 1. The third argument shouldn't be a boolean.  It should be a flag, so that we can expand it later if needs be.

2. We probably don't want the third argument to be part of the common-case API.  TMem will mostly only be used by the IDL, so we should just provide another function that enables us to pass in the flags.  Perhaps something like `void *cbuf_alloc_ext(unsigned int sz, cbuf_t *cbid, unsigned int flags)`.  Most code that explicitly calls cbuf functions will not be using the flags (unless one of them in the future becomes essential), so we might as well optimize for that common case, and have `cbuf_alloc(sz, *cb)` be the function prototype.  This will obviously be a simple wrapper around the `_ext` function.

What do you guys thing?  Is it worth optimizing away the flags as we will essentially never use them in common code?

src/components/implementation/cbufp/default/cbufp.c
```diff
@@ -8,11 +8,9 @@
  */
```
gparmer: Shouldn't we change `implementation/cbufp/default/cbufp.c` to `implementation/cbuf/default/cbuf.c` now as we're getting rid of the distinction between cbuf/cbufp?

src/components/implementation/cbufp/default/cbufp.c
```diff
@@ -381,12 +378,14 @@ cbufp_create(spdid_t spdid, int size, long cbid)
...
 	memset(meta, 0, sizeof(struct cbuf_meta));
-	meta->nfo.c.flags |= CBUFM_TOUCHED | 
-		             CBUFM_OWNER  | CBUFM_WRITABLE;
-	meta->nfo.c.ptr    = cbi->owner.addr >> PAGE_ORDER;
-	meta->sz           = cbi->size >> PAGE_ORDER;
-	if (meta->nfo.c.refcnt == CBUFP_REFCNT_MAX) assert(0);
-	meta->nfo.c.refcnt++;
+	CBUF_SET_OWNER(meta);
+	CBUF_SET_WRITABLE(meta);
+	CBUF_SET_TOUCHED(meta);
```
gparmer: This is not how flags work.  You're relying on the compiler to do a lot of optimizations to not make this horribly slow.  It is reasonable that it will in many cases, but I'm not sure why we care to make it work hard in this case.

More importantly, everyone knows how to use bit fields for flags, and how to modify them.  So why break that convention and intuition there?  Better might be:

`CBUF_FLAGS_ADD(CBUFM_TOUCHED | CBUFM_OWNER  | CBUFM_WRITABLE);`

and, if you wish, a `CBUF_FLAGS_REM(...)` corresponding call.

The `CBUFM_PTR_SET(...)` is fine, but you might as well add the same for size and cbid.

Please remember the naming conventions.  You always try for `noun_verb`, not `verb_noun`.  So when you have `SET_PTR`, you should really have `PTR_SET` (same for REFCNT, etc....).

src/components/implementation/cbufp/default/cbufp.c
```diff
@@ -618,14 +616,14 @@ cbufp_retrieve(spdid_t spdid, int cbid, int size)
 	assert(page);
 	if (cbufp_map(spdid, dest, page, size, MAPPING_READ))
 		valloc_free(cos_spd_id(), spdid, (void *)dest, 1);
-
-	meta->nfo.c.flags |= CBUFM_TOUCHED;
-	meta->nfo.c.ptr    = map->addr >> PAGE_ORDER;
-	meta->sz           = cbi->size >> PAGE_ORDER;
-	ret                = 0;
+	memset(meta, 0, sizeof(struct cbuf_meta));
```
gparmer: Note that this will set the contents of meta to `0` starting from the lower bytes, to the higher bytes.  Please confirm that there is no race between this and the client (i.e. the client sees a 0 at one lower-byte field, and does something that gets erased by a concurrent `memset`.

src/components/implementation/tests/micro_mbox_client/mbox_client.c
```diff
@@ -61,13 +61,13 @@ void cos_init(void *arg)
...
-		d = cbufp_alloc(sz, &cb1);
+		d = cbuf_alloc(sz, &cb1, 0);
 		if (!d) goto done;
-		cbufp_send(cb1);
+		cbuf_send(cb1);
 		rdtscll(end);
 		((u64_t *)d)[0] = end;
 		ret = twritep(cos_spd_id(), serv, cb1, sz);
-		cbufp_deref(cb1); 
+		cbuf_free(cb1);
```
gparmer: You got rid of `cbuf_deref` here (replaced with `cbuf_free`), but kept `cbuf_send_deref`.  I suggest you change `cbuf_send_deref` into `cbuf_send_free`.

src/components/implementation/tests/micro_mbox_server/mbox_server.c
```diff
@@ -68,8 +68,8 @@ void cos_init(void *arg)
 			if ((int)cb1<0) evt_wait(cos_spd_id(), evt2);
 			else            break;
 		}
-		buf = cbufp2buf(cb1,sz);
-		cbufp_deref(cb1);
+		buf = cbuf2buf(cb1,sz);
+		cbuf_free(cb1);
```
gparmer: I'm interested in knowing why in the `mbox`, you have this pattern.  Why is the cb1 not used at all?

src/components/implementation/tests/micro_ping_cbuf/micro_ping.c
```diff
@@ -37,16 +37,15 @@ cbuf_tests(void)
         /* CACHING */
 	printc("\n<<< WARM UP CBUF CACHE.......");
 	for (i = 0; i < NCBUF ; i++){
-		prev_sz += 4;
-		prev_sz &= PAGE_SIZE-1;
+		prev_sz = 2*PAGE_SIZE;//4;
+		//prev_sz &= PAGE_SIZE-1;
```
gparmer: If you keep all of this, please clean it up.

src/components/implementation/tests/micro_pong_cbuf/micro_pong.c
```diff
@@ -11,7 +11,7 @@
 #define printv(fmt,...) 
 #endif
 
-#define ITER 2000
+#define ITER 1/*2000*/
```
gparmer: This is certainly not good for testing....

Please clean this up.

src/components/include/cbuf.h
```diff
@@ -18,11 +18,12 @@
...
 #include <cos_component.h>
+#include <util.h>
 #include <cos_debug.h>
-#include <cbuf_c.h>
+#include <cbuf_meta.h>
 #include <cbufp.h>
-#define cbid_to_meta_idx(cid) ((cid) << 1)
-#define meta_to_cbid_idx(mid) ((mid) >> 1)
+#define cbid_to_meta_idx(cid) ((cid) << 2)
+#define meta_to_cbid_idx(mid) ((mid) >> 2)
```
gparmer: BTW, the correct fix for this, further down the line, is to use the `ertrie`s in the `ppos` branch.  For now, this is fine, but we'll eventually want to replace the meta tree with that implementation (i.e. it is actually maintained and well-used).

src/components/include/cbuf.h
```diff
@@ -219,40 +214,35 @@ struct cbuf_agg {
...
-extern int  __cbuf_2buf_miss(int cbid, int len, int tmem);
-extern void __cbuf_desc_free(struct cbuf_alloc_desc *d);
-extern cvect_t meta_cbuf, meta_cbufp;
+extern struct cbuf_meta * __cbuf_alloc_slow(int size, int *len);
+extern int  __cbuf_2buf_miss(int cbid, int len);
+extern cvect_t meta_cbuf;
 
 static inline struct cbuf_meta *
-cbuf_vect_lookup_addr(long idx, int tmem)
+cbuf_vect_lookup_addr(long idx)
```
gparmer: Would it be possible to hide the `cbid_to_meta_idx` calls within this function so that the index translation is only done here and not in all of the calling functions?

src/components/include/cbuf.h
```diff
@@ -262,63 +252,38 @@ cbuf_vect_lookup_addr(long idx, int tmem)
...
-	}
-	/* if (unlikely(cm->sz && (((int)cm->sz)<<PAGE_ORDER) < len)) goto done; */
-	/* ci_new.v       = ci.v; */
-	/* ci_new.c.flags = ci.c.flags | CBUFM_RECVED | CBUFM_IN_USE; */
-	/* if (unlikely(!cos_cas((unsigned long *)&cm->nfo.v,  */
-	/* 		      (unsigned long)   ci.v,  */
-	/* 		      (unsigned long)   ci_new.v))) goto again; */
-	ret = ((void*)(cm->nfo.c.ptr << PAGE_ORDER));
+	assert(cm->cbid.cbid == id);
+	if (unlikely(CBUF_OWNER(cm))) goto done;
```
gparmer: For consistency, shouldn't this be `CBUFM_OWNER(...)`?

src/components/include/cbuf.h
```diff
@@ -262,63 +252,38 @@ cbuf_vect_lookup_addr(long idx, int tmem)
...
+	if (unlikely((len >> PAGE_ORDER) > cm->sz)) goto done;
+	assert(CBUFM_GET_REFCNT(cm) < CBUF_REFCNT_MAX);
+	CBUFM_INC_REFCNT(cm);
+	assert(cm->owner_nfo.c.nrecvd < CBUF_SENDRECV_MAX);
+	cm->owner_nfo.c.nrecvd++;      //???
```
gparmer: If you have a question or comment, please actually write it down.  This `//???` is cryptic.

src/components/include/cbuf.h
```diff
@@ -332,66 +297,37 @@ cbufp2buf(cbufp_t cb, int len) { return __cbuf2buf((cbuf_t)cb, len, 0); }
...
+	assert(CBUFM_GET_REFCNT(cm));
+	assert(cm->owner_nfo.c.nsent < CBUF_SENDRECV_MAX);
 
 	cm->owner_nfo.c.nsent++;
-	if (free) cm->nfo.c.refcnt--;
+	if (free) CBUFM_DEC_REFCNT(cm);
 	/* really should deal with this case correctly */
-	assert(!(cm->nfo.c.flags & CBUFM_RELINQ)); 
-	CBUF_RELEASE();
+	assert(!CBUF_RELINQ(cm));     //???
```
gparmer: Again: `//???`...

src/components/include/cbuf.h
```diff
@@ -332,66 +297,37 @@ cbufp2buf(cbufp_t cb, int len) { return __cbuf2buf((cbuf_t)cb, len, 0); }
...
+{ __cbuf_send(cb, 0); }
 
-extern cvect_t alloc_descs; 
-struct cbuf_alloc_desc {
-	int cbid, length, tmem;
-	void *addr;
-	struct cbuf_meta *meta;
-	struct cbuf_alloc_desc *next, *prev, *flhead; /* freelist */
+struct cbuf_freelist {
+	struct cbuf_meta freelist_head[64];    //???
```
gparmer: `64` should be a macro.  Is 64 chosen for any reason?  Are these power of 2 freelists?  If they are, we're maxing out at a ridiculously large number of pages (2^64 pages) which is completely unnecessary.

src/components/include/cbuf.h
```diff
@@ -332,66 +297,37 @@ cbufp2buf(cbufp_t cb, int len) { return __cbuf2buf((cbuf_t)cb, len, 0); }
...
-{
-	assert(d && m && d->addr);
-	/* we don't want the manager changing this under us */
-	assert(m->nfo.c.refcnt);
-	return (unlikely((unsigned long)d->addr >> PAGE_SHIFT != m->nfo.c.ptr ||
-			 d->meta != m /*|| length*/));
-}
-
+PERCPU_DECL(struct cbuf_freelist, cbuf_alloc_freelists);
+PERCPU_EXTERN(cbuf_alloc_freelists);
```
gparmer: Did you check with Qi to make sure you used the proper macros here?

src/components/include/cbuf.h
```diff
@@ -405,241 +341,131 @@ __cbuf_alloc_meta_inconsistent(struct cbuf_alloc_desc *d, struct cbuf_meta *m)
...
-
+	struct cbuf_meta *m;
 	assert(pow2(size) && size >= PAGE_SIZE);
 	assert(order >= 0 && order < WORD_SIZE-PAGE_ORDER);
-	d = &cbufp_alloc_freelists[order];
-	assert(d->length == size);
-
-	return d;
+	m = &(fl->freelist_head[order]);
+	if (!CBUFM_GET_NEXT(m)) {
```
gparmer: Again, remember `noun_verb`.

src/components/include/cbuf.h
```diff
@@ -405,241 +341,131 @@ __cbuf_alloc_meta_inconsistent(struct cbuf_alloc_desc *d, struct cbuf_meta *m)
...
+		CBUFM_SET_NEXT(m, m);
```
gparmer: Shouldn't this be atomic?

src/components/implementation/cbuf_mid/mid_level/cbuf_mid.c
gedare: I had the same thought about the API. Hiding the flags makes good sense to me so we don't have another API change when we put tmem's into generated code only.

src/components/include/cbuf.h
```diff
@@ -405,241 +341,131 @@ __cbuf_alloc_meta_inconsistent(struct cbuf_alloc_desc *d, struct cbuf_meta *m)
...
+	sz = nlepow2(round_up_to_page(sz));
+	fl = __cbuf_freelist_get(sz);
 again:
-	d = FIRST_LIST(fl, next, prev);
-	if (unlikely(EMPTY_LIST(d, next, prev))) {
-		d    = __cbuf_alloc_slow(sz, &len, tmem);
-		assert(d);
-		ret  = d->addr;
-		cbid = d->cbid;
+	target = (unsigned long long *)(&fl->next_flag);
```
gparmer: Why are these `unsigned long long`s?  If 64 bits is required, then we might want to just use `u64_t`, but I am not quite following why 64 bits are needed.

src/components/include/cbuf.h
```diff
@@ -405,241 +341,131 @@ __cbuf_alloc_meta_inconsistent(struct cbuf_alloc_desc *d, struct cbuf_meta *m)
...
+	old    = (unsigned long long *)(&old_head.next_flag);
+	update = (unsigned long long *)(&new_head.next_flag);
+	do {
+		*old   = *target;
```
gparmer: line up all of the `=`s

src/components/include/cbuf.h
```diff
@@ -405,241 +341,131 @@ __cbuf_alloc_meta_inconsistent(struct cbuf_alloc_desc *d, struct cbuf_meta *m)
...
+		cm		   = CBUFM_GET_NEXT(fl);
+		new_head.next_flag = cm->next_flag;
+		new_head.cbid.tag  = fl->cbid.tag+1;
```
gparmer: The structure nesting is causing confusion here.  `tag` within `cbid` is strange.  Is this fixable?

src/components/interface/cbufp/cbuf_meta.h
```diff
@@ -0,0 +1,139 @@
...
+};
+union cbid_tag {
+	u32_t cbid;
+	unsigned int tag;
+};
+struct cbuf_meta {
+	unsigned int nfo;
+	u16_t sz;			/* # of pages || 0 == TMEM */
+	union cbufm_ownership owner_nfo;
+	struct cbuf_meta *next_flag;    /*single circular linked list*/
```
gparmer: Why is this a circular list?

src/components/include/cbuf.h
```diff
@@ -405,241 +341,131 @@ __cbuf_alloc_meta_inconsistent(struct cbuf_alloc_desc *d, struct cbuf_meta *m)
...
+	} while(unlikely(!cos_dcas(target, *old, *update)));
```
gparmer: Why pass in in pointers to the values here as opposed to just passing in the old and update values by value?  I'm not sure it matters, but we'll rely a little less on the compiler to do the right thing if we just pass in values.

src/components/include/cbuf.h
```diff
@@ -405,241 +341,131 @@ __cbuf_alloc_meta_inconsistent(struct cbuf_alloc_desc *d, struct cbuf_meta *m)
...
+	if (unlikely(cm == CBUFM_GET_NEXT(cm))) {
+		cm   = __cbuf_alloc_slow(sz, &len);
+		assert(cm);
 		goto done;
 		/*
 		 * TODO: check if this cbuf has been taken by another
```
gparmer: I'm not sure this comment is relevant anymore.  You?

src/components/include/cbuf.h
```diff
@@ -405,241 +341,131 @@ __cbuf_alloc_meta_inconsistent(struct cbuf_alloc_desc *d, struct cbuf_meta *m)
...
+	unsigned int p, n;
+	p = cm->nfo;
+	n = p+CBUFM_INC_UNIT;
+	if (unlikely(!cos_cas((unsigned long *)&cm->nfo, p, n))) {
+		CBUFM_SET_NEXT_NULL(cm);
+		goto again;
+	}
+	CBUFM_SET_NEXT_NULL(cm);
+	cm->owner_nfo.c.nsent = cm->owner_nfo.c.nrecvd = 0;
+	/*if (unlikely(tmem)) {
```
gparmer: Remove the code here if it is commented out?

src/components/include/cbuf.h
```diff
@@ -405,241 +341,131 @@ __cbuf_alloc_meta_inconsistent(struct cbuf_alloc_desc *d, struct cbuf_meta *m)
...
-		 * cbuf_slab_free to see the surprising fact that we
-		 * do _not_ deallocate the slab descriptor there to
-		 * reinforce that point.
-		 */
-		__cbuf_desc_free(d);
-		if (likely(!already_used)) {
-			cm->nfo.c.flags &= ~CBUFM_TOUCHED;
-			cm->nfo.c.refcnt--;
-		}
+	inconsistent = CBUF_INCONSISENT(cm);
```
gparmer: We need to review the inconsistency rules, and you need to document them in a comment somewhere.  I can't review that here.

src/components/include/cbuf.h
```diff
@@ -405,241 +341,131 @@ __cbuf_alloc_meta_inconsistent(struct cbuf_alloc_desc *d, struct cbuf_meta *m)
...
-static inline void *
-cbufp_alloc(unsigned int sz, cbufp_t *cb)  { return __cbuf_alloc(sz, (cbuf_t*)cb, 0); }
-
 /* 
  * precondition:  cbuf lock must be taken.
  * postcondition: cbuf lock has been released.
  */
 static inline void
-__cbuf_done(int cbid, int tmem)
+cbuf_free(cbuf_t cb)
```
gparmer: I'm not sure it is reasonable to expect the user to always track the whole cb struct, and pass it in to free.  Perhaps it is, but I think that components like the file-system (when reprogrammed to use cbufs) will be the test.

src/components/include/cbuf.h
```diff
@@ -405,241 +341,131 @@ __cbuf_alloc_meta_inconsistent(struct cbuf_alloc_desc *d, struct cbuf_meta *m)
...
-		assert(d && !__cbuf_alloc_meta_inconsistent(d, cm));
-
-		fl = d->flhead;
-		assert(fl);
-		ADD_LIST(fl, d, next, prev);
-		cm->owner_nfo.thdid = 0;
+		assert(owner);
+		fl = __cbuf_freelist_get(cm->sz << PAGE_ORDER);
+		CBUFM_SET_NEXT(cm, CBUFM_GET_NEXT(fl));
+		cos_mem_fence();
```
gparmer: You need to document the shared memory protocol for communication between client and server.  I'll need those to understand these fences.

src/components/include/cbuf.h
```diff
@@ -405,241 +341,131 @@ __cbuf_alloc_meta_inconsistent(struct cbuf_alloc_desc *d, struct cbuf_meta *m)
...
+		CBUFM_DEC_REFCNT(cm);
+		cos_mem_fence();
+		target = (unsigned long long *)(&fl->next_flag);
+		old    = (unsigned long long *)(&head.next_flag);
+		update = (unsigned long long *)(&new_head.next_flag);
+		new_head.next_flag = cm;
+		do {
+			*old = *target;
```
gparmer: Line up the `=`s.

src/components/include/cbuf.h
```diff
@@ -405,241 +341,131 @@ __cbuf_alloc_meta_inconsistent(struct cbuf_alloc_desc *d, struct cbuf_meta *m)
...
+			new_head.cbid.tag  = fl->cbid.tag+1;
+		} while(unlikely(!cos_dcas(target, *old, *update)));
 	}
-	/* do this last, so that we can guarantee the manager will not steal the cbuf before now... */
-	cm->nfo.c.refcnt--;
-	if (tmem) cos_comp_info.cos_tmem_available[COMP_INFO_TMEM_CBUF]++;
-	else      relinq = cm->nfo.c.flags & CBUFM_RELINQ;
-	CBUF_RELEASE();
-	
+	else CBUFM_DEC_REFCNT(cm);
```
gparmer: Should be 

`...
} else {
        CBUFM_DEC_REFCNT(cm);
}

src/components/include/cbuf.h
```diff
@@ -405,241 +341,131 @@ __cbuf_alloc_meta_inconsistent(struct cbuf_alloc_desc *d, struct cbuf_meta *m)
...
-			assert(lock_contested(&cbuf_lock) != cos_get_thd_id());
-			return;
-		}
-	} else if (unlikely(relinq)) {
-		cbufp_delete(cos_spd_id(), cbid);
-		assert(lock_contested(&cbuf_lock) != cos_get_thd_id());
-		return;
-	}
-	
+	if (unlikely(relinq)) cbufp_delete(cos_spd_id(), id);
```
gparmer: Update the interface namespace?  `cbufp_delete` -> `cbuf_delete`

src/components/include/cbuf.h
```diff
@@ -405,241 +341,131 @@ __cbuf_alloc_meta_inconsistent(struct cbuf_alloc_desc *d, struct cbuf_meta *m)
...
-	CBUF_TAKE();
-	d  = __cbuf_alloc_lookup(idx);
-	id = (likely(d)) ? d->cbid : 0;
-	CBUF_RELEASE();
-	assert(lock_contested(&cbuf_lock) != cos_get_thd_id());
-
-	return id;
-}
-
 /* Is the pointer a cbuf?  */
```
gparmer: I just love to see this much deleted code.

src/components/include/cbuf.h
```diff
@@ -405,241 +341,131 @@ __cbuf_alloc_meta_inconsistent(struct cbuf_alloc_desc *d, struct cbuf_meta *m)
...
 static inline int 
 cbuf_is_cbuf(void *buf)
 {
-	return cbuf_id(buf) != 0;
+	return false;   //???
```
gparmer: What is `false`?  I don't think that should be defined anywhere.  `return 0;`.

This is used to see if a passed in memory region is actually backed by a cbuf.  Once Gedare's work is done, this will always be true, so we can probably remove this function.

src/components/include/cos_component.h
```diff
@@ -10,6 +10,7 @@
 
 #include <consts.h>
 #include <cos_types.h>
+#include <util.h>
```
gparmer: Probably not a great name for a header file.  Too generic.

src/components/interface/cbufp/cbuf_c.c
```diff
@@ -0,0 +1,189 @@
+/**
```
gparmer: Again, in the future, we'll want to `cbufp` -> `cbuf`.  If you have the time, do it now, unless it will take a lot of time.

src/components/interface/cbufp/cbuf_c.c
```diff
@@ -0,0 +1,189 @@
...
+#include <cbufp.h>
+#include <cbuf_vect.h>
+#include <cos_vect.h>
+#include <cos_debug.h>
+#include <cos_alloc.h>
+#include <valloc.h>
+
+#define CSLAB_ALLOC(sz)   alloc_page()
+#define CSLAB_FREE(x, sz) free_page(x)
+#include <cslab.h>
```
gparmer: You aren't using slabs any more (for freelist node allocation), so you should really remove this.

src/components/interface/cbufp/cbuf_c.c
```diff
@@ -0,0 +1,189 @@
...
+ */
+extern struct cos_component_information cos_comp_info;
+
+CVECT_CREATE_STATIC(meta_cbuf);
+PERCPU_VAR(cbuf_alloc_freelists);
+
+/*** Slow paths for each cbuf operation ***/
+
+/* 
+ * Precondition: cbuf lock is taken.
```
gparmer: Update the comments please.  No more lock, so I don't think that a precondition is that the lock is taken.

src/components/interface/cbufp/cbuf_c.c
```diff
@@ -0,0 +1,189 @@
...
+		assert((unsigned)amnt <= CSP_BUFFER_SIZE);
+
+		if (amnt > 0) {
+			if (CK_RING_DEQUEUE_SPSC(cbufp_ring, &csp->ring, &el)) {
+				cbid = el.cbid;
+				/* own the cbuf we just collected */
+				cm = cbuf_vect_lookup_addr(cbid_to_meta_idx(cbid));
+				assert(cm);
+				assert(cm->cbid.cbid == cbid);
+				/* (should be atomic) */
```
gparmer: If it should be atomic, make it so.

src/components/interface/cbufp/cbuf_c.c
```diff
@@ -0,0 +1,189 @@
...
+				CBUF_SET_TOUCHED(cm);
+				assert(CBUFM_GET_REFCNT(cm) < CBUF_REFCNT_MAX);
+				CBUFM_INC_REFCNT(cm);
+			} else {
+				/* Someone stole the cbufs I collected! */
+				amnt = 0;
+			}
+		}
+		/* ...add the rest back into freelists. construct a temporal freelist
```
gparmer: temporal -> temporary

src/components/interface/cbufp/cbuf_c.c
```diff
@@ -0,0 +1,189 @@
...
+			} else {
+				/* Someone stole the cbufs I collected! */
+				amnt = 0;
+			}
+		}
+		/* ...add the rest back into freelists. construct a temporal freelist 
+		 * first, then add this temporal list to freelist atomically*/
+		struct cbuf_meta *meta, *tail, *head, old_head, new_head;
+		int cb, idx;
+		if (CK_RING_DEQUEUE_SPSC(cbufp_ring, &csp->ring, &el)) {
```
gparmer: This is all rather dense code.  Can you add in a few empty lines to separate out the separate thoughts?

src/components/interface/cbufp/cbuf_c.c
```diff
@@ -0,0 +1,189 @@
...
+	int amnt = 0, i;
+	static struct cbufp_shared_page *csp = NULL;
+
+	assert(cbid <= 0);
+	if (cbid == 0) {
+		struct cbuf_meta *cm;
+		struct cbufp_ring_element el;
+		if (!csp) csp = (struct cbufp_shared_page*)cbufp_map_collect(cos_spd_id());
+		/* Do a garbage collection */
+		amnt = cbufp_collect(cos_spd_id(), size);
```
gparmer: An open question:  Do we want to collect only the cbufs of the specific size, or do we want to collect all cbufs, with a priority for those of the given size?  This is a trade-off between amortizing the cbuf_collect calls, and doing computation that is guaranteed to not be useful at the current time.

src/components/interface/cbufp/cbuf_c.c
```diff
@@ -0,0 +1,189 @@
...
+				new_head.cbid.tag = meta->cbid.tag+1;
+			} while(unlikely(!cos_dcas(target, *old, *update)));
+		}
+	}
+	/* Nothing collected...allocate a new cbufp! */
+	if (amnt == 0) {
+		cbid = cbufp_create(cos_spd_id(), size, cbid*-1);
+		if (cbid == 0) assert(0);
+	} 
+	/* TODO update correctly */
```
gparmer: What does "correctly" mean?

src/components/interface/cbufp/cbuf_c.c
```diff
@@ -0,0 +1,189 @@
...
+	assert(cm && CBUFM_GET_REFCNT(cm));
+done:   
+	return cm;
+}
+
+CCTOR static void
+cbuf_init(void)
+{
+	//initialize freelist head
+	return ;  //???
```
gparmer: If nothing is done here, then the function is not needed.

src/components/interface/cbufp/cbuf_meta.h
```diff
@@ -0,0 +1,139 @@
...
+ *
+ * Author: Gabriel Parmer, gparmer@gwu.edu, 2012
+ */
+
+#ifndef CBUF_META_H
+#define CBUF_META_H
+#define CBUFM_REFCNT_SZ     9
+#define CBUFM_FIRST_FLAGS_SZ      3
+#define CBUFM_SECOND_FLAGS_SZ      4
+#define CBUFM_NEXT_MASK 0xf
```
gparmer: Capital letters.  `0xF`.

src/components/interface/cbufp/cbuf_meta.h
```diff
@@ -0,0 +1,139 @@
...
+#define CBUFM_PTR_MASK 0xfff
+#define CBUFM_INC_UNIT     (1<<CBUFM_FIRST_FLAGS_SZ)
+#define CBUFM_DEC_UNIT     (~(CBUFM_INC_UNIT)+1)
+#define CBUF_REFCNT_MAX   ((1<<CBUFM_REFCNT_SZ)-1)
+#define CBUF_SENDRECV_MAX  ((1<<8)-1)
+
+/*get flags*/
+#define CBUF_OWNER(meta)  ((meta)->nfo & CBUFM_OWNER)
```
gparmer: try lining up as many of these as is reasonable within the different groups.  It is hard to quickly see where the macro name ends, and the contents begin without this.

src/kernel/include/shared/util.h
```diff
@@ -0,0 +1,67 @@
+#ifndef UTIL_H
```
gparmer: Qi should already have definitions for many of these.  Talk to him about where he put them, and what they're called.  `util.h` is too generic a name.  Currently the file should be called `atomic_sync.h`, I'd imagine.

gparmer: Yes, also see Gedare's comment about the concurrency toolkit

src/components/interface/cbuf_mgr/cbuf_meta.h
```diff
@@ -0,0 +1,139 @@
...
+	CBUFM_TOUCHED  = 1<<2,
+	/* 
+	 * Is the mapping writable, or read-only?
+	 * Invariant: CBUFM_OWNER by default
+	 */
+	CBUFM_WRITABLE = 1<<3
+} cbufm_second_flags_t;
+
+union cbufm_ownership {
+	u16_t thdid;        	/*  CBUFM_TMEM && sz == 0 */
```
gparmer: remove this if it isn't used.

src/components/interface/cbuf_mgr/cbuf_meta.h
```diff
@@ -0,0 +1,139 @@
...
+	(meta)->next_flag = (struct cbuf_meta *)((int)((meta)->next_flag) & CBUFM_NEXT_MASK);   \
+	(meta)->next_flag = (struct cbuf_meta *)((int)((meta)->next_flag) | (int)(next)); } while(0)
+#define CBUFM_SET_NEXT_NULL(meta) (meta)->next_flag = (struct cbuf_meta *)((int)((meta)->next_flag) & CBUFM_NEXT_MASK)
+
+/** 
+ * manually assign each bit to prevent compiler reorder   
+ *                                                        
+ * structure of nfo in meta data as follows:              
+ * +-------20-----+-----9-----+---------3---------------+ 
+ * | page pointer |  refcnt   | owner|tmem|inconsistent |
```
gparmer: You can shrink the `refcnt` to 8 bits, and move the `relinq` bit to this flag field so that you can get rid of the other flags.  We don't need `writable` or `touched`.  The `next_flag` struct can be devoted to only the next pointer.

src/components/interface/cbuf_mgr/cbuf_meta.h
```diff
@@ -0,0 +1,139 @@
...
+ * +-------20-----+-----9-----+---------3---------------+ 
+ * | page pointer |  refcnt   | owner|tmem|inconsistent | 
+ * +--------------+-----------+-------------------------+ 
+ *
+ * structure of next_flag in meta data as follows:              
+ * +-------28------+-----------------4------------------+ 
+ * | next pointer  |  writable| touched| relinq| unused | 
+ * +---------------+------------------------------------+ 
+ * Shared page between the target component, and us */
+typedef	struct spd_cbvect_range shared_component_info;
```
gparmer: Why the big space here?

PR #117, opened 2015-02-17
cbuf fix some bugs

ryuxin:
I fix some bugs in current cbuf manager code.
I am afraid those bugs affect Gedare's work, so please merge it if possible.
First is cbuf meta range compare is incorrect.
Second is a dead loop: header is removed from the list, we never get to header again.
Last one is update the cbuf_bin.
As I change the READ-ONLY cbuf, now it can pass unit_torrent and unit_cbuf test.


gedare:
Looks good to me.

PR #118, opened 2015-02-18
Rm stkmgr

gedare:
I removed pgfault/stk_trace, pruned unused functions from the stkmgr interface, implemented the remaining 2 functions in cbuf_mgr, and deleted the stkmgr implementation.

The implementation in cbuf_mgr still uses the cinfo shared page to "return" the allocated stack. This way I did not have to change the assembly code that deals with stack allocation (yet).

Tested with: unit_cbuf.sh


gparmer:
It comes with great pleasure that Gedare has excised from the code-base this monstrosity of complexity.  Sometimes good research doesn't result in a maintainable system.

However, please do a pull, and fix merge conflicts with Yuxin's fix (just merged).  Sorry for the pain.

gedare:
Fixed. Tested with unit_torrent.sh. Please merge.

PR #122, opened 2015-03-03
cbboot: cbuf-backed components

gedare:
There are two main changes here, two bug fixes, and a bit of reformatting.

One bug fix is for cbuf_map_at() where the interface was trying to use too many arguments. Now the 'flags' field, which is used to specify MAPPING_READ or MAPPING_RW, is stashed in the low-order bits of the virtual address used for the map destination. Note: we could use even more bits if we want to enforce that all cbuf_map_at() are page-aligned, which is currently how the interface gets used anyways. I also use another bit of the flags to indicated whether cbuf_map_at() should also reserve the mapped address in the destination component using valloc_alloc_at(), or not.

Another bug fix is in the booter code, which was not updating the stored heap pointer correctly when it was modified within the booter via access to the cos_comp_info. **Note: there may still be problems if the heap pointer is modified in cos_comp_info and cinfo_set_heap_pointer() is not also called.**

The changes:
1. Refactor the booter logic to first allocate memory, then map it, then populate it. Previously, the booter would allocate+map, and then populate. Splitting allocation makes it easier to ensure a contiguous address space for booted components in higher-level booters, which simplifies the map and populate logic.
2. Use cbufs as the backing storage for components booted by cbboot. The cbboot component now cbuf_alloc()s page-sized cbufs instead of using mman_get_page. The cbuf's memory is mman_alias()ed into cbboot overlapping with the pre-allocated contiguous region that booter expects to use to populate component memory. **Note: cbboot uses about twice as much memory because it allocates both a contiguous region AND cbufs for component memory.** If memory pressure becomes a problem, we can probably think of some ways to remove the duplication. Also, the cbufs that get allocated are currently just tracked through a massive array, which should be improved in the future. **The array of cbufs is currently static-allocated with an estimated size that should be made either dynamic or statically-bounded.** However, for the current users of cbboot (unit_cbboot_cbuf.sh), the current solution works fine.

Tested with unit_cbboot_cbuf.sh.

### DIFF
--- a/src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
+++ b/src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
@@ -456,11 +456,12 @@ free:
 }
 
 vaddr_t
-cbuf_map_at(spdid_t s_spd, cbuf_t cbid, spdid_t d_spd, vaddr_t d_addr, int flags)
+cbuf_map_at(spdid_t s_spd, cbuf_t cbid, spdid_t d_spd, vaddr_t d_addr)
 {
 	vaddr_t ret = (vaddr_t)NULL;
 	struct cbuf_info *cbi;
 	u32_t id;
+	int flags;
 	
 	cbuf_unpack(cbid, &id);
 	CBUF_TAKE();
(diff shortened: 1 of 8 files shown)


gedare:
I'm working on the following improvements:
1. Revert cinfo to store the cos_comp_info pointer and use that to get the hp. This fixes an issue with updates to the hp after storing it via cinfo interface.
2. Convert page-at-a-time cbuf allocations to sect-at-a-time.
3. Use sect->flags to inform cbuf_map_at flags.
4. Replace the array_of_cbufs with a cos_vect that maps from spdid to a (dynamic) array of cbufs. [Is there a better data structure to use in cos?]

These improvements take care of the first and third **bold-face notes** in the pull request description. The second note is going to remain an issue for now I think, for various reasons we will need to have this duplication of memory in the final cbboot system.

I anticipate the improvements to be done by tomorrow and will add them to this pull request if it is still open.

gparmer:
Easy to review, easy to merge.  Looks great.   The memory trade-off seems fine for now.  A vector is probably the data-structure you're looking for as it will be more dynamic.

gedare:
OK I will create a new pull request after I finish the improvements I laid out.

PR #125, opened 2015-03-11
Cbboot sect

gedare:
Allocate cbufs in "sections" at a time rather than pages for cbboot-loaded components. This should make it easier to manage these components. cbboot now also uses read-only mappings for sections that do not have the COBJ_SECT_WRITE flag set.

This is done by refactoring the logic for bood_spd_map_memory() to make a call to a new function in boot_deps.h that must map all of the memory for a section. A side effect is that local_mman calls are no longer used in booter.c, but now there is some duplication of code in boot/boot_deps.h and llboot/boot_deps.h

### DIFF
--- a/src/components/implementation/no_interface/boot/boot_deps.h
+++ b/src/components/implementation/no_interface/boot/boot_deps.h
@@ -48,7 +48,7 @@ boot_deps_init(void)
 }
 
 static void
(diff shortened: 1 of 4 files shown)


gedare:
I've made boot and llboot also honor the COBJ_SECT_FLAGS for read/write permissions. This required copying the "deps" page due to it being modified during component booting. An alternate solution would be to put cos_poly[2] into a writable section from cos_loader, but I didn't see a simple way to do that.

PR #126, opened 2015-03-23
mem_man: save flags in struct mapping

gedare:
This is a latent bug in the mman code. The flags are not correctly saved, they're always 0.


gedare:
Added NULL initializer for child mapping.

PR #127, opened 2015-05-21
update fpu code, bug fix in mem_mgr and valloc

ryuxin:
Tested with lfpu, lmicro_fpu, unit_cbuf and micro_cbuf

### DIFF
--- a/src/components/implementation/mem_mgr/naive/mem_man.c
+++ b/src/components/implementation/mem_mgr/naive/mem_man.c
@@ -87,13 +87,20 @@ frame_alloc(void)
 }
 
 static inline void 
+frame_free(struct frame *f)
+{
+	assert(f->nmaps == 0);
+	f->c.free = freelist;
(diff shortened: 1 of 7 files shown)


gedare:
Looks good. I'm not sure why you use two different functions to set the ext_size in similar code in valloc_alloc() and valloc_alloc_at() though.

ryuxin:
ext_size is used in different ways in valloc_alloc() and valloc_alloc_at(). You used it as the total size of memory in valloc_alloc_at(), but before it is used as the number of pages in valloc_alloc().
I can unify the usage in both functions.

gedare:
I don't remember doing it that way for any good reason. Maybe it was more convenient to have it in bytes for me, for some reason, I don't know.

gparmer:
In the future, please separate separate features into separate pulls (this is two features: fpu and a memory manager fix).  Sorry for the delay on this!

PR #128, opened 2015-06-04
Upcall Component Interface

gedare:
These patches add an interface, implementation, and unit test for a new mechanism, upcall_invoke(), that can make upcalls into arbitrary components at the highest priority passing an upcall type and a single extra argument. This mechanism writes to a structure in the low-level booter and then switches to a dedicated thread that runs at highest priority in llboot. This thread reads the structure to determine the parameters to pass to cos_upcall. The mechanism and code stems from Jiguo's work.

In addition, cos_upcall() is changed to take an upcall_type_t as its first argument, and "recovery" is renamed to "upcall" in the low-level booter to make the upcall framework more general than only used for recovery.

### DIFF
--- a/src/components/implementation/no_interface/llboot/Makefile
+++ b/src/components/implementation/no_interface/llboot/Makefile
@@ -1,7 +1,7 @@
 C_OBJS=booter.o
 ASM_OBJS=cos_asm_scheduler.o
 COMPONENT=llboot.o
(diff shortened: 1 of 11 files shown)


gparmer:
Sorry for the delay, you caught me in the lead-up to the vacation.

I'll pull this as the pending changes are small.

PR #130, opened 2015-06-12
C stub

gedare:
Converts the assembly-based default stubs to C-based stubs. Optimizes the existing cstub.h to remove some extra instructions. There are about 6 more instructions (give or take) along the invocation path, due to the compiler-generated code for pushing a new frame in the prolog and then popping it in the epilog.

Below picture shows the objdump on the left is the assembly-based code, and on the right is the new C-based code.
print jpg-1


gedare:
With this change, unit_torrent is failing. I need to dig some more.

gparmer:
The torrent interface is (I *believe*) the only interface that returns multiple arguments in registers.  This is likely breaking the calling return conventions.

I'm not sold on this as I don't see how it will work well with the C^3 work.  I am probably just missing something in the assembly, but I don't see where the proper faulty return value (`EFAULT` or something comparable) is being returned (i.e. the `$0xffff2153` in the original code).

Repull when you have the next version that has proper return calling conventions, with tests for maintaining C^3 functionality.

Issue #129, opened 2015-06-11
cstub using C wrappers can be optimized more


gedare:
When client stubs are generated using the `CSTUB_INVOKE` macro from cstub.h, the compiler will generate code to save and restore registers that are also being saved and restored by the inline assembly. With a little bit of digging, we can eliminate some of these instructions in the inline assembly.

PR #132, opened 2015-06-15
c_stub: rewrite default cstubs to use C

gedare:
Replaces the assembly in c_stub.S with C function wrapping the macros from cstub.h that use inline assembly to invoke a component. See #129 for possible performance ramifications that should be investigated further. I was not successful in generalizing a way to use the compiler's automatic frame setup and saving of ebx/esi/edi registers, because the sysenter expects the frame to have a different layout so we need to create one. Instead, the stubs always save and restore the same set of registers regardless of what was needed.

Currently the error code is translated to a binary condition, 0 is stored into ecx tied to the fault argument in case there was no fault handling, and 1 is stored into ecx in case there was fault handling. This is consistent with current work in C^3, but I plan to eventually relay more information back from the fault handler directly, which may even write directly into ecx and eax so that the return path just "Gets it".


src/components/lib/c_stub.c
```diff
@@ -0,0 +1,11 @@
+
+#include <cos_types.h>
+#include <cstub.h>
+
+__attribute__((regparm(1))) int
```
gparmer: clever use of regparm here to capture the dynamically generated stub's behavior (from `gen_client_stubs.c`).

gparmer:
I'm not sure I see much benefit to moving this to C.  This code is not inlined even with linker tricks, and all of the assembly is still there...  However, generically, the more that is in C, the better.

Someone just likes to hack assembly ;-)

gedare:
Just for some context, I have plans to expand the stub code, and having a nice way to hook into it from C will be good.

gparmer:
That makes sense.  Jiguo is already doing this using the C^3 stubs, so you might want to look into their mechanisms for constructing the C + assembly.

gparmer:
https://github.com/songjiguo/Composite/blob/master/src/components/interface/sched/__stubs_rec/c_stub.c#L157 for example

PR #134, opened 2015-06-22
cstubs: add fault return path and return fault code from handlers

gedare:
Leverage the return value from the fault handler invocation to relay fault codes. The return path in the default client stub after fault handling now invokes a new C function (```SS_ipc_client_fault```). Currently, the default just translates the fault handler return value into an errno code.


gedare:
I've extended the framework to permit redefining the behavior of the default cstub fault handling path. Now ```SS_ipc_client_fault``` is weak in the default version. I have added a strong replacement in cbuf, which will likely go away eventually after we figure out a better way to provide alternative fault handling paths, but for now is useful for demonstration purposes.

PR #135, opened 2015-07-02
Patch for building on x86_64 machines

RobertGiff:


gparmer:
Can't automerge as I pulled Qi's request.  Can you fix the pull and reissue the request?  The conflict is likely that he fixed the same bug in `pong.c` that you did.

RobertGiff:
Should be good now.

gparmer:
I don't really know what happened here, but it looks like you're recommitting a bunch of changes I made to the build system (see all the moved `Makefile`s).  What's happening there?

gparmer:
You should also name this somewhat more appropriately:  the name would give the impression that this is code for the native booting to boot on x86_64 machines.  Instead, it simple enables building the system on x86_64 systems.

gparmer:
Looking at the commits for this, it seems like you're also including commits that I've made.  You likely just want to just pull the most up-to-date branch, and make the appropriate changes.  Then do a pull request.  If you want to delay this to instead keep working on an older commit, that's fine.  Perhaps @atryus can commit the same as got everything compiling on x86_64 already as well.

RobertGiff:
Rebase done. That was annoying.

PR #139, opened 2015-07-10
Fixed the Linux booting up environment. Code cleanup.

interwq:


gparmer:
Only comment is this:  Please stop using `COS_LINUX`.  We should really use `COS_PLATFORM` to put an end to the endless hacks.

PR #142, opened 2015-07-14
Basic implementation for rump kernel / unikernel booter.

RobertGiff:
Also implemented the option to pass the -v or --verbose flag to the cos_linker program.


gedare:
Is unikernboot a replacement for llboot, or does it rely on llboot? I ask because there are some llboot features in the unikernboot's boot_deps.h that seem out of place. Also, the runscript unikernboot.sh does not load the .o file, so it is incomplete for testing purposes.

RobertGiff:
Unikernboot is a replacement for llboot. I removed as much as of llboot-like features out of its boot_deps.h as I thought was necessary in a single go, to simplify the booter while still getting it to the execution that I need. As such I'm sure there are a lot of llboot related features still left over and any recommendations for what to take out would be awesome.

unikernboot.o is linked with a standalone rump kernel application so that the unikernbooter can natively execute and boot the rump kernel. After linking the two together, the resulting .o is renamed to be llboot.o for cos_linker related reasons.

gedare:
OK cool. Please have a look at the current llboot with the rename from recovery to upcall. I'm not certain you really need to support the recovery/upcall mechanisms at this point, which would let you get rid of some more complexity by removing that code. Gabe?

gparmer:
Yeah, sorry Robbie, but this still has way too much stuff in it.  I'm not convinced it needs much more than sched_init calling cos_init.

Most of the allocation methods should be replaced with `cos_kernel_api.h|c` that I'm working on now (wanted to finish by Monday).  All of the MM/sched stuff can be removed.  Perhaps all that should stay is the print functions and sched_init/cos_init.  Do you need anything else?

Sorry for the delay here.

RobertGiff:
No, I don't believe so. Thank you both for the feedback!

PR #143, opened 2015-07-20
support constructor and destructor under gcc 4.8

ryuxin:
After 4.8, gcc puts constructor and destructor into init_array and fini_array section, instead of ctors and dtors.
Test with cbuf constructor in cbuf_mgr.


src/platform/linux/link_load/cos_loader.c
```diff
@@ -147,6 +147,8 @@ typedef enum {
 	RODATA_S, 
 	CTORS_S, 
 	DTORS_S, 
+	INIT_ARRAR_S,
```
gedare: s/ARRAR/ARRAY

PR #153, opened 2015-10-26
Changes to run on bare-metal

phanikishoreg:
Major changes:
1. memset on thread struct.
2. parsing mod->cmdline for bare-metal.

Other changes: 
1. Added 'FOREVER' in khalt() which is useful in debugging especially on Bare-metal.
2. platform/i386/vga.c is picked up from online sources (may not comply with Composite style guide). 
3. Double fault stub added.

### DIFF
--- a/src/platform/i386/kernel.c
+++ b/src/platform/i386/kernel.c
@@ -13,6 +13,10 @@
 #include <component.h>
 #include <thd.h>
 
+#define ADDR_STR_LEN	8
+#define CMDLINE_MAX_LEN	32
+#define CMDLINE_REQ_LEN	(ADDR_STR_LEN * 2 + 1)
+
 struct mem_layout glb_memlayout;
 
 static int
@@ -68,14 +72,19 @@ kern_memory_setup(struct multiboot *mb, u32_t mboot_magic)
 	for (i = 0 ; i < mb->mods_count ; i++) {
 		struct multiboot_mod_list *mod = &mods[i];
 
-		printk("\t- %d: [%08x, %08x)", i, mod->mod_start, mod->mod_end);
+		char *cmdline = (char *)mod->cmdline;
+		int cmdline_len = strnlen((const char*)cmdline, CMDLINE_MAX_LEN);
+		assert(cmdline_len >= CMDLINE_REQ_LEN);
+		int addr_offset = cmdline_len - CMDLINE_REQ_LEN;
+
+		printk("\t- %d: [%08x, %08x) : %s", i, mod->mod_start, mod->mod_end, mod->cmdline);
 		/* These values have to be higher-half addresses */
 		glb_memlayout.mod_start = chal_pa2va((paddr_t)mod->mod_start);
 		glb_memlayout.mod_end   = chal_pa2va((paddr_t)mod->mod_end);
 
-		glb_memlayout.bootc_vaddr = (void*)hextol((char*)mod->cmdline);
-		assert(((char*)mod->cmdline)[8] == '-');
-		glb_memlayout.bootc_entry = (void*)hextol(&(((char*)mod->cmdline)[9]));
+		glb_memlayout.bootc_vaddr = (void*)hextol((char *)(cmdline + addr_offset));
+		assert(cmdline[addr_offset + ADDR_STR_LEN] == '-');
+		glb_memlayout.bootc_entry = (void*)hextol((char *)(cmdline + addr_offset + ADDR_STR_LEN + 1));
 		printk(" @ virtual address %p, _start = %p.\n",
 		       glb_memlayout.bootc_vaddr, glb_memlayout.bootc_entry);
 	}
@@ -90,7 +99,7 @@ kern_memory_setup(struct multiboot *mb, u32_t mboot_magic)
 	/* FIXME: check memory layout vs. the multiboot memory regions... */
 
 	/* Validate the memory layout. */
-	assert(mem_kern_end()   == mem_bootc_start());
+	assert(mem_kern_end()   <= mem_bootc_start());
 	assert(mem_bootc_end()  <= mem_boot_start());
 	assert(mem_boot_start() >= mem_kmem_start());
 	assert(mem_kmem_start() == mem_bootc_start());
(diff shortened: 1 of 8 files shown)


src/platform/i386/kernel.c
```diff
@@ -13,6 +13,10 @@
 #include <component.h>
 #include <thd.h>
 
+#define ADDR_STR_LEN	8
+#define CMDLINE_MAX_LEN	32
+#define CMDLINE_REQ_LEN	(ADDR_STR_LEN * 2 + 1)
```
gparmer: this is very strange spacing.

src/platform/i386/kernel.c
```diff
@@ -13,6 +13,10 @@
 #include <component.h>
 #include <thd.h>
 
+#define ADDR_STR_LEN	8
+#define CMDLINE_MAX_LEN	32
```
gparmer: make all of these values line up

src/platform/i386/kernel.c
```diff
@@ -68,14 +72,19 @@ kern_memory_setup(struct multiboot *mb, u32_t mboot_magic)
...
+
+		printk("\t- %d: [%08x, %08x) : %s", i, mod->mod_start, mod->mod_end, mod->cmdline);
 		/* These values have to be higher-half addresses */
 		glb_memlayout.mod_start = chal_pa2va((paddr_t)mod->mod_start);
 		glb_memlayout.mod_end   = chal_pa2va((paddr_t)mod->mod_end);
 
-		glb_memlayout.bootc_vaddr = (void*)hextol((char*)mod->cmdline);
-		assert(((char*)mod->cmdline)[8] == '-');
-		glb_memlayout.bootc_entry = (void*)hextol(&(((char*)mod->cmdline)[9]));
+		glb_memlayout.bootc_vaddr = (void*)hextol((char *)(cmdline + addr_offset));
```
gparmer: Why did you replace the previous support for placement of the vaddr of the boot component based on the file name for the module with this?  Won't that break quite a bit?  won't that require a lot of info that the user doesn't have on the command line?

src/platform/i386/vga.c
```diff
@@ -0,0 +1,204 @@
+#define ENABLE_VGA
```
gparmer: LIkely this shouldn't be here.  It should likely be in `cos_config.h`.

src/platform/i386/vga.c
```diff
@@ -0,0 +1,204 @@
...
+/* The number of columns. */
+#define COLUMNS                 80
+/* The number of lines. */
+#define LINES                   25
+/* The attribute of an character. */
+#define ATTRIBUTE               0x0F
+/* The video memory address. */
+#define VIDEO                   0xB8000
+
+#define SPACE			0x20
```
gparmer: Line all of these up....I'm not oging to comment on this file as you know it needs to be changed to accommodate cos style.

gparmer:
We've already discussed many of the necessary changes, including

- style in vga.c
- check that we can use the code in vga.c due to licenses
- use at least a commit per logical change

Additionally:

- Remove `FOREVER` and just use `while(1);`.  This what we do at other places, so lets be consistent.
- Note that by adding this loop, you're making qemu much more annoying to use.  This might be the right trade-off, but you need to be aware that you're making such a trade-off
- You're making qemu impossible to use by disabling serial.  The rule is to not break things that were working before.  Thus you should make serial and vga configuration options (likely in `cos_config.h`), and set serial to be enabled by default
- It appears to me that you're breaking the previous way of using a module by using the command line stuff.  Why?  Am I misinterpreting this.  We don't want to break things that worked.

src/platform/i386/kernel.c
phanikishoreg: I've not replaced the previous logic, I've only modified to also accommodate bare-metal command line. 
When we run from Qemu, mod->cmdline will only be XXXXXXXX-YYYYYYYY but for Grub, we give the absolute path to the module and we have them in /boot/. So, I'm only stripping down the first "/boot/" if it exists. 

This logic works both on Qemu and Bare-metal as I've been testing every change on both.

src/platform/i386/kernel.c
phanikishoreg: I'll do that. Thank you.

src/platform/i386/kernel.c
phanikishoreg: Is it only the line 16 that's strange or all 3 lines? I've mainly used TAB in Vim in all cases I guess. But I missed to align line 16 correctly.

src/platform/i386/vga.c
phanikishoreg: Actually I saw that
```
#define ENABLE_SERIAL
```
was done in platform/i386/kernel.c but I thought it isn't the right place and moved it to kernel.h instead. I'll look into cos_config.h and will make necessary changes for this. Thank you.

src/platform/i386/vga.c
phanikishoreg: I'll work on the whole file including licensing.

phanikishoreg:
I'll take a look at the necessary changes and try to close them soon.
In additional part:
* Sure, I'll replace with while(1);. Yes, I was aware that I'm making it annoying to use in Qemu because it hangs and every time we have to close the terminal and start a new one. But on bare-metal it is impossible to figure otherwise, as it quickly reboots. I'll keep it and will try to work to see if I can find any other solution.
* I've not modified any part of the serial and while running in Qemu, I've mostly used the serial functionality. Only a few times I've tested by removing -nographics but mostly used with it. Because, the number of lines are limited to 24, It's very tough to get the whole snapshot of logs, so I still use -nographics from Qemu. And again, I've not disabled SERIAL, it is kept as it is in kernel.c. Actually as I mentioned previously, in the current code, #define ENABLE_SERIAL is at the top of kernel.c and not in cos_config.h, so it still is enabled and untouched. 
* No. I've explained my observation with the way we use module in Grub, I felt it is necessary to make room but I've not broken the previous functionality, I've only extended it. 

I'll discuss with you on this on Slack to get more inputs from you, thank you for providing all the feed back and I'm sorry that I've so many gaps in the code to resolve.

PR #156, opened 2015-10-29
Bare metal boot stuff, VGA driver - review comments incorporated

phanikishoreg:


src/kernel/include/shared/cos_config.h
```diff
@@ -119,4 +119,11 @@
 /* print out to dmesg? */
 /* #define COS_PRINT_DMESG 1 */
 
+/**
+ * Configuration to enable/disable functionality in Kernel.
+ */
+#define ENABLE_VGA
+#define ENABLE_SERIAL
+#define ENABLE_TIMER
```
gparmer: This commit conflicts with Jakobs.  Fix the merge conflict (sorry!).  I believe he removed `ENABLE_TIMER`.

src/platform/i386/Makefile
```diff
@@ -41,7 +41,8 @@ OBJS += serial.o
 OBJS += timer.o
 OBJS += chal.o
 OBJS += boot_comp.o
-OBJS += console.o
+#OBJS += console.o
```
gparmer: If this is commented out, isn't it going to do nothing when `ENABLE_SERIAL` is defined?  Why is this not compiled now?

src/platform/i386/isr.h
```diff
@@ -6,11 +6,13 @@
 #include "chal_asm_inc.h"
 
 enum {
-	IRQ_PAGE_FAULT = 14,
-	IRQ_PIT        = 32,
-	IRQ_KEYBOARD   = 33,
-	IRQ_SERIAL     = 36
+	IRQ_DOUBLE_FAULT = 8,
+	IRQ_PAGE_FAULT   = 14,
+	IRQ_PIT          = 32,
```
gparmer: You'll find that Jakob had to alter this as well.  Please make sure your changes are consistent with his.

src/platform/i386/kernel.c
```diff
@@ -138,6 +147,9 @@ kmain(struct multiboot *mboot, u32_t mboot_magic, u32_t esp)
        	comp_init();
```
gparmer: Wow, why is the spacing off on these?

src/platform/i386/vga.c
```diff
@@ -0,0 +1,211 @@
...
+	size_t i = 0;
+
+	cll();
+	for (i = 0; i < strnlen((const char*)text, STRLEN_MAX); i++) {
+		putchar(text[i]);
+	}
+	move_csr();
+}
+
+void keyboard_handler(struct pt_regs *regs)
```
gparmer: Don't we already have a keyboard handler?  See Jakob's code.

src/kernel/include/shared/cos_config.h
phanikishoreg: Will do that. :)

src/platform/i386/Makefile
phanikishoreg: Console.o is also a VGA driver but I had created a new one. I can even remove it totally. What do you think? ENABLE_SERIAL is for serial.o and not this one.

src/platform/i386/isr.h
phanikishoreg: Will check and resolve conflicts.

src/platform/i386/kernel.c
phanikishoreg: I think it has been like that from before. I've not touched that part, there is no diff of base to my version at these lines. I'll fix it anyway in my next pull request. :)

src/platform/i386/vga.c
phanikishoreg: Do we have one? I'll check his code. This was the only thing I merged from console.o in this new vga.o.

phanikishoreg:
Unmerged again!! :( 2 failed attempts in a row for me! :P

src/platform/i386/vga.c
phanikishoreg: Just checked, only keyboard_handler() is in console.o and that is not disabled in Jacob's code but in my commit, console.o is disabled and replaced with vga.o and I've included this handler in vga.o, that should be fine right?

PR #165, opened 2015-11-09
Fixing style

georgit:
Hopefully this is everything you wanted!


src/components/implementation/tests/unit_cbufmalloc/posix_test.c
```diff
@@ -0,0 +1,48 @@
...
+
+void
+cos_init(void *args)
+{
+	int err=0;
+	int i, j;
+
+	for (i = 0; i < NUM; i++) {
+		mptr[i] = (char *)malloc((i+1) * sizeof(int));
+
```
gparmer: I see no reason to have a space between every line here.  Empty lines denote the equivalent of paragraphs.  In no case would the two lines in 
```
a = malloc(...);
assert(a)
```

Be considered separate thoughts.  These are clearly part of the same high-level conceptual operation.

src/components/implementation/tests/unit_cbufmalloc/posix_test.c
```diff
@@ -0,0 +1,48 @@
...
+		assert(mptr[i]);
+		
+		for (j = 0; j < i; j++) {
```
gparmer: This is small:  `for (j = 0 ; j < i ; j++) {...` is correct.  Notice the use of spaces.

gparmer:
There are still minor style issues, but I'm pulling this.  Just keep these in mind in the future.

gparmer:
Added `cbuf` support in `malloc` and `free`.

PR #166, opened 2015-11-12
fix multiple returns in client stub

ryuxin:
test with unit_cbuf, unit_torrent and unit_cbboot_cbuf


gparmer:
Thanks again for this Yuxin!  Great job.

PR #168, opened 2015-12-14
Refactoring conditional directives for if wrappers.

wmarquesr:
A suggestion to compile entire statements and expressions, as suggested by code style guidelines from the Linux Kernel and practitioners.

* [Linux kernel coding style](https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/CodingStyle#n892)
* [Living in the #ifdef Hell](https://www.cqse.eu/en/blog/living-in-the-ifdef-hell/)

It might improve code understanding, maintainability and error-proneness.

### DIFF
--- a/src/components/lib/lwip/lwip-1.3.0/src/api/api_msg.c
+++ b/src/components/lib/lwip/lwip-1.3.0/src/api/api_msg.c
@@ -6,9 +6,9 @@
 
 /*
  * Copyright (c) 2001-2004 Swedish Institute of Computer Science.
- * All rights reserved. 
- * 
- * Redistribution and use in source and binary forms, with or without modification, 
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without modification,
  * are permitted provided that the following conditions are met:
  *
  * 1. Redistributions of source code must retain the above copyright notice,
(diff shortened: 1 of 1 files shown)


gedare:
Hi. This would be better discussed as an Issue rather than a Pull request. I'm not sure how the code changes in the Pull Request relate to anything. Perhaps it would also be worth opening a discussion on the Composite Googlegroup mailing list.

wmarquesr:
The idea is always to encompass only complete/entire statements/expressions with preprocessor conditional directives. It is something that is getting popular in many different projects (started mainly on the Linux Kernel).

We agree totally that breaking statements with directives influences code quality, i.e., understanding, maintainability and error proneness. We found some evidence based on interviews with developers and an online survey regarding this negative effect on quality.

[http://www.cs.cmu.edu/~ckaestne/pdf/ecoop15.pdf](http://www.cs.cmu.edu/~ckaestne/pdf/ecoop15.pdf)

Don't you agree with that?

gparmer:
@wmarquesr:  Thanks for the interest, and for finding this in the source!  I couldn't agree with you more that the source that you pointed out is difficult to follow, and that you're modification increases the readability.

However, the change you propose is in an external library (`lwip`) that we include in our system.  The main software engineering concern with that we prioritize with such code is making sure that our diff with their tested code is small, so that if we have to upgrade to a new version of the library, we can do so with fewer changes, thus are less likely to make an error.

I would fully encourage you to talk to the `lwip` folks to accept your change.  I absolutely agree with you, but the right way to make that change in in the upsteam library.  We do frown on such style in our source (though some might have accidentally sneaked in).  If you find any instances of it in our source (external libraries aside), don't hesitate to let us know (the mailing list, or an issue might be more appropriate).

Thanks again for the feedback and pull.  I hope my reasons for not accepting it are clear.

gparmer:
BTW, if you're interested in some strange uses of the preprocessor, have a look at `https://github.com/gparmer/Composite/blob/ppos/src/kernel/include/ertrie.h`.  It might make a good argument about why people use the preprocessor given the lack of certain language features.  Thanks for the link to your paper!

wmarquesr:
Thanks for the feedback, this patch is part of an academic research and just knowing that this kind of refactoring is accepted is already very useful, but of course I will check the link.

Thanks again.

PR #170, opened 2016-01-19
Fixed timer interrupts and created micro-benchmarking tests for SINV/Thread-switch/ASYNC

phanikishoreg:


src/components/implementation/tests/micro_booter/micro_booter.c
```diff
@@ -5,6 +5,8 @@
 #include <cobj_format.h>
 #include <cos_kernel_api.h>
 
+#define	ITERATIONS	1000000
```
gparmer: Why is there no space after `#define`.  Is this a github rendering issue?

src/components/implementation/tests/micro_booter/micro_booter.c
```diff
@@ -56,6 +58,50 @@ static void
 tls_set(size_t off, unsigned long val)
 { __asm__ __volatile__("movl %0, %%gs:(%1)" : : "r" (val), "r" (off) : "memory"); }
 
+long long total_swt_cycles = 0LL;
+long long start_swt_cycles = 0LL, end_swt_cycles = 0LL;
+static void
+thd_fn_perf(void *d)
+{
+	/* printc("\tNew thread %d with argument %d, capid %d\n", cos_thdid(), (int)d, tls_test[(int)d]); */
```
gparmer: Please remove commented out code.

src/components/implementation/tests/micro_booter/micro_booter.c
```diff
@@ -56,6 +58,50 @@ static void
 tls_set(size_t off, unsigned long val)
 { __asm__ __volatile__("movl %0, %%gs:(%1)" : : "r" (val), "r" (off) : "memory"); }
 
+long long total_swt_cycles = 0LL;
```
gparmer: shouldn't all of these variables be local within the function?  They aren't accessed outside of the function, right?  So why global?

src/components/implementation/tests/micro_booter/micro_booter.c
```diff
@@ -97,6 +143,66 @@ test_mem(void)
...
+
+	cos_asnd(sc);
+	/* printc("sending thread!!\n"); */
+
+	start_asnd_cycles = 0LL, end_asnd_st_arcv_cycles = 0LL, end_arcv_cycles = 0LL;
+	/* simplest measurement: not accurate. */
+	long long iters = 0LL;
+	rdtscll(start_asnd_cycles);
+	for (; iters < ITERATIONS; iters ++) {
+		/* start_asnd_cycles = 0LL, end_asnd_st_arcv_cycles = 0LL, end_arcv_cycles = 0LL;
```
gparmer: lots of cleaning up to do here

src/components/implementation/tests/micro_booter/micro_booter.c
```diff
@@ -97,6 +143,66 @@ test_mem(void)
...
+		cos_asnd(sc);
+		/* rdtscll(end_arcv_cycles);
+		total_asnd_cycles += (end_asnd_st_arcv_cycles - start_asnd_cycles);
+		total_arcv_cycles += (end_arcv_cycles - end_asnd_st_arcv_cycles); */
+	}
+	rdtscll(end_arcv_cycles);
+	total_asnd_cycles = (end_arcv_cycles - start_asnd_cycles) / 2ll ;
+	printc("Average ASND/ARCV (Total: %lld / Iterations: %lld ): %lld\n", 
+		total_asnd_cycles, (long long) (ITERATIONS), (total_asnd_cycles / (long long)(ITERATIONS)));
+	/* printc("Average ARCV (Total: %lld / Iterations: %lld ): %lld\n",
```
gparmer: This file is as much of documentation of how to use the API as it is a microbenchmark.  Thus we need it to be unambiguous and clear.

src/components/implementation/tests/micro_booter/micro_booter.c
```diff
@@ -249,6 +390,41 @@ test_inv(void)
...
+{
+	compcap_t cc;
+	sinvcap_t ic;
+
+	cc = cos_comp_alloc(&booter_info, booter_info.captbl_cap, booter_info.pgtbl_cap, (vaddr_t)NULL);
+	assert(cc > 0);
+	ic = cos_sinv_alloc(&booter_info, cc, (vaddr_t)__inv_test_serverfn);
+	assert(ic > 0);
+
+	long long iters = 0LL;
```
gparmer: One more time:  variable definitions should be at the top of a block.

src/components/implementation/tests/micro_booter/micro_booter.c
```diff
@@ -249,6 +390,41 @@ test_inv(void)
...
+	long long total_cycles = 0LL;
+	long long total_inv_cycles = 0LL, total_ret_cycles = 0LL;
+	for (; iters < ITERATIONS; iters ++) {
+		midinv_cycles = 0LL;
+		long long start_cycles = 0LL, end_cycles = 0LL;
```
gparmer: ditto.

src/components/implementation/tests/micro_booter/micro_booter.c
phanikishoreg: Sorry.. Its a TAB SPACE. I'll change that.

src/components/implementation/tests/micro_booter/micro_booter.c
phanikishoreg: I kept the code for doing it in 2 ways. 1. Sum of cycles for all iterations. (commented, and in this case I needed that variable to be global.) 2. Total time in all iterations. I thought it would be useful. I'll remove it!

src/components/implementation/tests/micro_booter/micro_booter.c
phanikishoreg: Sure. I'll have code only for measuring using one method and remove other which I kept in comments.

src/kernel/capinv.c
```diff
@@ -434,8 +433,12 @@ cap_switch_thd(struct pt_regs *regs, struct thread *curr, struct thread *next,
 		return 0;
 	}
 
-	copy_gp_regs(regs, &curr->regs);
-	__userregs_set(&curr->regs, 0, __userregs_getsp(&curr->regs), __userregs_getip(&curr->regs));
+	if (!(curr->state & THD_STATE_PREEMPTED)) {
```
gparmer: We talked about modifying the return path in the `entry.S` file.  I'll need to think through the implications of this change.  The downside is that it slows down the thread switch path.

phanikishoreg:
Sorry! Made a silly assumption that test suite code needn't be strictly adhering to style guide. My bad! Will fix it all soon!

src/kernel/capinv.c
```diff
@@ -569,9 +577,8 @@ capinv_int_snd(struct thread *rcv_thd, struct pt_regs *regs)
 	assert(rcv_tcap);
 
 	next = asnd_process(rcv_thd, thd, rcv_tcap, tcap);
-	if (next == thd) return 0;
+	if (next == thd) return 1; /* current thread is a preempted one! */
```
gparmer: Was this a bug before?

src/platform/i386/Makefile
```diff
@@ -5,7 +5,7 @@ LD := ld -m elf_i386
 AS := as --32 -g
 
 INCPATH := ../../kernel/include
-CFLAGS := -g3 -ffreestanding -Wno-unused-function -Wno-unused-variable -Wno-unused-but-set-variable -nostdlib -mno-red-zone -I$(INCPATH)
+CFLAGS := -g3 -O3 -ffreestanding -Wno-unused-function -Wno-unused-variable -Wno-unused-but-set-variable -nostdlib -mno-red-zone -I$(INCPATH)
```
gparmer: -g3 and -O3 at the same time is strange.  I don't know what the output should be.  Shouldn't we just chose one?

src/platform/i386/entry.S
```diff
@@ -72,29 +72,13 @@ PT_OLDSS	= 0x40
...
-
-	/* set up the activation record required by iret */
-	pushl $(SEL_UDSEG)
-	pushl PT_OLDESP(%eax)
-	pushl PT_EFLAGS(%eax)
-	pushl $(SEL_UCSEG)
-	pushl PT_EIP(%eax)
-
-	movl PT_EAX(%eax), %eax
+	RESTORE_REGS_GENERAL
```
gparmer: Nice, you got this working!!!! awesome.  Looks much better.  I believe we can remove `cos_get_curr_thd_asm` now.  Check that it has no additional uses, and remove it if you can.

src/platform/i386/entry.S
```diff
@@ -110,14 +94,10 @@ restore_from_thd:
...
-	popl %ds; 		\
-	popl %es;		\
-	popl %fs;		\
-	popl %gs;		\
-	addl $4, %esp; 		\
-	sti;			\
-	iret
+	testl %eax, %eax;	\
+	jne restore_from_thd;	\
+	RESTORE_REGS_ALL;	\
```
gparmer: We might be able to `jmp` to `restore_from_thd` to minimize the instruction cache footprint?

src/platform/i386/idt.c
```diff
@@ -110,5 +110,6 @@ idt_init(void)
 	idtr.length = idt_ptr.limit;
 	idtr.base = (unsigned long)idt_entries;
 
-	asm volatile("lidt (%0)" : : "p"(&idtr));
+	/* asm volatile("lidt (%0)" : : "p"(&idtr)); */
+	idt_flush((u32_t)&idtr);
```
gparmer: Why do we need this?  Why doesn't the inline assembly work?  This was the annoying compiler issue, right?

src/platform/i386/timer.c
```diff
@@ -54,7 +54,7 @@
 #define TN_FSB_EN_CNF	(1ll << 14)	/* 1 = deliver interrupts via FSB instead of APIC */
 #define TN_FSB_INT_DEL_CAP	(1ll << 15)	/* read only, 1 = FSB delivery available */
 
-#define HPET_INT_ENABLE 0x3	/* Sets ENABLE_CNF and LEG_RT_CNF */
+#define HPET_INT_ENABLE(n) (0x1 << n)	/* Clears the INT n for level-triggered mode. */
```
gparmer: I'm a little confused why these changes are in this patch?  Did you get them from Jakob?

src/kernel/capinv.c
phanikishoreg: My analysis: In order to modify the return path in entry.S, I needed to save the registers correctly when there is a thread switch to this thread next time. If it was pre-empted, I have to have the snapshot of all the registers, and if it wasn't, I only needed GP registers to save. This is required for making a decision of either doing a IRET or SYSEXIT. 

Also, Thread-switch (co-operative switching) has improved drastically. Before all these changes, it was 468cycles, now its around 310cycles. 

I understand that you'll need to think through the implications. Will look forward to it!

src/kernel/capinv.c
phanikishoreg: I don't think it was relevant before, because timer Interrupt handler had no decision making code for IRET/SYSEXIT and also, capinv_int_snd() was commented.

src/platform/i386/Makefile
phanikishoreg: We need -g for symbols. If you remember, we were not able to use objdump to get C code inlined with the assembly dump. So I kept both. I had initially set to only -O3, thinking that it would slow down if there are symbols but later after discussing with you, I kept both, because symbols are useful and -g won't affect the performance.

src/platform/i386/entry.S
phanikishoreg: Sure, I'll do that!

src/kernel/capinv.c
gparmer: I don't know if the return value of the function is checked anywhere, so I'm not sure if it matters either way.

src/components/implementation/tests/micro_booter/micro_booter.c
gparmer: The code should only reflect the current state of the code.  We won't remember different "possibilities" in a month.

src/components/implementation/tests/micro_booter/micro_booter.c
gparmer: if you want different types of tests, then just add conditionals.

src/platform/i386/entry.S
phanikishoreg: It won't be possible. Because, we are making a decision based on the return value of 'fn', Whether to restore a pre-empted thread or to return to a thread that was co-operatively switched-out of.

src/kernel/capinv.c
gparmer: I think you've done a great job thinking this through.  Looks fantastic.  I've come to the same conclusions as you did.

It looks *great*.

src/platform/i386/Makefile
gparmer: Interesting.  I remember talking about this, but it goes against many things that I've internalized in the past.

Thanks, you learn something every day!

src/kernel/capinv.c
phanikishoreg: Yeah I have that doubt. When I kept it as 'return 0', it was page-faulting. Anyway, I'll check further into this, this is also useful in the other comment where you suggested to just do jmp restore_from_thd. I'll check that! Thanks!

src/platform/i386/idt.c
phanikishoreg: Yeah, it was that annoying issue! Did try a few suggestions from you and online before moving on! I thought its time consuming to deviate into fixing this.

src/platform/i386/timer.c
phanikishoreg: No, I didn't speak to him. I went through the HPET Specification document and made the changes. His comments don't seem right, though it is doing what it had to. As of now, we use Periodic timer in timer[0], so it sets that bit 0 for clearing INT 0. It also sets Bit 1 but that doesn't matter as we don't use it. So it works. I modified it to set the bit that corresponds to the timer that we use.

src/kernel/capinv.c
gparmer: Hmmm, if it was faulting, it would be good to know why.

src/platform/i386/entry.S
gparmer: Can we remove the `PT_*` now that you got rid of the thread restore crap?

src/platform/i386/entry.S
```diff
@@ -72,29 +72,13 @@ PT_OLDSS	= 0x40
...
-	movl PT_EAX(%eax), %eax
+	RESTORE_REGS_GENERAL
+	popl %ds
+	popl %es
+	popl %fs
+	popl %gs 
+
+	addl $4, %esp /* page fault information */
 
 	sti
```
gparmer: `RET_TO_USER` ???

src/platform/i386/entry.S
phanikishoreg: I think so. I overlooked that cleanup part! Will do it! Thank you.

src/platform/i386/entry.S
phanikishoreg: `RET_TO_USER` does `SYSEXIT` but we need to do `IRET` here.

src/platform/i386/entry.S
gparmer: I was missing something.  You're actually doing exactly what I meant to be asking.  Again, great job.

src/platform/i386/idt.c
gparmer: OK, we'll leave it for now.

src/platform/i386/timer.c
gparmer: Damn, awesome.  Please send an email to Jakob and me updating him on these changes.

PR #171, opened 2016-01-22
Exception stubs, Timer Interrupts and Micro-benchmarking test suite

phanikishoreg:
Fixed review comments. 
And, in micro_booter.c, added BUG_DIVZERO() macro to test/demonstrate Exception stub.


src/components/implementation/tests/micro_booter/micro_booter.c
```diff
@@ -36,9 +36,11 @@ printc(char *fmt, ...)
 #define PRINT_FN prints
 #define debug_print(str) (PRINT_FN(str __FILE__ ":" STR(__LINE__) ".\n"))
 #define BUG() do { debug_print("BUG @ "); *((int *)0) = 0; } while (0);
+#define BUG_DIVZERO() do { debug_print("BUG @ "); int i = 1 / 0; } while (0);
 
 struct cos_compinfo booter_info;
 
+#define ITERATIONS 1000000
```
gparmer: No change necessary here.  However, keep in mind in the future that if you're going to be dividing, it is always beneficial to do so with a power of two (so that you can use shifts instead of divide instructions).  1<<20 would work here.  That said, we don't care about the performance of these divides, so it doesn't matter.  However, this is a good habit to get into.

src/components/implementation/tests/micro_booter/micro_booter.c
```diff
@@ -57,6 +59,41 @@ tls_set(size_t off, unsigned long val)
...
+	long long total_swt_cycles = 0LL;
+	long long start_swt_cycles = 0LL, end_swt_cycles = 0LL;
+	long long iters = 0LL;
+
+	ts = cos_thd_alloc(&booter_info, booter_info.comp_cap, thd_fn_perf, (void *)i);
+	assert(ts);
+	cos_thd_switch(ts);
+
+	rdtscll(start_swt_cycles);
+	for (iters = 0LL; iters < ITERATIONS; iters ++) {
```
gparmer: No change required.  Minor style point.  Should be " ; " not "; ".

PR #173, opened 2016-02-21
Hardware interface with Asnd -> Rcv capabilities

phanikishoreg:
* Incorporated all the review comments you provided me in my local repo @ https://github.com/phanikishoreg/Composite/commit/2ca55a281ae97152dedb3ef70920a574913ec32f. (Thank you!)
* FIXME:  ATTACH and DETACH API need synchronization. 
* Clean up for previous commits (div-by-zero warning, etc)


src/kernel/capinv.c
```diff
@@ -552,33 +553,48 @@ cap_asnd_op(struct cap_asnd *asnd, struct thread *thd, struct pt_regs *regs,
...
 
 int
-capinv_int_snd(struct thread *rcv_thd, struct pt_regs *regs)
+cap_hw_asnd(struct cap_asnd *asnd, struct pt_regs *regs)
 {
-	struct comp_info *ci;
-	struct thread *thd, *next;
-	struct tcap *tcap, *rcv_tcap;
+	int restore_curr_thd = 1;
+	if (asnd->h.type != CAP_ASND) return restore_curr_thd;
```
gparmer: Don't interleave code and variable definitions.  Thought we went over this before.

src/kernel/include/hw.h
```diff
@@ -0,0 +1,73 @@
...
+	return asnd_construct(&hw_asnd_caps[hwid], rcvc, rcv_cap, 0, 0);
+}
+
+static int
+hw_detach_rcvcap(struct cap_hw *hwc, hwid_t hwid)
+{
+	if (hwid < HW_IRQ_EXTERNAL_MIN || hwid > HW_IRQ_EXTERNAL_MAX) return -EINVAL;
+	if (!(hwc->hw_bitmap & (1 << (hwid - HW_IRQ_EXTERNAL_MIN)))) return -EINVAL;
+
+	/* FIXME: Need to synchronize using __xx_pre and __xx_post perhaps in asnd_deconstruct() */
```
gparmer: I don't think this fits onto a single line.  If that's the case then comments should use:

```
/*
 * content here
 */
```

src/platform/i386/exception.c
```diff
@@ -173,7 +173,7 @@ int
 machine_check_abort_handler(struct pt_regs *regs)
 {
 	print_regs_state(regs);
-	die("FAULT: Machine Check\n");
+	die("ABORT: Machine Check\n");
```
gparmer: good catch

src/platform/i386/idt.c
```diff
@@ -63,6 +63,19 @@ idt_set_gate(u8_t num, u32_t base, u16_t sel, u8_t flags)
 	// The OR is used for ring once we get usermode up and running
 }
 
+int
+hw_handler(struct pt_regs *regs)
+{
+	int preempt = 1;
+	printk("%d", regs->orig_ax);
+
+	/* TODO: ack here? or after user-level interrupt(rcv event) processing? */
```
gparmer: Same comment as before about this comment.

src/components/lib/cos_kernel_api.c
```diff
@@ -509,6 +509,21 @@ cos_asnd_alloc(struct cos_compinfo *ci, arcvcap_t arcvcap, captblcap_t ctcap)
 	return cap;
 }
 
+
+hwcap_t
```
gparmer: maybe a TODO here concerning the fact that this breaks access control, and in the future we need to make it so that you can only crate a hwcap from an existing one.

src/kernel/include/inv.h
```diff
@@ -90,6 +90,26 @@ sret_deactivate(struct cap_captbl *t, capid_t capin, livenessid_t lid)
 { return cap_capdeactivate(t, capin, CAP_SRET, lid); }
 
 static int
+asnd_construct(struct cap_asnd *asndc, struct cap_arcv *arcvc, capid_t rcv_cap, u32_t budget, u32_t period)
```
gparmer: Did you ignore my previous comment about using this within the normal `asnd_activate` routine, or am I missing where it is used here?  This is code duplication right?  We don't like that if I'm right.

src/kernel/capinv.c
phanikishoreg: Sorry, I was thinking returning immediately in this case will save time. I'll move it back, weird judgements I make I believe. ;)

src/kernel/include/hw.h
phanikishoreg: Okay! I'll make multi-line comment.

src/components/lib/cos_kernel_api.c
phanikishoreg: sure. i'll add that.

src/kernel/include/inv.h
phanikishoreg: We decided to keep it as FIXME because we cannot have captbl in this case to call asnd_activate.

gparmer: there might have been a miscommunication.  The goal is to avoid replication where possible.  I don't see a good reason why we can't do that here.  *After* we resolve the `struct cap_asnd` in `asnd_activate`, it should be able to call this (or something like this), right?  It is literally a copy and paste so I don't understand your justification.

You need to think through ways to make sure that you're creating maintainable, reasonable, code.

phanikishoreg: I still am a bit confused right now. There are 2 scenario here to call `asnd_construct`.
1. For `hw_attach_rcvcap`: In this case we cannot use `asnd_activate` so we call `asnd_construct` directly. I guess this is Okay for now. (Except for synchronization fixes that it would require.)
2. From `asnd_activate` for `cos_asnd_alloc` API: If we could just call `asnd_construct` from here, we could avoid code duplication. Its a big mistake, I should have considered it! Sorry.. And I guess you're referring to this case in your previous comments? 

That was the confusion, all this while I was thinking it was the first case you were referring to and I wasn't sure how to call `asnd_activate` from `hw_attach_rcvcap`. Now if I got it right, your comment isn't for that case, its for 2nd one. And if that's the case, yes, I'll change it and I'll keep in mind for any such things I do in future.

PR #180, opened 2016-05-27
Grub2 Integration for Bare metal boot 

phanikishoreg:
### Summary of this PR

Added support to create Bootable ISO with Grub2 bootloader to be able to boot Composite the easy way from USB or CD-ROM.

### Code Quality

Style:

- [x] All other aspects of the SG are adhered to, or exceptions are justified in this pull request: **Mostly scripting**

Code Craftsmanship:


gparmer:
- Is the old form of booting from qemu still supported?
- Is there any way to automate more of this?  The "note: for this to work" section seems pretty glitchy.

phanikishoreg:
1. Yes, it is supported. 
2. It is automated, except for flashing it to USB and rebooting the PC is not. 
    That note is for the second section which is mainly information for people who would want to add booting Composite into their Host(Ubuntu or any linux) Grub and have composite files on DISK instead of booting from USB. I don't think we need to automate that, right?

PR #187, opened 2016-07-28
bug fix: remove redundant liveness_tbl, undo refcnt and fix __captbl_…

ryuxin:
### Summary of this PR

bug fix: remove redundant liveness_tbl, undo refcnt and fix __captbl_getleaf

### Code Quality

Style:

Code Craftsmanship:

### DIFF
--- a/src/kernel/include/component.h
+++ b/src/kernel/include/component.h
@@ -48,18 +48,14 @@ comp_activate(struct captbl *t, capid_t cap, capid_t capin, capid_t captbl_cap, 
 	if (cos_cas((unsigned long *)&ptc->refcnt_flags, v, v + 1) != CAS_SUCCESS) return -ECASFAIL;
 
 	v = ctc->refcnt_flags;
-	if (v & CAP_MEM_FROZEN_FLAG) {
-		ret = -EINVAL;
-		goto undo_ptc;
-	}
+	if (v & CAP_MEM_FROZEN_FLAG) cos_throw(undo_ptc, -EINVAL);
 	if (cos_cas((unsigned long *)&ctc->refcnt_flags, v, v + 1) != CAS_SUCCESS) {
 		/* undo before return */
-		ret = -ECASFAIL;
-		goto undo_ptc;
+		cos_throw(undo_ptc, -ECASFAIL);
 	}
 	
 	compc = (struct cap_comp *)__cap_capactivate_pre(t, cap, capin, CAP_COMP, &ret);
-	if (!compc) goto undo_ctc;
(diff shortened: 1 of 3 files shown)


src/kernel/include/component.h
```diff
@@ -48,15 +48,18 @@ comp_activate(struct captbl *t, capid_t cap, capid_t capin, capid_t captbl_cap,
...
-	if (v & CAP_MEM_FROZEN_FLAG) return -EINVAL;
+	if (v & CAP_MEM_FROZEN_FLAG) {
+		ret = -EINVAL;
+		goto undo_ptc;
+	}
 	if (cos_cas((unsigned long *)&ctc->refcnt_flags, v, v + 1) != CAS_SUCCESS) {
 		/* undo before return */
-		cos_faa((int *)&ptc->refcnt_flags, -1);
-		return -ECASFAIL;
+		ret = -ECASFAIL;
```
gparmer: I'm pretty sure this should use `cos_throw` instead.

src/kernel/include/component.h
```diff
@@ -48,15 +48,18 @@ comp_activate(struct captbl *t, capid_t cap, capid_t capin, capid_t captbl_cap,
 	if (cos_cas((unsigned long *)&ptc->refcnt_flags, v, v + 1) != CAS_SUCCESS) return -ECASFAIL;
 
 	v = ctc->refcnt_flags;
-	if (v & CAP_MEM_FROZEN_FLAG) return -EINVAL;
+	if (v & CAP_MEM_FROZEN_FLAG) {
+		ret = -EINVAL;
```
gparmer: `cos_throw`

src/kernel/include/component.h
```diff
@@ -48,15 +48,18 @@ comp_activate(struct captbl *t, capid_t cap, capid_t capin, capid_t captbl_cap,
...
 		/* undo before return */
-		cos_faa((int *)&ptc->refcnt_flags, -1);
-		return -ECASFAIL;
+		ret = -ECASFAIL;
+		goto undo_ptc;
 	}
 	
 	compc = (struct cap_comp *)__cap_capactivate_pre(t, cap, capin, CAP_COMP, &ret);
-	if (!compc) return ret;
+	if (!compc) goto undo_ctc;
```
gparmer: I'd even use `cos_throw(ret, undo_ctc)` here even though it feels wasteful...it makes it more explicit where ret is getting updated.  I do wonder if the resulting `ret = ret;` will cause a compiler warning, in which case we should not do it.

ryuxin: Thanks. I changed to use cos_throw. All of them work.

PR #188, opened 2016-08-01
add a new page table at a fixed offset to manage all system local memory

ryuxin:
### Summary of this PR
add a new page table at a fixed offset to manage all system local memory.
This PR is only for code review.
  - confirm my understanding of using another page table to manage system memory is correct
  - confirm I correctly use to api to manipulate cap/page table.
Although the micro booter works, I think I should have bugs.
mirco booters uses many system calls which requires a cap to a page table. I am not sure which cap_pgtbl to pass, a cap to the page table used for address translation or the page table manage system memory.
I modify some, not all  syscalls to use the new page table.

### Code Quality

Style:

Code Craftsmanship:

### DIFF
--- a/src/components/include/cos_kernel_api.h
+++ b/src/components/include/cos_kernel_api.h
@@ -30,7 +30,7 @@ struct cos_meminfo {
 /* Component captbl/pgtbl allocation information */
 struct cos_compinfo {
 	/* capabilities to higher-order capability tables (or -1) */
-	capid_t pgtbl_cap, captbl_cap, comp_cap;
+	capid_t pgtbl_cap, captbl_cap, comp_cap, local_pgtbl_cap;
 	/* the frontier of unallocated caps, and the allocated captbl range */
 	capid_t cap_frontier, caprange_frontier;
 	/* the frontier for each of the various sizes of capability */
@@ -42,7 +42,7 @@ struct cos_compinfo {
 	struct cos_meminfo mi;	     /* only populated for the component with real memory */
 };
 
-void cos_compinfo_init(struct cos_compinfo *ci, captblcap_t pgtbl_cap, pgtblcap_t captbl_cap, compcap_t comp_cap, vaddr_t heap_ptr, capid_t cap_frontier, struct cos_compinfo *ci_resources);
+void cos_compinfo_init(struct cos_compinfo *ci, pgtblcap_t pgtbl_cap, captblcap_t captbl_cap, pgtblcap_t local_pgtbl_cap, compcap_t comp_cap, vaddr_t heap_ptr, capid_t cap_frontier, struct cos_compinfo *ci_resources);
 /*
  * This only needs be called on compinfos that are managing resources
  * (i.e. likely only one).  All of the capabilities will be relative
--- a/src/components/lib/cos_kernel_api.c
+++ b/src/components/lib/cos_kernel_api.c
@@ -40,9 +40,9 @@ __compinfo_metacap(struct cos_compinfo *ci)
 { return ci->memsrc; }
 
 void
-cos_compinfo_init(struct cos_compinfo *ci, captblcap_t pgtbl_cap, pgtblcap_t captbl_cap,
-		  compcap_t comp_cap, vaddr_t heap_ptr, capid_t cap_frontier,
-		  struct cos_compinfo *ci_resources)
+cos_compinfo_init(struct cos_compinfo *ci, pgtblcap_t pgtbl_cap, captblcap_t captbl_cap,
+		  pgtblcap_t local_pgtbl_cap, compcap_t comp_cap, vaddr_t heap_ptr, 
+		  capid_t cap_frontier, struct cos_compinfo *ci_resources)
 {
 	assert(ci && ci_resources);
 	assert(cap_frontier % CAPMAX_ENTRY_SZ == 0);
@@ -50,11 +50,12 @@ cos_compinfo_init(struct cos_compinfo *ci, captblcap_t pgtbl_cap, pgtblcap_t cap
 	ci->memsrc = ci_resources;
 	assert(ci_resources->memsrc == ci_resources); /* prevent infinite data-structs */
 
-	ci->pgtbl_cap    = pgtbl_cap;
-	ci->captbl_cap   = captbl_cap;
-	ci->comp_cap     = comp_cap;
-	ci->vas_frontier = heap_ptr;
-	ci->cap_frontier = cap_frontier;
+	ci->pgtbl_cap       = pgtbl_cap;
+	ci->captbl_cap      = captbl_cap;
+	ci->local_pgtbl_cap = local_pgtbl_cap;
+	ci->comp_cap        = comp_cap;
+	ci->vas_frontier    = heap_ptr;
+	ci->cap_frontier    = cap_frontier;
 	/*
 	 * The first allocation should trigger PTE allocation, unless
 	 * it is in the middle of a PGD, in which case we assume one
@@ -109,7 +110,7 @@ __mem_bump_alloc(struct cos_compinfo *__ci, int km)
 		/* are we dealing with a kernel memory allocation? */
 		syscall_op_t op = km ? CAPTBL_OP_MEM_RETYPE2KERN : CAPTBL_OP_MEM_RETYPE2USER;
 
-		if (call_cap_op(ci->pgtbl_cap, op, ret, 0, 0, 0)) return 0;
+		if (call_cap_op(ci->local_pgtbl_cap, op, ret, 0, 0, 0)) return 0;
 	}
 
 	*ptr += PAGE_SIZE;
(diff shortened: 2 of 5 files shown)


src/components/include/cos_kernel_api.h
```diff
@@ -30,7 +30,7 @@ struct cos_meminfo {
 /* Component captbl/pgtbl allocation information */
 struct cos_compinfo {
 	/* capabilities to higher-order capability tables (or -1) */
-	capid_t pgtbl_cap, captbl_cap, comp_cap;
+	capid_t pgtbl_cap, captbl_cap, comp_cap, local_pgtbl_cap;
```
gparmer: I would think that `pgtbl_cap` will still be the local one.  `global_pgtbl_cap` should be the new one.  I'm just thinking that the common case shared with other parts of the system will still be the normal local operations.  We want to use the global ones intentionally, so they should be the exceptional case.  What do you think?

src/components/include/cos_kernel_api.h
```diff
@@ -42,7 +42,7 @@ struct cos_compinfo {
 	struct cos_meminfo mi;	     /* only populated for the component with real memory */
 };
 
-void cos_compinfo_init(struct cos_compinfo *ci, captblcap_t pgtbl_cap, pgtblcap_t captbl_cap, compcap_t comp_cap, vaddr_t heap_ptr, capid_t cap_frontier, struct cos_compinfo *ci_resources);
+void cos_compinfo_init(struct cos_compinfo *ci, pgtblcap_t pgtbl_cap, captblcap_t captbl_cap, pgtblcap_t local_pgtbl_cap, compcap_t comp_cap, vaddr_t heap_ptr, capid_t cap_frontier, struct cos_compinfo *ci_resources);
```
gparmer: wow, this mistyping is horrible.  Thanks for catching it.  See above for local vs. global.

src/components/lib/cos_kernel_api.c
```diff
@@ -298,7 +299,7 @@ __page_bump_valloc(struct cos_compinfo *ci)
 
 		/* PTE */
 		if (call_cap_op(meta->captbl_cap, CAPTBL_OP_PGTBLACTIVATE,
-				pte_cap, meta->pgtbl_cap, ptemem_cap, 1)) {
+				pte_cap, meta->local_pgtbl_cap, ptemem_cap, 1)) {
```
gparmer: If we keep the local as `pgtbl_cap`, then we don't have all of these changes.  Shows that keeping local as default is likely useful.

src/components/lib/cos_kernel_api.c
```diff
@@ -476,7 +477,7 @@ cos_compinfo_alloc(struct cos_compinfo *ci, vaddr_t heap_ptr, vaddr_t entry,
 	compc = cos_comp_alloc(ci_resources, ctc, ptc, entry);
 	assert(compc);
 
-	cos_compinfo_init(ci, ptc, ctc, compc, heap_ptr, 0, ci_resources);
+	cos_compinfo_init(ci, ptc, ctc, ptc, compc, heap_ptr, 0, ci_resources);
```
gparmer: I'm starting to wonder if the right way to implement this is *not* to include a separate pgtbl in `cos_compinfo`, and rather to just set up a `cos_compinfo` that uses entirely global resources (the superblock should provide these, right?).  If we want a component to use local, we set the `cos_meminfo` to be the default `cos_compinfo`.  If we want a component to use global, we set `cos_meminfo` in it to the global `cos_compinfo`.

This would make it hard to use a combination of local and global, but that could be done using multiple `cos_compinfo`s to represent different portions of the memory and capability namespaces.

src/kernel/include/shared/cos_types.h
```diff
@@ -235,16 +236,17 @@ static inline unsigned long captbl_idsize(cap_t c)
...
-	BOOT_CAPTBL_KM_PTE     = 16,
-
-	BOOT_CAPTBL_COMP0_CT           = 18,
-	BOOT_CAPTBL_COMP0_PT           = 20,
+	BOOT_CAPTBL_SRET          = 0,
+	BOOT_CAPTBL_SELF_CT       = 4,
+	BOOT_CAPTBL_SELF_PT       = 6,
+	BOOT_CAPTBL_SELF_COMP     = 8,
+	BOOT_CAPTBL_BOOTVM_PTE    = 12,
+	BOOT_CAPTBL_SELF_LOCAL_PT = 14,
```
gparmer: did you properly take into account all of the offsets for different sized capabilities here?

src/platform/i386/boot_comp.c
```diff
@@ -155,8 +158,12 @@ kern_boot_comp(void)
 	 * Need to account for the pages that will be allocated as
 	 * PTEs
 	 */
+	local_pgd = (pgtbl_t)mem_boot_alloc(1);
+	local_pgd = (pgtbl_t)chal_va2pa(local_pgd);
+        if (pgtbl_activate(ct, BOOT_CAPTBL_SELF_CT, BOOT_CAPTBL_SELF_LOCAL_PT, local_pgd, 0)) assert(0);
```
gparmer: spaces here instead of tabs

src/platform/i386/boot_comp.c
```diff
@@ -155,8 +158,12 @@ kern_boot_comp(void)
 	 * Need to account for the pages that will be allocated as
 	 * PTEs
 	 */
+	local_pgd = (pgtbl_t)mem_boot_alloc(1);
+	local_pgd = (pgtbl_t)chal_va2pa(local_pgd);
```
gparmer: simplify with `(pgtbl_t)chal_va2pa(mem_boot_alloc(1));`

src/platform/i386/boot_comp.c
```diff
@@ -144,7 +147,7 @@ kern_boot_comp(void)
 
 	printk("\tCapability table and page-table created.\n");
 
-	ret = boot_pgtbl_mappings_add(ct, pgtbl, BOOT_CAPTBL_BOOTVM_PTE, "booter VM", mem_bootc_start(),
+	ret = boot_pgtbl_mappings_add(ct, BOOT_CAPTBL_SELF_PT, BOOT_CAPTBL_BOOTVM_PTE, "booter VM", mem_bootc_start(),
```
gparmer: I don't think that we want the boot component to be in global memory.  I think that we want it to have access to global memory through a captbl and pgtbl.  The `_SELF_PT` cap is the offset of the boot component's page-table.  Not of only a "container" of resources.

So we want to keep the self pt the same as it was as we want it in local memory, but we want the boot component to have access to the component structure through references to separate (i.e. not the component's pgtbl) references  to a pgtbl and captbl.

src/components/include/cos_kernel_api.h
interwq: I'm probably a bit rusty here -- why we need a global_pgtbl_cap in the compinfo? What are the valid use cases? Bear with me if I'm asking stupid questions.

src/kernel/include/shared/cos_types.h
interwq: Wow this reminds me of a lot of (ugly) hacking. Thanks for making it better!

interwq: Well I thought this was simplified. Seems not the case. Agree that we should have sanity checks around them.

src/components/include/cos_kernel_api.h
ryuxin: Yes, I agree. If local memory is a common case, it is better so keep current local memory management code unchanged. As you said below, if we only have one new page table for global memory, we can avoid all changes I made for local memory.
But I do not think we want to merge the code for global memory into the mainstream, right?
I can modify the code in my own branch.

src/components/lib/cos_kernel_api.c
ryuxin: agreed.

src/components/lib/cos_kernel_api.c
ryuxin: I think this is a cool idea. This provides a more unified way to handle different type of memory.
I will try to implement this to see if I have unexpected issues.

src/kernel/include/shared/cos_types.h
ryuxin: Yes, Phani explained the offset and different capability size to me.

src/components/include/cos_kernel_api.h
gparmer: @interwq: this is to support persistent memory that is shared between non-coherent nodes.  We're porting the kernel's quiescence mechanisms to handle cache non-coherence.

@ryuxin: we do want to merge it, but we want it to likely be unused in the common configuration.  I don't think that the vkernel work wants it, and we don't want to waste the resources on the pgtbl if we don't need to.

I'd imagine we'll have the slot reserved in the captbl for captbl/pgtbl.  We'll have a specific `#define` in `cos_config.h` to activate the kernel parts that build the structures, and you'll write a specific booter that uses the global stuff.

src/kernel/include/shared/cos_types.h
gparmer: @interwq: most of the `cos_kernel_api` simplifies the kernel interface.  It provides what look like real system calls, and it takes care of all of the tedious memory/resource table allocations transparently using bump pointers (so, it is allocation-only).  The hard-coding of the capability layout won't likely change.  I realized that we are actually checking to see if the capability layout is valid already:  by creating the capabilities, and checking the return value, we're essentially already sanity checking the layout.

src/platform/i386/boot_comp.c
gparmer: I should be more clear:  There are spaces here instead of tabs.  There likely should not be.

PR #189, opened 2016-08-05
use separate page tables to manage different type of memory

ryuxin:
### Summary of this PR

support multiple page tables to manage different type of memory.
current implementation uses a separate page tables to manage local memory.
modify cos_meminfo struct to include a cap to page table which provides memory.
some explanation to the page table capability.
meta (struct cos_compinfo \*memsrc) maintains the manager component's information, the pgtbl_cap in meta is the capability to the manager's own page table.
The pgtbl_cap in **local_mi** is the capability to a page table which tracks memory resource, currently it tracks all local memory.
**ci** maintains the information of a random client component. The pgtbl_cap in **ci** a capability to client's page table, and that capability is **within manager's (meta) capability table.**
The same logic applies to the captbl_cap

### Code Quality

Style:

Code Craftsmanship:

### DIFF
--- a/src/components/implementation/tests/micro_booter/micro_booter.c
+++ b/src/components/implementation/tests/micro_booter/micro_booter.c
@@ -568,7 +568,7 @@ term_fn(void *d)
 void
 cos_init(void)
 {
-	cos_meminfo_init(&booter_info.mi, BOOT_MEM_KM_BASE, COS_MEM_KERN_PA_SZ);
+	cos_meminfo_init(&booter_info.local_mi, BOOT_MEM_KM_BASE, COS_MEM_KERN_PA_SZ, BOOT_CAPTBL_SELF_LOCAL_PT);
 	cos_compinfo_init(&booter_info, BOOT_CAPTBL_SELF_PT, BOOT_CAPTBL_SELF_CT, BOOT_CAPTBL_SELF_COMP,
 			  (vaddr_t)cos_get_heap_ptr(), BOOT_CAPTBL_FREE, &booter_info);
 
--- a/src/components/include/cos_kernel_api.h
+++ b/src/components/include/cos_kernel_api.h
@@ -25,6 +25,7 @@ typedef capid_t hwcap_t;
 struct cos_meminfo {
 	vaddr_t untyped_ptr,      umem_ptr,      kmem_ptr;
 	vaddr_t untyped_frontier, umem_frontier, kmem_frontier;
+	capid_t pgtbl_cap;
 };
 
 /* Component captbl/pgtbl allocation information */
@@ -39,16 +40,16 @@ struct cos_compinfo {
 	vaddr_t vas_frontier, vasrange_frontier;
 	/* the source of memory */
 	struct cos_compinfo *memsrc; /* might be self-referential */
-	struct cos_meminfo mi;	     /* only populated for the component with real memory */
+	struct cos_meminfo local_mi;	     /* only populated for the component with real memory */
 };
 
-void cos_compinfo_init(struct cos_compinfo *ci, captblcap_t pgtbl_cap, pgtblcap_t captbl_cap, compcap_t comp_cap, vaddr_t heap_ptr, capid_t cap_frontier, struct cos_compinfo *ci_resources);
+void cos_compinfo_init(struct cos_compinfo *ci, pgtblcap_t pgtbl_cap, captblcap_t captbl_cap, compcap_t comp_cap, vaddr_t heap_ptr, capid_t cap_frontier, struct cos_compinfo *ci_resources);
 /*
  * This only needs be called on compinfos that are managing resources
  * (i.e. likely only one).  All of the capabilities will be relative
  * to this component's captbls.
  */
-void cos_meminfo_init(struct cos_meminfo *mi, vaddr_t untyped_ptr, unsigned long untyped_sz);
+void cos_meminfo_init(struct cos_meminfo *mi, vaddr_t untyped_ptr, unsigned long untyped_sz, pgtblcap_t pgtbl_cap);
 
 /*
  * This uses the next three functions to allocate a new component and
--- a/src/components/lib/cos_kernel_api.c
+++ b/src/components/lib/cos_kernel_api.c
@@ -29,10 +29,11 @@ printd(char *fmt, ...)
 #endif
 
 void
-cos_meminfo_init(struct cos_meminfo *mi, vaddr_t untyped_ptr, unsigned long untyped_sz)
+cos_meminfo_init(struct cos_meminfo *mi, vaddr_t untyped_ptr, unsigned long untyped_sz, pgtblcap_t pgtbl_cap)
 {
 	mi->untyped_ptr = mi->umem_ptr = mi->kmem_ptr = mi->umem_frontier = mi->kmem_frontier = untyped_ptr;
 	mi->untyped_frontier = untyped_ptr + untyped_sz;
+	mi->pgtbl_cap = pgtbl_cap;
 }
 
 static inline struct cos_compinfo *
@@ -40,7 +41,7 @@ __compinfo_metacap(struct cos_compinfo *ci)
 { return ci->memsrc; }
 
 void
-cos_compinfo_init(struct cos_compinfo *ci, captblcap_t pgtbl_cap, pgtblcap_t captbl_cap,
+cos_compinfo_init(struct cos_compinfo *ci, pgtblcap_t pgtbl_cap, captblcap_t captbl_cap,
 		  compcap_t comp_cap, vaddr_t heap_ptr, capid_t cap_frontier,
 		  struct cos_compinfo *ci_resources)
 {
@@ -90,17 +91,17 @@ __mem_bump_alloc(struct cos_compinfo *__ci, int km)
 	assert(ci && ci == __compinfo_metacap(__ci));
 
 	if (km) {
-		ptr      = &ci->mi.kmem_ptr;
-		frontier = &ci->mi.kmem_frontier;
+		ptr      = &ci->local_mi.kmem_ptr;
+		frontier = &ci->local_mi.kmem_frontier;
 	} else {
-		ptr      = &ci->mi.umem_ptr;
-		frontier = &ci->mi.umem_frontier;
+		ptr      = &ci->local_mi.umem_ptr;
+		frontier = &ci->local_mi.umem_frontier;
 	}
 	if (*ptr == *frontier) {
 		/* TODO: expand frontier if introspection says there is more memory */
-		if (ci->mi.untyped_ptr == ci->mi.untyped_frontier) return 0;
-		ret                 = ci->mi.untyped_ptr;
-		ci->mi.untyped_ptr += RETYPE_MEM_SIZE; /* TODO: atomic */
+		if (ci->local_mi.untyped_ptr == ci->local_mi.untyped_frontier) return 0;
+		ret                 = ci->local_mi.untyped_ptr;
+		ci->local_mi.untyped_ptr += RETYPE_MEM_SIZE; /* TODO: atomic */
 		*ptr                = ret;
 		*frontier           = ret + RETYPE_MEM_SIZE;
 	}
@@ -109,7 +110,7 @@ __mem_bump_alloc(struct cos_compinfo *__ci, int km)
 		/* are we dealing with a kernel memory allocation? */
 		syscall_op_t op = km ? CAPTBL_OP_MEM_RETYPE2KERN : CAPTBL_OP_MEM_RETYPE2USER;
 
-		if (call_cap_op(ci->pgtbl_cap, op, ret, 0, 0, 0)) return 0;
+		if (call_cap_op(ci->local_mi.pgtbl_cap, op, ret, 0, 0, 0)) return 0;
 	}
 
 	*ptr += PAGE_SIZE;
@@ -200,7 +201,7 @@ __capid_captbl_check_expand(struct cos_compinfo *ci)
 
 	printd("__capid_captbl_check_expand->pre-captblactivate (%d)\n", CAPTBL_OP_CAPTBLACTIVATE);
 	/* captbl internal node allocated with the resource provider's captbls */
-	if (call_cap_op(meta->captbl_cap, CAPTBL_OP_CAPTBLACTIVATE, captblcap, meta->pgtbl_cap, kmem, 1)) {
+	if (call_cap_op(meta->captbl_cap, CAPTBL_OP_CAPTBLACTIVATE, captblcap, meta->local_mi.pgtbl_cap, kmem, 1)) {
 		assert(0); /* race condition? */
 		return -1;
 	}
@@ -298,7 +299,7 @@ __page_bump_valloc(struct cos_compinfo *ci)
 
 		/* PTE */
 		if (call_cap_op(meta->captbl_cap, CAPTBL_OP_PGTBLACTIVATE,
-				pte_cap, meta->pgtbl_cap, ptemem_cap, 1)) {
+				pte_cap, meta->local_mi.pgtbl_cap, ptemem_cap, 1)) {
 			assert(0); /* race? */
 			return 0;
 		}
@@ -338,7 +339,7 @@ __page_bump_alloc(struct cos_compinfo *ci)
 	if (!umem) return 0;
 
 	/* Actually map in the memory. FIXME: cleanup! */
-	if (call_cap_op(meta->pgtbl_cap, CAPTBL_OP_MEMACTIVATE, umem,
+	if (call_cap_op(meta->local_mi.pgtbl_cap, CAPTBL_OP_MEMACTIVATE, umem,
 			ci->pgtbl_cap, heap_vaddr, 0)) {
 		assert(0);
 		return 0;
@@ -386,7 +387,7 @@ __cos_thd_alloc(struct cos_compinfo *ci, compcap_t comp, int init_data)
 	if (__alloc_mem_cap(ci, CAP_THD, &kmem, &cap)) return 0;
 	assert((size_t)init_data < sizeof(u16_t)*8);
 	/* TODO: Add cap size checking */
-	if (call_cap_op(ci->captbl_cap, CAPTBL_OP_THDACTIVATE, (init_data << 16) | cap, ci->pgtbl_cap, kmem, comp)) BUG();
+	if (call_cap_op(ci->captbl_cap, CAPTBL_OP_THDACTIVATE, (init_data << 16) | cap, __compinfo_metacap(ci)->local_mi.pgtbl_cap, kmem, comp)) BUG();
 
 	return cap;
 }
@@ -421,7 +422,7 @@ cos_captbl_alloc(struct cos_compinfo *ci)
 	assert(ci);
 
 	if (__alloc_mem_cap(ci, CAP_CAPTBL, &kmem, &cap)) return 0;
-	if (call_cap_op(ci->captbl_cap, CAPTBL_OP_CAPTBLACTIVATE, cap, ci->pgtbl_cap, kmem, 0)) BUG();
+	if (call_cap_op(ci->captbl_cap, CAPTBL_OP_CAPTBLACTIVATE, cap, __compinfo_metacap(ci)->local_mi.pgtbl_cap, kmem, 0)) BUG();
 
 	return cap;
 }
@@ -437,7 +438,7 @@ cos_pgtbl_alloc(struct cos_compinfo *ci)
 	assert(ci);
 
 	if (__alloc_mem_cap(ci, CAP_PGTBL, &kmem, &cap)) return 0;
-	if (call_cap_op(ci->captbl_cap, CAPTBL_OP_PGTBLACTIVATE, cap, ci->pgtbl_cap, kmem, 0))  BUG();
+	if (call_cap_op(ci->captbl_cap, CAPTBL_OP_PGTBLACTIVATE, cap, __compinfo_metacap(ci)->local_mi.pgtbl_cap, kmem, 0))  BUG();
 
 	return cap;
 }
@@ -647,7 +648,7 @@ cos_tcap_alloc(struct cos_compinfo *ci, tcap_prio_t prio)
 
 	if (__alloc_mem_cap(ci, CAP_TCAP, &kmem, &cap)) return 0;
 	/* TODO: Add cap size checking */
-	if (call_cap_op(ci->captbl_cap, CAPTBL_OP_TCAP_ACTIVATE, (cap << 16) | ci->pgtbl_cap, kmem, prio_hi, prio_lo)) BUG();
+	if (call_cap_op(ci->captbl_cap, CAPTBL_OP_TCAP_ACTIVATE, (cap << 16) | __compinfo_metacap(ci)->local_mi.pgtbl_cap, kmem, prio_hi, prio_lo)) BUG();
 
 	return cap;
 }
--- a/src/kernel/include/shared/cos_types.h
+++ b/src/kernel/include/shared/cos_types.h
@@ -226,6 +226,7 @@ static inline unsigned long captbl_idsize(cap_t c)
  * 16-17 = km pte
  * 18-19 = comp0 captbl,
  * 20-21 = comp0 pgtbl root,
+ * 22-23 = local memory pgtbl root
  * 24-27 = comp0 component,
  * 28~(20+2*NCPU) = per core alpha thd
  *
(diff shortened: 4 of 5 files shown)


src/components/implementation/tests/micro_booter/micro_booter.c
```diff
@@ -568,7 +568,7 @@ term_fn(void *d)
 void
 cos_init(void)
 {
-	cos_meminfo_init(&booter_info.mi, BOOT_MEM_KM_BASE, COS_MEM_KERN_PA_SZ);
+	cos_meminfo_init(&booter_info.local_mi, BOOT_MEM_KM_BASE, COS_MEM_KERN_PA_SZ, BOOT_CAPTBL_SELF_LOCAL_PT);
```
gparmer: I thought we discussed making local default, and "global" the one that causes changes of variable names.  Local is the one that is typical.  Global is the one that is aberrant.  Why would we change all of the code to make local the one that (from a variable name-space perspective) is special cased?

src/components/include/cos_kernel_api.h
```diff
@@ -39,16 +40,16 @@ struct cos_compinfo {
 	vaddr_t vas_frontier, vasrange_frontier;
 	/* the source of memory */
 	struct cos_compinfo *memsrc; /* might be self-referential */
-	struct cos_meminfo mi;	     /* only populated for the component with real memory */
+	struct cos_meminfo local_mi;	     /* only populated for the component with real memory */
```
gparmer: Or do you mean something else by "local" here?

src/kernel/include/shared/cos_types.h
```diff
@@ -235,16 +236,17 @@ static inline unsigned long captbl_idsize(cap_t c)
...
-	BOOT_CAPTBL_KM_PTE     = 16,
-
-	BOOT_CAPTBL_COMP0_CT           = 18,
-	BOOT_CAPTBL_COMP0_PT           = 20,
+	BOOT_CAPTBL_SRET          = 0,
+	BOOT_CAPTBL_SELF_CT       = 4,
+	BOOT_CAPTBL_SELF_PT       = 6,
+	BOOT_CAPTBL_SELF_COMP     = 8,
+	BOOT_CAPTBL_BOOTVM_PTE    = 12,
+	BOOT_CAPTBL_SELF_LOCAL_PT = 14,
```
gparmer: shouldn't SELF_PT be local, and SELF_GLOBAL_PT be global?  Didn't we already talk about this?

If some of the comments don't make sense, ask.  I don't like making the same comments multiple times.

src/components/lib/cos_kernel_api.c
```diff
@@ -421,7 +422,7 @@ cos_captbl_alloc(struct cos_compinfo *ci)
 	assert(ci);
 
 	if (__alloc_mem_cap(ci, CAP_CAPTBL, &kmem, &cap)) return 0;
-	if (call_cap_op(ci->captbl_cap, CAPTBL_OP_CAPTBLACTIVATE, cap, ci->pgtbl_cap, kmem, 0)) BUG();
+	if (call_cap_op(ci->captbl_cap, CAPTBL_OP_CAPTBLACTIVATE, cap, __compinfo_metacap(ci)->local_mi.pgtbl_cap, kmem, 0)) BUG();
```
gparmer: I'm not sure how this worked before as well.  Perhaps @phani didn't use this with a meta.

src/platform/i386/boot_comp.c
```diff
@@ -15,14 +15,17 @@ extern u8_t *boot_comp_pgd;
 int boot_nptes(unsigned int sz) { return round_up_to_pow2(sz, PGD_RANGE)/PGD_RANGE; }
 
 int
-boot_pgtbl_mappings_add(struct captbl *ct, pgtbl_t pgtbl, capid_t ptecap, const char *label,
+boot_pgtbl_mappings_add(struct captbl *ct, capid_t pgdcap, capid_t ptecap, const char *label,
```
gparmer: why remove the type?

src/platform/i386/boot_comp.c
```diff
@@ -15,14 +15,17 @@ extern u8_t *boot_comp_pgd;
...
 			void *kern_vaddr, unsigned long user_vaddr, unsigned int range, int uvm)
 {
 	int ret;
 	u8_t *ptes;
 	unsigned int nptes = 0, i;
-	struct cap_pgtbl *pte_cap;
+	struct cap_pgtbl *pte_cap, *pgd_cap;
+	pgtbl_t pgtbl;
 
+	pgd_cap = (struct cap_pgtbl*)captbl_lkup(ct, pgdcap);
```
gparmer: `if (CAP_TYPECHK(...)) assert(0);`

have to make sure it is of the correct type.

src/platform/i386/boot_comp.c
```diff
@@ -112,7 +115,7 @@ kern_boot_comp(void)
         struct captbl *ct;
         unsigned int i;
 	u8_t *boot_comp_captbl;
-	pgtbl_t pgtbl = (pgtbl_t)chal_va2pa(&boot_comp_pgd);
+	pgtbl_t pgtbl = (pgtbl_t)chal_va2pa(&boot_comp_pgd), local_pgd;
```
gparmer: I don't think this variable is being used at all.  Please always do a `git diff` before committing.  Clean up things like this.

src/platform/i386/boot_comp.c
```diff
@@ -144,7 +147,7 @@ kern_boot_comp(void)
 
 	printk("\tCapability table and page-table created.\n");
 
-	ret = boot_pgtbl_mappings_add(ct, pgtbl, BOOT_CAPTBL_BOOTVM_PTE, "booter VM", mem_bootc_start(),
+	ret = boot_pgtbl_mappings_add(ct, BOOT_CAPTBL_SELF_PT, BOOT_CAPTBL_BOOTVM_PTE, "booter VM", mem_bootc_start(),
```
gparmer: I'm missing why this change has been made.  Why?

Is this so that both pgtbls are mapped into the booter pt?

src/platform/i386/boot_comp.c
```diff
@@ -155,8 +158,10 @@ kern_boot_comp(void)
 	 * Need to account for the pages that will be allocated as
 	 * PTEs
 	 */
+	local_pgd = (pgtbl_t)chal_va2pa(mem_boot_alloc(1));
+	if (pgtbl_activate(ct, BOOT_CAPTBL_SELF_CT, BOOT_CAPTBL_SELF_LOCAL_PT, local_pgd, 0)) assert(0);
 	nkmemptes = boot_nptes(mem_utmem_end() - mem_boot_end());
-	ret = boot_pgtbl_mappings_add(ct, pgtbl, BOOT_CAPTBL_KM_PTE, "untyped memory", mem_boot_nalloc_end(nkmemptes),
+	ret = boot_pgtbl_mappings_add(ct, BOOT_CAPTBL_SELF_LOCAL_PT, BOOT_CAPTBL_KM_PTE, "untyped memory", mem_boot_nalloc_end(nkmemptes),
```
gparmer: Are you parameterizing which PT to use so that you can use it in the future for the global memory?

src/components/implementation/tests/micro_booter/micro_booter.c
ryuxin: Because in the future, I want to two meminfo, local_mi and global_mi. I think this explicit distinction makes the code more clear.
In addition,  it seems global memory will become hot research topic and we may heavily use global memory in the near future. I am not sure local one is still typical or not.
But if you want to keep the name unchanged, it is totally fine.

src/components/include/cos_kernel_api.h
ryuxin: no, "local" just refers to local memory.

src/kernel/include/shared/cos_types.h
ryuxin: We definitely discussed this before. But before we put multiple page tables in the cos_info, which makes the code a little tedious, so want to simplify the code.
But now we have to use multiple mem_info. So again, I favor explicit separation between pgtbls for address translation and resource tracking.
And using multiple page tables consumes more memory, I do not think this is the issue we really care about now.
But if we use SELF_PT both for resource tracking and address translation, this will stress the address space resource of the manager.  Think about memory manager, if it managers 1G untyped memory, it just wastes lots of PTEs for the untyped memory, which limits the size of address space the manager can use. I probably miss something here.

src/components/lib/cos_kernel_api.c
ryuxin: Yes,  @phani can help us to make sure when we use ci or meta.
I think a general rule is whenever you provide a memory capability (kmem here), you need to provide the corresponding page table who holds that capability.
I think in current code, all memory are allocated from the metta's page table.

src/platform/i386/boot_comp.c
ryuxin: Yes, this replies to all your above related comments.
I want to parameterize to which pgtbl we add mapping.
So I change the argument to use the pgtbl capability instead of physical address of the pgtbl root.

gparmer:
I'm glad that you're doing this iteratively.  This PR is mainly here to 1. put the pgtbl in the mem_info struct, and 2. use that appropriately when allocating resource.  The latter fixes where resources are allocated for a few kernel objects.

Is this accurate?  Is there more?

Your summary says that you're trying to provide multiple page tables, but I don't see that, so I'm not sure if that is part of this PR.  Is that true?

src/platform/i386/boot_comp.c
ryuxin: No, this is used for the new page table we created to track local system memory.
See code L161 in the same file.
But if we decide to only use the "SELF_PT", I can remove it.

src/components/implementation/tests/micro_booter/micro_booter.c
gparmer: 1. that means that there are ton of modifications in this PR that make it very hard to go through for almost no gain.

2. That would make your improvements very limited in their intended use.  What if there are other devices that expose physical memory but don't expose global memory?  Xeon Phis are an example.

You're providing a general facility for using multiple physical page-tables for allocation.  Why bind it one one use-case?

src/components/include/cos_kernel_api.h
gparmer: See my previous comment.  I see no reason to broadly change all of this code for little reason.

src/components/lib/cos_kernel_api.c
gparmer: I agree.  I've asked @phani to look at this to make sure.

src/kernel/include/shared/cos_types.h
gparmer: I totally missed the point of this!  Do I have this right:  You're trying to put all the untyped memory into a *separate* page table instead of in the booter's PT?  If that's the case, then you're going in exactly the right direction.  We want that anyway.  If that's the case, we might as well start mapping at address 0.  Is this understanding correct?

ryuxin:
My summary is perhaps inaccurate.
First, we try to find a approach who has the ability to support multiple page tables. Current approach is to extend mem_info to include a pgtbl_cap. If current code works, adding more page tables is trivial, as we can add more mem_info structs.
Second it uses mulitple page tables in this PR. As I have "SELF_PT"and "SELF_LOCAL_PT"

src/platform/i386/boot_comp.c
gparmer: OK, I think i get the point of all of this.

Aside from the obvious issues with the commit (types, unused local variables, etc...), I think this is OK.  I don't like using `local` in code that is supposed to be generic (i.e. not only used for *your* functionality).  In the enum of capabilities, perhaps we can use `BOOT_CAPTBL_SELF_UNTYPED_PT`?

src/kernel/include/shared/cos_types.h
ryuxin: Yes, I am exactly trying to put all the untyped memory into a separate page table (SELF_LOCAL_PT) instead of in the booter's SELF_PT.
But I did not start at address 0. I think keep this address unchanged makes it easy to debug. 
But it is necessary to change it to address 0.
After we agree on other parts, I will change it.

ryuxin: And keeping the mem_info variable name unchanged maybe better.
I think the key question we have is "do we want separate SELF_PT and SELF_LOCAL_PT"

gparmer:
Yeah, I understand where you're going with all of this now.  As I mentioned i the previous comment, `BOOT_CAPTBL_SELF_UNTYPED_PT` is perhaps more appropriate instead of `LOCAL`

src/platform/i386/boot_comp.c
ryuxin: I change the argument from pgtbl physical address to pgtbl cap
Does it make sense to you now?

src/kernel/include/shared/cos_types.h
gparmer: yes.  see my other comments.

ryuxin:
Ok, now we have those agreements.
1. keep the name of mem_info variable (mi) unchanged. This avoid a lot of code modification.
2. Use BOOT_CAPTBL_SELF_UNTYPED_PT. I think this is better. But it is still refer to a different page table from BOOT_CAPTBL_SELF_PT. The mapping in BOOT_CAPTBL_SELF_UNTYPED_PT starts at address 0.
The pending question is related to when we use "ci" and "meta".

src/platform/i386/boot_comp.c
gparmer: I don't understand what this means.  I see no physical address.  I only see a type that was made ambiguous.  I don't understand your statement as it seems trivially wrong.

gparmer:
I would appreciate if you could read your summary again and ask yourself how you could help us get to this level of understanding.  I don't have 2 hours for every pull request.  Especially one with 20 lines of changes.  It is very hard by looking at the code to really understand how it all fits together, and your current summary wasn't specific enough to really give me enough to grasp on to.

The other alternative is that I spend the 15 minutes I should, and you do another PR that I throw back at you, and we repeat N times.  That sounds much more frustrating for both of us.  Consider thinking about providing as much context as possible in the summary.  Organizing your summary will go along way toward making *other people* understand it.  Try saying "if the person reading this doesn't know anything about what this change is, how can I lead them into an understanding".

src/platform/i386/boot_comp.c
ryuxin: the type "pgtbl_t" is defined as `typedef struct pgtbl * pgtbl_t;`
More specific, we can see the code `pgtbl_t pgtbl = (pgtbl_t)chal_va2pa(&boot_comp_pgd);`
So "pgtbl_t" is the physical address of the page table.
I change the argument to use capability instead of physical address.
We extensively use "capid_t" everywhere in the kernel to refer to a capability.
Why do you say it is ambiguous?

gparmer: I see now.  Thanks for explaining.  This all makes sense now.

src/kernel/include/shared/cos_types.h
phanikishoreg: I'm missing the distinction of SELF_PT and UNTYPED_PT here, especially in case of Booter component. Isn't `pgtbl` of a component supposed to have all the pages that it has access to? How is separating it between two PT caps achieve this? Also, every time booter component or any component that has access to UNTYPED_PT, would then alloc, which will have to cpy from its UNTYPED_PT to SELF_PT to mean that it has access to its resources. 

One thing this simplifies I assume is, when we have a separate memmgr, we can just copy UNTYPED_PT cap to memmgr as opposed to copying as many pgtbl caps that it can manage resources for components, but that should not be much of a overhead if that happens only at boot-time. What am I missing here?

phanikishoreg: Also, having a separate comp_info for global memory would be much more useful because then, user can call the kernel api with different comp_info appropriately? If I need it to allocate memory in global memory, I would do `cos_page_bump_alloc(global_compinfo);` perhaps. But in the above design, wouldn't there be significant cos_kernel_api changes ?

src/components/lib/cos_kernel_api.c
```diff
@@ -109,7 +110,7 @@ __mem_bump_alloc(struct cos_compinfo *__ci, int km)
 		/* are we dealing with a kernel memory allocation? */
 		syscall_op_t op = km ? CAPTBL_OP_MEM_RETYPE2KERN : CAPTBL_OP_MEM_RETYPE2USER;
 
-		if (call_cap_op(ci->pgtbl_cap, op, ret, 0, 0, 0)) return 0;
+		if (call_cap_op(ci->local_mi.pgtbl_cap, op, ret, 0, 0, 0)) return 0;
```
phanikishoreg: Once we do RETYPE this, this component has access to that page, should that mean it has to be copied to its `SELF_PT`? I'm confused what remains in `SELF_PT` otherwise.

src/components/lib/cos_kernel_api.c
```diff
@@ -200,7 +201,7 @@ __capid_captbl_check_expand(struct cos_compinfo *ci)
 
 	printd("__capid_captbl_check_expand->pre-captblactivate (%d)\n", CAPTBL_OP_CAPTBLACTIVATE);
 	/* captbl internal node allocated with the resource provider's captbls */
-	if (call_cap_op(meta->captbl_cap, CAPTBL_OP_CAPTBLACTIVATE, captblcap, meta->pgtbl_cap, kmem, 1)) {
+	if (call_cap_op(meta->captbl_cap, CAPTBL_OP_CAPTBLACTIVATE, captblcap, meta->local_mi.pgtbl_cap, kmem, 1)) {
```
phanikishoreg: I guess, like Gabe mentioned before, `OP_CONS` which is called next must take in an additional parameter that indicates which `pgtbl/captbl` has this `captblcap`.

src/components/lib/cos_kernel_api.c
```diff
@@ -298,7 +299,7 @@ __page_bump_valloc(struct cos_compinfo *ci)
 
 		/* PTE */
 		if (call_cap_op(meta->captbl_cap, CAPTBL_OP_PGTBLACTIVATE,
-				pte_cap, meta->pgtbl_cap, ptemem_cap, 1)) {
+				pte_cap, meta->local_mi.pgtbl_cap, ptemem_cap, 1)) {
```
phanikishoreg: `OP_CONS` must know `pte_cap` is from `meta`'s PTE, right? Like in capid expand.

src/components/lib/cos_kernel_api.c
phanikishoreg: @gparmer it worked fine before because the threads were created from `vkernel` component and were copied to vm captbl, which as you said didn't use `meta`. And when the VMs started, they had their own untyped memory, so again the same no `meta` case there. 

@ryuxin this looks alright to me, but i've come to realize all the `ci->pgtbl_cap` are going to be replaced with `ci->local_mi.pgtbl_cap`, which pretty much removes `SELF_PT` from the picture. Isn't it?

src/components/lib/cos_kernel_api.c
gparmer: I think it will only include the virtual memory of the booter.

src/components/lib/cos_kernel_api.c
gparmer: still don't know if we actually want to do that.

src/components/lib/cos_kernel_api.c
gparmer: I don't understand how you're making the second deduction, @phani.  SELF_PT has to be somewhere in all of the pgtbl_caps

phanikishoreg: I mean, all the `ci->pgtbl_cap` which is `SELF_PT` seem to get replaced with `local_mi.pgtbl_cap` (`UNTYPED_PT`), so I said that. 
But your last comment about `SELF_PT` having only virtual memory of the booter, I'm trying to understand that. In that case booter and/or memmgr, should have this distinction of `SELF_PT` and `UNTYPED_PT` and all other components only refer to `SELF_PT`? I'll look at the source code again to see how all this fits.

src/kernel/include/shared/cos_types.h
ryuxin: "Isn't `pgtbl` of a component supposed to have all the pages that it has access to?"
Here "access" is a little ambiguous.
Take memory manager for example, it manages 1G memory. This does not mean the manager will actually use that 1G memory. Broadly there are two kinds of memory:
memory is used by the manager for its code data. Those memory are present in the page able referenced by SELF_PT.
other memory that is managed by the manager, but not actually accessed by the manager.
those memory are present in SELF_UNTYPED_PT.

"user can call the kernel api with different comp_info appropriately?"
This is also a little ambiguous.
A client component invokes a manager component without providing comp_info.
The manager finds the comp_info of that client based on its component id or other similar identifier.
It is possible for manager to maintain separate comp_info for each client. But I think comp_info is heavy weight in the case, as only we care about memory resource. comp_info also includes information about cap tables.

src/components/lib/cos_kernel_api.c
ryuxin: `ci->pgtbl_cap`  is not `SELF_PT`
`meta->pgtbl_cap` is `SELF_PT`

src/components/lib/cos_kernel_api.c
ryuxin: After RETYPE, we still can not access that page.
Only after we map the page into client's page table, client can access it. See CAPTBL_OP_MEMACTIVATE operation.

src/components/lib/cos_kernel_api.c
ryuxin: I think the current semantics of OP_CONS is fine.
It connects a subtree in **our own** cap table to a node within anther cap table which needs to be explicated provided.
**our own** cap table is always implicitly provided, unless we want to modify it. 
Whenever we want to modify a cap table including our own cap table, we have to provide a cap to it.
But we read some caps in **our own** cap table, we do not need to provide a captbl_cap to **our own** cap table.

src/components/lib/cos_kernel_api.c
ryuxin: See above comment. In this case, pte_cap has to be a cap in **our own** cap table.
We provide `meta->captbl_cap` because we modify **our own** cap table.
The interesting one is `ptemem_cap`. It is a cap in page table instead of cap table.
So we have to provide a cap to that page table, which is `meta->local_mi.pgtbl_cap`
Of course we need to be careful if you provide the correct page table

src/components/lib/cos_kernel_api.c
phanikishoreg: @ryuxin you're right but if we name it `UNTYPED_PT`, I was thinking, once it's retyped, if it would still be in that PT or would be mapped into SELF_PT. I understand that it is mainly physical memory mappings that will be part of UNTYPED_PT, virtual mappings will be in SELF_PT.

src/components/lib/cos_kernel_api.c
phanikishoreg: `captblcap` here is the new capability id that was created at the `frontier` not referring to our own captbl. That which was just activated at this step, will need to be `OP_CONS`ed into the `ci->captbl_cap` where we are trying to extend at `captblid_add`. I can see that there is no consistency check for this address in the kernel on `OP_CONS`, I wonder if that is required.

src/components/lib/cos_kernel_api.c
phanikishoreg: `SELF_PT` is the offset in any captbl in my understanding. It wasn't literal meaning of "SELF" that I was referring to. So, `ci->pgtbl_cap` is `SELF_PT` of `ci`, same way for `meta`. 

But I understand now what you mean by `SELF_PT`. And I also understand that we are going to have physical memory mappings in `UNTYPED_PT` and virtual memory mappings in `SELF_PT`, which is what we mostly use `SELF_PT` for anyway. Physical memory pages are only alloced and mapped in `memsrc`, so it should not impact the API much, I believe.

src/kernel/include/shared/cos_types.h
phanikishoreg: Well, I'm still trying to comprehend this whole idea of `UNTYPED_PT` distinction.
But I think the idea here is that, the `UNTYPED_PT` actually has mappings to the physical memory and `SELF_PT` has for virtual memory, which may as well be the memory that is used by memmgr or any component. Isn't it?

gparmer:
Not sure if closing this was the right thing to do.  Now we can't have
record of the conversation, and I can't reply to Phani.  Oh well.
g

src/kernel/include/shared/cos_types.h
ryuxin: Yes, I think so. SELF_PT tracks virtual memory of the manager

src/components/lib/cos_kernel_api.c
ryuxin: More concrete, `SELF_PT` is a fixed value while `ci->pgtbl_cap` is a variable.
But you are right. both `ci->pgtbl_cap` in manager's cap table and `SELF_PT` in ci's cap table refers to the same page table

PR #190, opened 2016-08-08
use separate page table to manage untyped memory

ryuxin:
### Summary of this PR

put all the untyped memory into a separate page table instead of in the booter's PT.
technique details:
kernel:
modify boot_comp.c 
create a new page table at the offset BOOT_CAPTBL_SELF_UNTYPED_PT.
map all untyped memory into that page table starting from the first page. Starting address 0 cause allocation to return NULL, which mistakes the allocation is fail.
modify boot_pgtbl_mappings_add function to parameterize which pgtbl to use, changing its argument from page table physical address to page table capability.
Mircor_booter:
 expand cos_meminfo struct to include a cap to page table which provides memory.
 fix some mistyping.

some explanation to the page table capability.
meta (struct cos_compinfo *memsrc) maintains the manager component's information, the pgtbl_cap in meta is the capability to the manager's own page table.
The pgtbl_cap in local_mi is the capability to a page table which tracks memory resource, currently it tracks all local memory.
ci maintains the information of a random client component. The pgtbl_cap in ci a capability to client's page table, and that capability is within manager's (meta) capability table.
The same logic applies to the captbl_cap
### Code Quality

Style:

Code Craftsmanship:

### DIFF
--- a/src/components/implementation/tests/micro_booter/micro_booter.c
+++ b/src/components/implementation/tests/micro_booter/micro_booter.c
@@ -568,7 +568,7 @@ term_fn(void *d)
 void
 cos_init(void)
 {
-	cos_meminfo_init(&booter_info.mi, BOOT_MEM_KM_BASE, COS_MEM_KERN_PA_SZ);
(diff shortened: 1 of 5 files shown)


gparmer:
This looks great.  I think it is ready to merge.

The pending question is how we are going to need to change/update the interface to allow the client of the library to choose which `mem_info` to use for a given kernel resource allocation.

PR #191, opened 2016-08-11
Adds vkernel feature, fix for copying Kernel PTEs to booter's pgtbl, other minor cleanups

phanikishoreg:
### Summary of this PR

Major feature added is vkernel, which boots up multiple VMs (components) and runs micro_booter (unit & ubench tests) in each of the components. vkernel component schedules VMs in a simple round-robin way. Added API to allocate untyped memory to each VM starting from the end of the untyped_frontier for isolated chunck of memory for each VM. 
Fixed a bug that doesn't copy kernel PTEs into the booter component's pgtbl. 

Todo in vkernel: (Which I'll add iteratively, along with if any review comments in this pull-request)
1. Shared-memory feature
2. cross-vm communication channels
3. asnd/rcv capabilities between vkernel<->vms for tcaps delegation.

### Code Quality

Style:

Code Craftsmanship:

- [x] I agree that there is no "throw-away" code, and that code in this PR is of high quality: *any such hackish code is well commented with `TODO`/`FIXME`s*

### DIFF
--- a/src/components/implementation/no_interface/vkernel/micro_booter.c
+++ b/src/components/implementation/no_interface/vkernel/micro_booter.c
@@ -1,0 +1,579 @@
+#include <stdio.h>
+#include <string.h>
+
+#undef assert
+#ifndef assert
+/* On assert, immediately switch to the "exit" thread */
+#define assert(node) do { if (unlikely(!(node))) { debug_print("assert error in @ "); cos_thd_switch(VM_CAPTBL_SELF_EXITTHD_BASE);} } while(0)
+#endif
+
+#define PRINT_FN prints
+#define debug_print(str) (PRINT_FN(str __FILE__ ":" STR(__LINE__) ".\n"))
+#define BUG() do { debug_print("BUG @ "); *((int *)0) = 0; } while (0);
+#define BUG_DIVZERO() do { debug_print("Testing divide by zero fault @ "); int i = num / den; } while (0);
+#define EXIT() do { while (1) cos_thd_switch(VM_CAPTBL_SELF_EXITTHD_BASE); } while (0);
+
+#include <cos_component.h>
+#include <cobj_format.h>
+#include <cos_kernel_api.h>
+
+static void
+cos_llprint(char *s, int len)
+{ call_cap(PRINT_CAP_TEMP, (int)s, len, 0, 0); }
+
+int
+prints(char *s)
+{
+	int len = strlen(s);
+
+	cos_llprint(s, len);
+
+	return len;
+}
+
+int __attribute__((format(printf,1,2)))
+printc(char *fmt, ...)
+{
+	  char s[128];
+	  va_list arg_ptr;
+	  int ret, len = 128;
+
+	  va_start(arg_ptr, fmt);
+	  ret = vsnprintf(s, len, fmt, arg_ptr);
+	  va_end(arg_ptr);
+	  cos_llprint(s, ret);
+
+	  return ret;
+}
+
+int vmid = -1;
+#define PRINTVM(fmt, args...) printc("%d: " fmt, vmid , ##args)
+
+struct cos_compinfo booter_info;
+/* For Div-by-zero test */
+int num = 1, den = 0;
+
+#define ITER 10000
+#define TEST_NTHDS 5
+unsigned long tls_test[TEST_NTHDS];
+
+static unsigned long
+tls_get(size_t off)
+{
+	unsigned long val;
+
+	__asm__ __volatile__("movl %%gs:(%1), %0" : "=r" (val) : "r" (off) : );
+
+	return val;
+}
+
+static void
+tls_set(size_t off, unsigned long val)
+{ __asm__ __volatile__("movl %0, %%gs:(%1)" : : "r" (val), "r" (off) : "memory"); }
+
+static void
+thd_fn_perf(void *d)
+{
+	cos_thd_switch(BOOT_CAPTBL_SELF_INITTHD_BASE);
+
+	while(1) {
+		cos_thd_switch(BOOT_CAPTBL_SELF_INITTHD_BASE);
+	}
+	PRINTVM("Error, shouldn't get here!\n");
+}
+
+static void
+test_thds_perf(void)
+{
+	thdcap_t ts;
+	long long total_swt_cycles = 0;
+	long long start_swt_cycles = 0, end_swt_cycles = 0;
+	int i;
+
+	ts = cos_thd_alloc(&booter_info, booter_info.comp_cap, thd_fn_perf, NULL);
+	assert(ts);
+	cos_thd_switch(ts);
+
+	rdtscll(start_swt_cycles);
+	for (i = 0 ; i < ITER ; i++) {
+		cos_thd_switch(ts);
+	}
+	rdtscll(end_swt_cycles);
+	total_swt_cycles = (end_swt_cycles - start_swt_cycles) / 2LL;
+
+	PRINTVM("Average THD SWTCH (Total: %lld / Iterations: %lld ): %lld\n",
+		total_swt_cycles, (long long) ITER, (total_swt_cycles / (long long)ITER));
+}
+
+static void
+thd_fn(void *d)
+{
+	PRINTVM("\tNew thread %d with argument %d, capid %ld\n", cos_thdid(), (int)d, tls_test[(int)d]);
+	/* Test the TLS support! */
+	assert(tls_get(0) == tls_test[(int)d]);
+	while (1) cos_thd_switch(BOOT_CAPTBL_SELF_INITTHD_BASE);
+	PRINTVM("Error, shouldn't get here!\n");
+}
+
+static void
+test_thds(void)
+{
+	thdcap_t ts[TEST_NTHDS];
+	int i;
+	int ret;
+
+	for (i = 0 ; i < TEST_NTHDS ; i++) {
+		ts[i] = cos_thd_alloc(&booter_info, booter_info.comp_cap, thd_fn, (void *)i);
+		assert(ts[i]);
+		tls_test[i] = i;
+		cos_thd_mod(&booter_info, ts[i], &tls_test[i]);
+		PRINTVM("switchto %d @ %x\n", (int)ts[i], cos_introspect(&booter_info, ts[i], 0));
+		cos_thd_switch(ts[i]);
+	}
+
+        PRINTVM("test done\n");
+}
+
+#define TEST_NPAGES (1024*2) 	/* Testing with 8MB for now */
+
+static void
+test_mem(void)
+{
+	char *p, *s, *t, *prev;
+	int i;
+	const char *chk = "SUCCESS";
+
+	p = cos_page_bump_alloc(&booter_info);
+	assert(p);
+	strcpy(p, chk);
+
+	assert(0 == strcmp(chk, p));
+	PRINTVM("%s: Page allocation\n", p);
+
+	s = cos_page_bump_alloc(&booter_info);
+	assert(s);
+	prev = s;
+	for (i = 0 ; i < TEST_NPAGES ; i++) {
+		t = cos_page_bump_alloc(&booter_info);
+		assert(t && t == prev + 4096);
+		prev = t;
+	}
+	memset(s, 0, TEST_NPAGES * 4096);
+	PRINTVM("SUCCESS: Allocated and zeroed %d pages.\n", TEST_NPAGES);
+}
+
+volatile arcvcap_t rcc_global, rcp_global;
+volatile asndcap_t scp_global;
+int async_test_flag = 0;
+
+static void
+async_thd_fn_perf(void *thdcap)
+{
+	thdcap_t tc = (thdcap_t)thdcap;
+	arcvcap_t rc = rcc_global;
+	int i;
+
+	cos_rcv(rc);
+
+	for (i = 0 ; i < ITER + 1 ; i++) {
+		cos_rcv(rc);
+	}
+
+	cos_thd_switch(tc);
+}
+
+static void
+async_thd_parent_perf(void *thdcap)
+{
+	thdcap_t tc = (thdcap_t)thdcap;
+	arcvcap_t rc = rcp_global;
+	asndcap_t sc = scp_global;
+	long long total_asnd_cycles = 0;
+	long long start_asnd_cycles = 0, end_arcv_cycles = 0;
+	int i;
+
+	cos_asnd(sc);
+
+	rdtscll(start_asnd_cycles);
+	for (i = 0 ; i < ITER ; i++) {
+		cos_asnd(sc);
+	}
+	rdtscll(end_arcv_cycles);
+	total_asnd_cycles = (end_arcv_cycles - start_asnd_cycles) / 2;
+
+	PRINTVM("Average ASND/ARCV (Total: %lld / Iterations: %lld ): %lld\n",
+		total_asnd_cycles, (long long) (ITER), (total_asnd_cycles / (long long)(ITER)));
+
+	async_test_flag = 0;
+	cos_thd_switch(tc);
+}
+
+static void
+async_thd_fn(void *thdcap)
+{
+	thdcap_t tc = (thdcap_t)thdcap;
+	arcvcap_t rc = rcc_global;
+	thdid_t tid;
+	int rcving;
+	cycles_t cycles;
+	int pending;
+
+	PRINTVM("Asynchronous event thread handler.\t<-- rcving...\n");
+	pending = cos_rcv(rc);
+	PRINTVM("<-- pending %d\t<-- rcving...\n", pending);
+	pending = cos_rcv(rc);
+	PRINTVM("<-- pending %d\t<-- rcving...\n", pending);
+	pending = cos_rcv(rc);
+	PRINTVM("<-- Error: manually returning to snding thread.\n");
+	cos_thd_switch(tc);
+	PRINTVM("ERROR: in async thd *after* switching back to the snder.\n");
+	while (1) ;
+}
+
+static void
+async_thd_parent(void *thdcap)
+{
+	thdcap_t  tc = (thdcap_t)thdcap;
+	arcvcap_t rc = rcp_global;
+	asndcap_t sc = scp_global;
+	int ret, pending;
+	thdid_t tid;
+	int rcving;
+	cycles_t cycles;
+
+	PRINTVM("--> sending\n");
+	ret = cos_asnd(sc);
+	if (ret) PRINTVM("asnd returned %d.\n", ret);
+	PRINTVM("--> Back in the asnder.\t--> sending\n");
+	ret = cos_asnd(sc);
+	if (ret) PRINTVM("--> asnd returned %d.\n", ret);
+	PRINTVM("--> Back in the asnder.\t--> receiving to get notifications\n");
+	pending = cos_sched_rcv(rc, &tid, &rcving, &cycles);
+	PRINTVM("--> pending %d, thdid %d, rcving %d, cycles %lld\n", pending, tid, rcving, cycles);
+
+	async_test_flag = 0;
+	cos_thd_switch(tc);
+}
+
+static void
+test_async_endpoints(void)
+{
+	thdcap_t  tcp,  tcc;
+	tcap_t    tccp, tccc;
+	arcvcap_t rcp,  rcc;
+	int ret;
+
+	PRINTVM("Creating threads, and async end-points.\n");
+	/* parent rcv capabilities */
+	tcp = cos_thd_alloc(&booter_info, booter_info.comp_cap, async_thd_parent, (void*)BOOT_CAPTBL_SELF_INITTHD_BASE);
+	assert(tcp);
+	tccp = cos_tcap_alloc(&booter_info, TCAP_PRIO_MAX + 2);
+	assert(tccp);
+	rcp = cos_arcv_alloc(&booter_info, tcp, tccp, booter_info.comp_cap, BOOT_CAPTBL_SELF_INITRCV_BASE);
+	assert(rcp);
+	if ((ret = cos_tcap_transfer(rcp, BOOT_CAPTBL_SELF_INITTCAP_BASE, TCAP_RES_INF, TCAP_PRIO_MAX + 1))) {
+		PRINTVM("transfer failed: %d\n", ret);
+		assert(0);
+	}
+
+	/* child rcv capabilities */
+	tcc = cos_thd_alloc(&booter_info, booter_info.comp_cap, async_thd_fn, (void*)tcp);
+	assert(tcc);
+	tccc = cos_tcap_alloc(&booter_info, TCAP_PRIO_MAX + 1);
+	assert(tccc);
+	rcc = cos_arcv_alloc(&booter_info, tcc, tccc, booter_info.comp_cap, rcp);
+	assert(rcc);
+	if (cos_tcap_transfer(rcc, BOOT_CAPTBL_SELF_INITTCAP_BASE, TCAP_RES_INF, TCAP_PRIO_MAX)) assert(0);
+
+	/* make the snd channel to the child */
+	scp_global = cos_asnd_alloc(&booter_info, rcc, booter_info.captbl_cap);
+	assert(scp_global);
+
+	rcc_global = rcc;
+	rcp_global = rcp;
+
+	async_test_flag = 1;
+	while (async_test_flag) cos_thd_switch(tcp);
+
+	PRINTVM("Async end-point test successful.\tTest done.\n");
+}
+
+static void
+test_async_endpoints_perf(void)
+{
+	thdcap_t tcp, tcc;
+	tcap_t tccp, tccc;
+	arcvcap_t rcp, rcc;
+
+	/* parent rcv capabilities */
+	tcp = cos_thd_alloc(&booter_info, booter_info.comp_cap, async_thd_parent_perf, (void*)BOOT_CAPTBL_SELF_INITTHD_BASE);
+	assert(tcp);
+	tccp = cos_tcap_alloc(&booter_info, TCAP_PRIO_MAX + 2);
+	assert(tccp);
+	rcp = cos_arcv_alloc(&booter_info, tcp, tccp, booter_info.comp_cap, BOOT_CAPTBL_SELF_INITRCV_BASE);
+	assert(rcp);
+	if (cos_tcap_transfer(rcp, BOOT_CAPTBL_SELF_INITTCAP_BASE, TCAP_RES_INF, TCAP_PRIO_MAX + 1)) assert(0);
+
+	/* child rcv capabilities */
+	tcc = cos_thd_alloc(&booter_info, booter_info.comp_cap, async_thd_fn_perf, (void*)tcp);
+	assert(tcc);
+	tccc = cos_tcap_alloc(&booter_info, TCAP_PRIO_MAX + 1);
+	assert(tccc);
+	rcc = cos_arcv_alloc(&booter_info, tcc, tccc, booter_info.comp_cap, rcp);
+	assert(rcc);
+	if (cos_tcap_transfer(rcc, BOOT_CAPTBL_SELF_INITTCAP_BASE, TCAP_RES_INF, TCAP_PRIO_MAX)) assert(0);
+
+	/* make the snd channel to the child */
+	scp_global = cos_asnd_alloc(&booter_info, rcc, booter_info.captbl_cap);
+	assert(scp_global);
+
+	rcc_global = rcc;
+	rcp_global = rcp;
+
+	async_test_flag = 1;
+	while (async_test_flag) cos_thd_switch(tcp);
+}
+
+#define TCAP_NLAYERS 3
+static volatile int child_activated[TCAP_NLAYERS][2];
+/* tcap child/parent receive capabilities, and the send capability */
+static volatile arcvcap_t tc_crc[TCAP_NLAYERS][2], tc_prc[TCAP_NLAYERS][2];
+static volatile asndcap_t tc_sc[TCAP_NLAYERS][3];
+
+static void
+tcap_child(void *d)
+{
+	arcvcap_t __tc_crc = (arcvcap_t)d;
+
+	while (1) {
+		int pending;
+
+		pending = cos_rcv(__tc_crc);
+		PRINTVM("tcap_test:rcv: pending %d\n", pending);
+	}
+}
+
+static void
+tcap_parent(void *d)
+{
+	int i;
+	asndcap_t __tc_sc = (asndcap_t)d;
+
+	for (i = 0 ; i < ITER ; i++) {
+		cos_asnd(__tc_sc);
+	}
+}
+
+/* static void */
+/* test_tcaps(void) */
+/* { */
+/* 	thdcap_t tcp, tcc; */
+/* 	tcap_t tccp, tccc; */
+/* 	arcvcap_t rcp, rcc; */
+
+/* 	/\* parent rcv capabilities *\/ */
+/* 	tcp = cos_thd_alloc(&booter_info, booter_info.comp_cap, tcap_parent, (void*)BOOT_CAPTBL_SELF_INITTHD_BASE); */
+/* 	assert(tcp); */
+/* 	tccp = cos_tcap_split(&booter_info, BOOT_CAPTBL_SELF_INITTCAP_BASE, 0, 0); */
+/* 	assert(tccp); */
+/* 	rcp = cos_arcv_alloc(&booter_info, tcp, tccp, booter_info.comp_cap, BOOT_CAPTBL_SELF_INITRCV_BASE); */
+/* 	assert(rcp); */
+
+/* 	/\* child rcv capabilities *\/ */
+/* 	tcc = cos_thd_alloc(&booter_info, booter_info.comp_cap, tcap_child, (void*)tcp); */
+/* 	assert(tcc); */
+/* 	tccc = cos_tcap_split(&booter_info, BOOT_CAPTBL_SELF_INITTCAP_BASE, 0, 0); */
+/* 	assert(tccc); */
+/* 	rcc = cos_arcv_alloc(&booter_info, tcc, tccc, booter_info.comp_cap, rcp); */
+/* 	assert(rcc); */
+
+/* 	/\* make the snd channel to the child *\/ */
+/* 	scp_global = cos_asnd_alloc(&booter_info, rcc, booter_info.captbl_cap); */
+/* 	assert(scp_global); */
+
+/* 	rcc_global = rcc; */
+/* 	rcp_global = rcp; */
+
+/* 	async_test_flag = 1; */
+/* 	while (async_test_flag) cos_thd_switch(tcp); */
+/* } */
+
+static void
+spinner(void *d)
+{ while (1) ; }
+
+static void
+test_timer(void)
+{
+	int i;
+	thdcap_t tc;
+	cycles_t c = 0, p = 0, t = 0;
+
+	PRINTVM("Starting timer test.\n");
+	tc = cos_thd_alloc(&booter_info, booter_info.comp_cap, spinner, NULL);
+
+	for (i = 0 ; i <= 16 ; i++) {
+		thdid_t tid;
+		int rcving;
+		cycles_t cycles;
+
+		cos_sched_rcv(BOOT_CAPTBL_SELF_INITRCV_BASE, &tid, &rcving, &cycles);
+		cos_thd_switch(tc);
+		p = c;
+		rdtscll(c);
+		if (i > 0) t += c-p;
+	}
+
+	PRINTVM("\tCycles per tick (10 microseconds) = %lld\n", t/16);
+
+	PRINTVM("Timer test completed.\tSuccess.\n");
+}
+
+long long midinv_cycles = 0LL;
+
+int
+test_serverfn(int a, int b, int c)
+{
+	rdtscll(midinv_cycles);
+	return 0xDEADBEEF;
+}
+
+extern void *__inv_test_serverfn(int a, int b, int c);
+
+static inline
+int call_cap_mb(u32_t cap_no, int arg1, int arg2, int arg3)
+{
+	int ret;
+
+	/*
+	 * Which stack should we use for this invocation?  Simple, use
+	 * this stack, at the current sp.  This is essentially a
+	 * function call into another component, with odd calling
+	 * conventions.
+	 */
+	cap_no = (cap_no + 1) << COS_CAPABILITY_OFFSET;
+
+	__asm__ __volatile__( \
+		"pushl %%ebp\n\t" \
+		"movl %%esp, %%ebp\n\t" \
+		"movl %%esp, %%edx\n\t" \
+		"movl $1f, %%ecx\n\t" \
+		"sysenter\n\t" \
+		"1:\n\t" \
+		"popl %%ebp" \
+		: "=a" (ret)
+		: "a" (cap_no), "b" (arg1), "S" (arg2), "D" (arg3) \
+		: "memory", "cc", "ecx", "edx");
+
+	return ret;
+}
+
+static void
+test_inv(void)
+{
+	compcap_t cc;
+	sinvcap_t ic;
+	unsigned int r;
+
+	cc = cos_comp_alloc(&booter_info, booter_info.captbl_cap, booter_info.pgtbl_cap, (vaddr_t)NULL);
+	assert(cc > 0);
+	ic = cos_sinv_alloc(&booter_info, cc, (vaddr_t)__inv_test_serverfn);
+	assert(ic > 0);
+
+	r = call_cap_mb(ic, 1, 2, 3);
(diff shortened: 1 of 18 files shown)


src/platform/i386/boot_comp.c
```diff
@@ -159,12 +174,14 @@ kern_boot_comp(void)
 	 * Need to account for the pages that will be allocated as
 	 * PTEs
 	 */
-	untyped_pgd = (pgtbl_t)chal_va2pa(mem_boot_alloc(1));
-	if (pgtbl_activate(ct, BOOT_CAPTBL_SELF_CT, BOOT_CAPTBL_SELF_UNTYPED_PT, untyped_pgd, 0)) assert(0);
+	if (pgtbl_activate(ct, BOOT_CAPTBL_SELF_CT, BOOT_CAPTBL_SELF_UNTYPED_PT, pgtbl, 0)) assert(0);
 	nkmemptes = boot_nptes(mem_utmem_end() - mem_boot_end());
 	ret = boot_pgtbl_mappings_add(ct, BOOT_CAPTBL_SELF_UNTYPED_PT, BOOT_CAPTBL_KM_PTE, "untyped memory", mem_boot_nalloc_end(nkmemptes),
```
ryuxin: As each page table reserves a range for the kernel, we add untyped memory starting from the address `BOOT_MEM_KM_BASE` (which is the first page currently). Is it possible to overwrite the kernel range in the page table?

phanikishoreg: We shouldn't. Ideally in High-kernels that most are, kernel is mapped in higher addresses of all address spaces, 3g-4g. And the rest of the virtual memory 0-3g are used for user-virtual memory. In our case, we do the same. so we will not be overwriting the PTEs of kernel mappings.

phanikishoreg: oh, they are called "Higher-half kernels". I'm sorry.

src/components/include/cos_kernel_api.h
```diff
@@ -83,8 +87,10 @@ int cos_sched_rcv(arcvcap_t rcv, thdid_t *thdid, int *rcving, cycles_t *cycles);
 
 int cos_introspect(struct cos_compinfo *ci, capid_t cap, unsigned long op);
 
-int cos_mem_alias(pgtblcap_t ptdst, vaddr_t dst, pgtblcap_t ptsrc, vaddr_t src);
-int cos_mem_move(pgtblcap_t ptdst, vaddr_t dst, pgtblcap_t ptsrc, vaddr_t src);
+vaddr_t cos_mem_alias(struct cos_compinfo *dstci, struct cos_compinfo *srcci, vaddr_t src);
+int cos_mem_alias_at(struct cos_compinfo *dstci, vaddr_t dst, struct cos_compinfo *srcci, vaddr_t src);
+vaddr_t cos_mem_move(struct cos_compinfo *dstci, struct cos_compinfo *srcci, vaddr_t src);
+int cos_mem_move_at(struct cos_compinfo *dstci, vaddr_t dst, struct cos_compinfo *srcci, vaddr_t src);
 int cos_mem_remove(pgtblcap_t pt, vaddr_t addr);
```
phanikishoreg: I think we should keep the API as before, passing in the `pgtblcap_t` instead of compinfo because we now have 2 pgtbl caps with each Comp and they can be copying from either. Right?

src/platform/i386/boot_comp.c
ryuxin: This makes sense to me. Thanks

src/components/implementation/no_interface/vkernel/micro_booter.c
```diff
@@ -0,0 +1,579 @@
+#include <stdio.h>
```
gparmer: A comment that you already know:  we don't want to replicate all of this code.

We want to break the `micro_booter` into the parts that are shared, and those that are different.  Shared should be soft-linked, and what's different should be in a separate `.c` file.

src/components/implementation/no_interface/vkernel/testinv.S
```diff
@@ -0,0 +1,18 @@
+/* Use the passed in stack in arg4 */
```
gparmer: Since you didn't change this at all, this should certainly be a soft link.

src/components/include/cos_kernel_api.h
gparmer: Isn't it still the right thing to do to use the `compinfo` level of abstraction where we can?

I think we need to figure out how we want to "address" the different page-tables.  We'd rather do it in a way that doesn't break the `compinfo` abstraction rather than directly addressing the page-tables.

src/components/lib/cos_kernel_api.c
```diff
@@ -276,18 +283,14 @@ __capid_bump_alloc(struct cos_compinfo *ci, cap_t cap)
...
 	struct cos_compinfo *meta = __compinfo_metacap(ci);
 
-	printd("__page_bump_alloc\n");
-
 	assert(meta == __compinfo_metacap(meta)); /* prevent unbounded structures */
-	heap_vaddr = ci->vas_frontier;
 
-	/* Do we need to allocate a PTE? */
-	if (heap_vaddr == ci->vasrange_frontier) {
+	for (addr = mem_ptr; addr < round_up_to_pgd_page(mem_ptr + mem_sz); addr += PGD_RANGE) {
```
gparmer: `for (addr = mem_ptr ; addr < round_up_to_pgd_page(mem_ptr + mem_sz) ; addr += PGD_RANGE) {`

This needs the emphasis around the 3 parts.  Style guide.

src/components/lib/cos_kernel_api.c
```diff
@@ -300,26 +303,78 @@ __page_bump_valloc(struct cos_compinfo *ci)
 		if (call_cap_op(meta->captbl_cap, CAPTBL_OP_PGTBLACTIVATE,
 				pte_cap, meta->mi.pgtbl_cap, ptemem_cap, 1)) {
 			assert(0); /* race? */
```
gparmer: Why remove the `return`s here?

src/components/lib/cos_kernel_api.c
```diff
@@ -300,26 +303,78 @@ __page_bump_valloc(struct cos_compinfo *ci)
...
+}
+
+static void
+__cos_meminfo_alloc(struct cos_compinfo *ci, vaddr_t untyped_ptr, unsigned long untyped_sz)
+{
+	vaddr_t addr;
+	struct cos_compinfo *meta = __compinfo_metacap(ci);
+
+	/*
+	 * FIXME: untyped_ptr and untyped_sz better be PGD_RANGE aligned.
```
gparmer: Add assertions for this?

src/components/lib/cos_kernel_api.c
```diff
@@ -300,26 +303,78 @@ __page_bump_valloc(struct cos_compinfo *ci)
...
+	 */
+	assert(__bump_mem_expand_range(ci, ci->mi.pgtbl_cap, untyped_ptr, untyped_sz) == untyped_ptr);
+	
+	vaddr_t start_addr        = meta->mi.untyped_frontier - untyped_sz;
+	meta->mi.untyped_frontier = start_addr;
+
+	for (addr = untyped_ptr; addr < untyped_ptr + untyped_sz; addr += PAGE_SIZE, start_addr += PAGE_SIZE) {
```
gparmer: style guide

src/components/lib/cos_kernel_api.c
```diff
@@ -300,26 +303,78 @@ __page_bump_valloc(struct cos_compinfo *ci)
...
+static void
+__cos_meminfo_alloc(struct cos_compinfo *ci, vaddr_t untyped_ptr, unsigned long untyped_sz)
+{
+	vaddr_t addr;
+	struct cos_compinfo *meta = __compinfo_metacap(ci);
+
+	/*
+	 * FIXME: untyped_ptr and untyped_sz better be PGD_RANGE aligned.
+	 */
+	assert(__bump_mem_expand_range(ci, ci->mi.pgtbl_cap, untyped_ptr, untyped_sz) == untyped_ptr);
```
gparmer: You don't want to put code that *must execute* in an assertion.  If assertions are compiled out, then the code isn't executed.  The assertion should be on the return value from a previous invocation.

src/components/lib/cos_kernel_api.c
```diff
@@ -300,26 +303,78 @@ __page_bump_valloc(struct cos_compinfo *ci)
...
+	
+	vaddr_t start_addr        = meta->mi.untyped_frontier - untyped_sz;
```
gparmer: style.  Don't mix definitions and code.

src/components/lib/cos_kernel_api.c
```diff
@@ -300,26 +303,78 @@ __page_bump_valloc(struct cos_compinfo *ci)
...
 
-		ci->vasrange_frontier += PGD_RANGE;
-		assert(ci->vasrange_frontier == round_up_to_pgd_page(ci->vasrange_frontier));
+	/* assert(round_up_to_pgd_page(addr) == round_up_to_pgd_page(mem_ptr + mem_sz)); */
+	
+	return mem_ptr;
+}
+
+static void
+__cos_meminfo_alloc(struct cos_compinfo *ci, vaddr_t untyped_ptr, unsigned long untyped_sz)
```
gparmer: I'm missing why the name of this function matches what it does.  It certainly does not allocate a `meminfo` struct.

src/components/lib/cos_kernel_api.c
```diff
@@ -300,26 +303,78 @@ __page_bump_valloc(struct cos_compinfo *ci)
...
+	vaddr_t start_addr        = meta->mi.untyped_frontier - untyped_sz;
+	meta->mi.untyped_frontier = start_addr;
+
+	for (addr = untyped_ptr; addr < untyped_ptr + untyped_sz; addr += PAGE_SIZE, start_addr += PAGE_SIZE) {
+		if (call_cap_op(meta->mi.pgtbl_cap, CAPTBL_OP_MEMMOVE, start_addr, ci->mi.pgtbl_cap, addr, 0)) BUG();
+	}
+}
+
+void
+cos_meminfo_alloc(struct cos_compinfo *ci, vaddr_t untyped_ptr, unsigned long untyped_sz)
```
gparmer: It seems a little odd that this function doesn't take a meminfo as an argument.  This is maybe not bad, but I'm trying to get an intuition.

src/components/lib/cos_kernel_api.c
```diff
@@ -557,6 +612,33 @@ void *
...
+}
+
+int
+cos_cap_cpy_at(struct cos_compinfo *dstci, capid_t dstcap, struct cos_compinfo *srcci, capid_t srccap)
+{
+	assert(srcci && dstci);
+
+	if (!dstcap) return 0;
+
+	if (call_cap_op(srcci->captbl_cap, CAPTBL_OP_CPY, srccap, dstci->captbl_cap, dstcap, 0))  BUG();
```
gparmer: Does this require modifying any frontiers?

src/components/lib/cos_kernel_api.c
```diff
@@ -601,10 +683,28 @@ cos_rcv(arcvcap_t rcv)
...
+cos_mem_alias(struct cos_compinfo *dstci, struct cos_compinfo *srcci, vaddr_t src)
+{
+	vaddr_t dst;
+
+	assert(srcci && dstci);
+
+	dst = __page_bump_valloc(dstci);
+	if (unlikely(!dst)) return 0;
+
+	if (call_cap_op(srcci->pgtbl_cap, CAPTBL_OP_CPY, src, dstci->pgtbl_cap, dst, 0))  BUG();
```
gparmer: same

src/components/lib/cos_kernel_api.c
```diff
@@ -615,10 +715,29 @@ cos_mem_remove(pgtblcap_t pt, vaddr_t addr)
...
+	if (unlikely(!dst)) return 0;
+
+	if (call_cap_op(srcci->pgtbl_cap, CAPTBL_OP_MEMMOVE, src, dstci->pgtbl_cap, dst, 0))  BUG();
+
+	return dst;
+}
+
 int
-cos_mem_move(pgtblcap_t ptdst, vaddr_t dst, pgtblcap_t ptsrc, vaddr_t src)
+cos_mem_move_at(struct cos_compinfo *dstci, vaddr_t dst, struct cos_compinfo *srcci, vaddr_t src)
```
gparmer: should we have an amount of memory to move here?  I don't know if it would be useful.

src/components/lib/cos_kernel_api.c
```diff
@@ -689,11 +808,33 @@ cos_hw_cycles_per_usec(hwcap_t hwc)
...
+	fva = __page_bump_valloc(ci);
+	if (unlikely(!fva)) return NULL;
+
+	if(call_cap_op(hwc, CAPTBL_OP_HW_MAP, ci->pgtbl_cap, fva, pa, 0)) BUG();
 
-	if (unlikely(!va)) return NULL;
+	sz -= PAGE_SIZE;
+	pa += PAGE_SIZE;
+
+	while (sz) {
```
gparmer: `for`

src/components/lib/cos_kernel_api.c
```diff
@@ -689,11 +808,33 @@ cos_hw_cycles_per_usec(hwcap_t hwc)
...
+	vaddr_t fva, va;
+
+	assert(ci && hwc && pa && len);
+
+	sz   = round_up_to_page(len);
+
+	fva = __page_bump_valloc(ci);
+	if (unlikely(!fva)) return NULL;
+
+	if(call_cap_op(hwc, CAPTBL_OP_HW_MAP, ci->pgtbl_cap, fva, pa, 0)) BUG();
```
gparmer: what am I missing?  Could all of this be simplified with a `do { } while(...);`?  I don't like seeing the redundancy.

src/kernel/include/shared/cos_config.h
```diff
@@ -17,23 +17,19 @@
...
-#define COS_MEM_USER_PA_SZ    (1<<25) /* start with 32MB of memory */
 
 /*
  * 1 MB, note that this is not the PA of kernel-usable memory, instead
  * it is the PA of the kernel.  If you change this, update the kernel
  * linker script (.ld) as well.
  */
 #define COS_MEM_KERN_PA (0x00100000)
-#define COS_MEM_KERN_PA_ORDER (25)
+#define COS_MEM_KERN_PA_ORDER (29)
```
gparmer: I don't remember what this impacts.  Have you checked that there are no ill side-effects?

src/kernel/include/shared/cos_types.h
```diff
@@ -187,6 +187,9 @@ typedef enum {
 
 #define CAPTBL_EXPAND_SZ 128
 
+#define COS_VIRT_MACH_COUNT        2
```
gparmer: `VM` instead of `VIRT_MACH`?

This is very specific to your component.  I don't think it should be in `cos_types.h` which is for system-wide customization.

src/kernel/include/shared/cos_types.h
```diff
@@ -252,15 +255,20 @@ enum {
 	BOOT_CAPTBL_SELF_INITTCAP_BASE = BOOT_CAPTBL_SELF_INITTHD_BASE + NUM_CPU_COS*CAP16B_IDSZ,
 	BOOT_CAPTBL_SELF_INITRCV_BASE  = round_up_to_pow2(BOOT_CAPTBL_SELF_INITTCAP_BASE + NUM_CPU_COS*CAP16B_IDSZ, CAPMAX_ENTRY_SZ),
 	BOOT_CAPTBL_SELF_INITHW_BASE   = round_up_to_pow2(BOOT_CAPTBL_SELF_INITRCV_BASE + NUM_CPU_COS*CAP64B_IDSZ, CAPMAX_ENTRY_SZ),
-	BOOT_CAPTBL_LAST_CAP           = BOOT_CAPTBL_SELF_INITHW_BASE + CAP32B_IDSZ,
+	BOOT_CAPTBL_LAST_CAP           = round_up_to_pow2(BOOT_CAPTBL_SELF_INITHW_BASE + CAP32B_IDSZ, CAPMAX_ENTRY_SZ),
 	/* round up to next entry */
 	BOOT_CAPTBL_FREE               = round_up_to_pow2(BOOT_CAPTBL_LAST_CAP, CAPMAX_ENTRY_SZ)
 };
 
 enum {
```
gparmer: This enum should likely be named, so that we know what it is for.  Right now, without context, it isn't clear at all what it is used for (though I presume for the VM's captbls given it is in this pull req).

src/kernel/include/shared/cos_types.h
```diff
@@ -840,8 +848,8 @@ static inline void
 cos_mem_fence(void)
 { __asm__ __volatile__("mfence" ::: "memory"); }
 
-// ncpu * 16 (or max 256) entries. can be increased if necessary.
-#define COS_THD_INIT_REGION_SIZE (((NUM_CPU*16) > (1<<8)) ? (1<<8) : (NUM_CPU*16))
+/* 256 entries. can be increased if necessary */
+#define COS_THD_INIT_REGION_SIZE (1<<8)
```
gparmer: Did you mean to remove all of the `NUM_CPU` logic here?  Should we vary this based on the # of cores?

src/platform/i386/boot_comp.c
```diff
@@ -99,13 +101,22 @@ kern_boot_thd(struct captbl *ct, void *thd_mem, void *tcap_mem)
...
 	ret = arcv_activate(ct, BOOT_CAPTBL_SELF_CT, BOOT_CAPTBL_SELF_INITRCV_BASE,
 			    BOOT_CAPTBL_SELF_COMP, BOOT_CAPTBL_SELF_INITTHD_BASE,
 			    BOOT_CAPTBL_SELF_INITTCAP_BASE, 0, 1);
 	assert(!ret);
 
+	/*
+	 * boot component's mapped into SELF_PT,
+	 * switching to boot component's pgd
+	 */
+	cap_pt = (struct cap_pgtbl *)captbl_lkup(ct, BOOT_CAPTBL_SELF_PT);
```
gparmer: return value testing with `CAP_TYPECHK`

src/platform/i386/boot_comp.c
```diff
@@ -116,14 +127,14 @@ kern_boot_comp(void)
...
-	pgtbl_t pgtbl = (pgtbl_t)chal_va2pa(&boot_comp_pgd), untyped_pgd;
 	void *thd_mem, *tcap_mem;
+	pgtbl_t pgtbl   = (pgtbl_t)chal_va2pa(&boot_comp_pgd), boot_vm_pgd;
 	u32_t hw_bitmap = 0xFFFFFFFF;
 
 	printk("Setting up the booter component.\n");
 
 	boot_comp_captbl = mem_boot_alloc(BOOT_CAPTBL_NPAGES);
-        ct = captbl_create(boot_comp_captbl);
+        ct               = captbl_create(boot_comp_captbl);
```
gparmer: Not necessarily a note for you, but all of this code in this function seems to have strange tab vs. space inconsistencies.  We should look into the blame for that.

src/platform/i386/boot_comp.c
```diff
@@ -141,12 +152,16 @@ kern_boot_comp(void)
...
-        if (pgtbl_activate(ct, BOOT_CAPTBL_SELF_CT, BOOT_CAPTBL_SELF_PT, pgtbl, 0)) assert(0);
 
 	hw_asndcap_init();
 	if (hw_activate(ct, BOOT_CAPTBL_SELF_CT, BOOT_CAPTBL_SELF_INITHW_BASE, hw_bitmap)) assert(0);
 
-	printk("\tCapability table and page-table created.\n");
+	/*
+	 * separate pgd for boot component virtual memory
+	 */
+	boot_vm_pgd = (pgtbl_t)mem_boot_alloc(1);
```
gparmer: assert the return value != NULL.

src/platform/i386/boot_comp.c
```diff
@@ -141,12 +152,16 @@ kern_boot_comp(void)
...
+	memcpy((void *)boot_vm_pgd + KERNEL_PGD_REGION_OFFSET,  (void *)(&boot_comp_pgd) + KERNEL_PGD_REGION_OFFSET, KERNEL_PGD_REGION_SIZE); 
+	if (pgtbl_activate(ct, BOOT_CAPTBL_SELF_CT, BOOT_CAPTBL_SELF_PT, (pgtbl_t)chal_va2pa(boot_vm_pgd), 0)) assert(0);
```
gparmer: I wonder if we should have a function for doing this stuff.  Is this operation done multiple times?

src/platform/i386/boot_comp.c
gparmer: perhaps we should have an explicit check that we don't go above the kernel virtual memory start address.

src/platform/i386/boot_comp.c
ryuxin: Yes, I think so. Each time we create a new page table at the boot time, we need those operations

src/platform/i386/boot_comp.c
phanikishoreg: I guess we should. But in general, we have to make sure for every page-table that we create don't overwrite kernel memory then? that would be at every PGTBL_ACTIVATE?

src/components/include/cos_kernel_api.h
phanikishoreg: compinfo is the right thing to do because that is consistent across the cos_kernel_api. 

Yes, we need to figure out that. Perhaps by passing in an extra argument for each of the pgtbl operation api, that would change the API.

src/components/implementation/no_interface/vkernel/micro_booter.c
phanikishoreg: yes, i'll work on it.

src/components/implementation/no_interface/vkernel/testinv.S
phanikishoreg: yes, got it.

src/components/lib/cos_kernel_api.c
phanikishoreg: thank you, i'll fix that.

src/components/lib/cos_kernel_api.c
phanikishoreg: I was thinking that returns after assert(0) aren't useful but after I saw your other comment below, I realized my mistake. I'll put them back. Sorry.

src/components/lib/cos_kernel_api.c
phanikishoreg: I will in my next commit when I change the `KM_BASE` to be PGD_ALIGNED. At this point, `untyped_ptr` passed in is `PAGE_SIZE`, so I didn't and added FIXME.

src/components/lib/cos_kernel_api.c
phanikishoreg: got it. thank you.

src/components/lib/cos_kernel_api.c
phanikishoreg: got it.

src/components/lib/cos_kernel_api.c
phanikishoreg: I took care of every other declarations but I missed this. :( thank you.

src/components/lib/cos_kernel_api.c
phanikishoreg: This was a helper function to `cos_meminfo_alloc()` which actually allocates untyped region and updates meminfo of that component. I wasn't sure clear what to name it, __cos_meminfo_alloc() or __cos_untyped_alloc() or wasn't sure.

src/components/lib/cos_kernel_api.c
phanikishoreg: It mainly allocates untyped and updates the meminfo with the arguments passed, again wasn't sure what would be the right name, untyped_alloc or meminfo_alloc. Because functionally it does what meminfo_init does and other than allocating untyped memory, so I kept it this way.

src/components/lib/cos_kernel_api.c
phanikishoreg: I guess this API copies the srccap at any arbitrary location in dstci passed in as dstcap. So, if the caller passed in the arbitrary location, should have taken care of modifying cap_frontier. That was my intuition, do you think I should change it?

src/components/lib/cos_kernel_api.c
phanikishoreg: Guess we could. Every mem related API could but from what I've learned from our discussions, this is not going to be much of a use because we would essentially be doing what user would be doing otherwise with a for-loop.

src/components/lib/cos_kernel_api.c
phanikishoreg: I'll fix this.

src/kernel/include/shared/cos_config.h
phanikishoreg: There weren't any. I checked through the code for its usage and also executed micro_booter and vkernel. This has also been in effect in the rump2cos for a while now and haven't had any issues with this.

src/kernel/include/shared/cos_types.h
phanikishoreg: I wasn't sure, VM_COUNT sounded a bit ambiguous, Virtual Memory or Virtual Machine, so I was being explicit in this case. But when I put all these in vkernel, I'll name them VM. I guess, over there, I would have to be explicit about Memory not Machine. 

I'll put them in vkernel component, have a header within that and do it. Perhaps even the capabilities offsets.

src/kernel/include/shared/cos_types.h
phanikishoreg: okay, I'll name it. I wanted to have it consistent in cos_types, but I'll move this to vkernel and name it.

src/kernel/include/shared/cos_types.h
phanikishoreg: I guess I should put it back. I don't have a good/valid reason for removing it, except for changing the constant there from 16 to 32/64.

src/platform/i386/boot_comp.c
phanikishoreg: got it. thank you.

src/platform/i386/boot_comp.c
phanikishoreg: yes I noticed that too, I could have indented it by whole but wasn't sure if you'd approve it.

src/platform/i386/boot_comp.c
phanikishoreg: thank you.

src/platform/i386/boot_comp.c
phanikishoreg: Probably with yuxin's current work we might need to do this repeatedly. Should I make a function?

gparmer:
Another nitpick:  If you only have 4 commits, then there is virtually no way to do a good code-review.  Normally going commit by commit makes the most sense as it follows a progression of the code.  So I'm sure that I'm missing things in this review, which means you'll have to fix any of the fallout.

src/components/include/cos_kernel_api.h
gparmer: In the short term, I think that @ryuxin will hack it to make it work for his use case.  We might create a temporary API that takes page-table caps.  But I don't want to destroy the proper API for an edge case.

I think we're all on the same page.

src/components/lib/cos_kernel_api.c
gparmer: This isn't *just* allocating.  It is really initializing and moving memory into it, right?  The second part is not obvious from the name at all.

Maybe `__cos_meminfo_populate`?

src/components/lib/cos_kernel_api.c
gparmer: I'm not sure right now how to properly use it.  I'll look through the vkernel stuff and see an example.  this could be OK.

src/components/lib/cos_kernel_api.c
gparmer: It would save the user some effort, and would mimic the `memcpy` API that people are comfortable with.

The downside is what to do when we get an error from the kernel after moving half of the entries.  But we would likely rather handle that in the library, than rely on the user of the library to do it likely incorrectly.

src/components/lib/cos_kernel_api.c
gparmer: my comment about the for might be wrong if you could eliminate more redundancy using do while.

src/kernel/include/shared/cos_types.h
gparmer: OK, good point.  Maybe full VIRTUAL_MACHINE then.  MACH has historical definitions.

src/kernel/include/shared/cos_types.h
gparmer: You changed it to 256, not 32/64, right?

It is OK to make it higher.  But I wonder if we want it to increase with increasing number of cores.  Perhaps not, and we actually just have a backup linked list if the array is not large enough.

OK, you can just leave it.

src/components/implementation/no_interface/vkernel/vkernel.c
```diff
@@ -0,0 +1,195 @@
...
+struct vms_info {
+	struct cos_compinfo cinfo;
+	captblcap_t ct;
+	pgtblcap_t pt, utpt;
+	compcap_t cc;
+	thdcap_t initthd, exitthd;
+	thdid_t inittid;
+	tcap_t inittcap;
+	arcvcap_t initrcv;
+} vmx_info[COS_VIRT_MACH_COUNT];
```
gparmer: if this is the only place that this variable is used, then it should likely just be in defined this file.

src/components/implementation/no_interface/vkernel/vkernel.c
```diff
@@ -0,0 +1,195 @@
...
+	thdid_t tid;
+	int rcving;
+	cycles_t cycles;
+	int index;
+
+	while (ready_vms) {
+		cos_sched_rcv(BOOT_CAPTBL_SELF_INITRCV_BASE, &tid, &rcving, &cycles);
+		index = i++ % COS_VIRT_MACH_COUNT;
+		
+		if (vmx_info[index].initthd && vk_info.vminitasnd[index]) {
```
gparmer: If a VM is not allocated, the snd won't happen, and we'll rcv again.  If there are no pending events, we might block.  I don't think you want that here.

Shouldn't there just be a loop going through and finding the next active VM, and snd it?

This is a very simple policy.  I think that we should not switch to a VM after `cos_sched_rcv` reports that it is `rcving`, right?

src/components/implementation/no_interface/vkernel/vkernel.c
```diff
@@ -0,0 +1,195 @@
...
+	assert(COS_VIRT_MACH_COUNT >= 2);
+
+	cos_meminfo_init(&vk_cinfo->mi, BOOT_MEM_KM_BASE, COS_MEM_KERN_PA_SZ, BOOT_CAPTBL_SELF_UNTYPED_PT);
+	cos_compinfo_init(vk_cinfo, BOOT_CAPTBL_SELF_PT, BOOT_CAPTBL_SELF_CT, BOOT_CAPTBL_SELF_COMP,
+			(vaddr_t)cos_get_heap_ptr(), BOOT_CAPTBL_FREE, vk_cinfo);
+
+	vk_info.termthd = cos_thd_alloc(vk_cinfo, vk_cinfo->comp_cap, vk_terminate, NULL);
+	assert(vk_info.termthd);
+
+	for (id = 0; id < COS_VIRT_MACH_COUNT; id ++) {
```
gparmer: style

src/components/implementation/no_interface/vkernel/vkernel.c
```diff
@@ -0,0 +1,195 @@
...
+		vm_info->cc = cos_comp_alloc(vk_cinfo, vm_info->ct, vm_info->pt, (vaddr_t)&cos_upcall_entry);
+		assert(vm_info->cc);
+
+		cos_meminfo_init(&vm_cinfo->mi, BOOT_MEM_KM_BASE, COS_VIRT_MACH_UNTYPED_SIZE, vm_info->utpt);
+		cos_compinfo_init(vm_cinfo, vm_info->pt, vm_info->ct, vm_info->cc,
+				(vaddr_t)BOOT_MEM_VM_BASE, VM_CAPTBL_FREE, vk_cinfo);
+
+		vm_info->initthd = cos_thd_alloc(vk_cinfo, vm_cinfo->comp_cap, vm_init, (void *)id);
+		assert(vm_info->initthd);
+		vm_info->inittid = (thdid_t)cos_introspect(vk_cinfo, vm_info->initthd, 9);
```
gparmer: What's the 9?  Is there no DEFINED_VARIABLE for that?

src/components/implementation/no_interface/vkernel/vkernel.c
```diff
@@ -0,0 +1,195 @@
...
+#include <cos_kernel_api.h>
+
+#undef assert
+#define assert(node) do { if (unlikely(!(node))) { debug_print("assert error in @ "); *((int *)0) = 0; } } while (0)
+#define PRINT_FN prints
+#define debug_print(str) (PRINT_FN(str __FILE__ ":" STR(__LINE__) ".\n"))
+#define BUG() do { debug_print("BUG @ "); *((int *)0) = 0; } while (0);
+
+struct vms_info {
+	struct cos_compinfo cinfo;
```
gparmer: Why the redundancy with the `cinfo` that contains pt, ct, comp, ..., and then having the variables below to hold the same thing?

src/components/implementation/no_interface/vkernel/vkernel.c
```diff
@@ -0,0 +1,195 @@
...
+		/*
+		 * Create send end-point in VKernel to each VM's INITRCV end-point
+		 */
+		vk_info.vminitasnd[id] = cos_asnd_alloc(vk_cinfo, vm_info->initrcv, vk_cinfo->captbl_cap);
+		assert(vk_info.vminitasnd[id]);
+
+		/*
+		 * Create and copy booter comp virtual memory to each VM
+		 */
+		vm_range = (int)cos_get_heap_ptr() - BOOT_MEM_VM_BASE;
```
gparmer: `int` will break on 64 bit.

`unsigned long`?  `vaddr_t`?

src/components/implementation/no_interface/vkernel/vkernel.c
```diff
@@ -0,0 +1,195 @@
...
+		assert(vm_range > 0);
+		printc("\tMapping in Booter component's virtual memory (range:%u)\n", vm_range);
+		for (i = 0; i < vm_range; i += PAGE_SIZE) {
+			vaddr_t spg = (vaddr_t)cos_page_bump_alloc(vk_cinfo), dpg;
```
gparmer: I'm missing what `spg` stands for

src/components/implementation/no_interface/vkernel/vkernel.c
```diff
@@ -0,0 +1,195 @@
...
+	printc("------------------[ VKernel & VMs init complete ]------------------\n");
+
+	scheduler();
+
+	cos_hw_detach(BOOT_CAPTBL_SELF_INITHW_BASE, HW_PERIODIC);
+
+	printc("vkernel: END\n");
+	cos_thd_switch(vk_info.termthd);
+
+	printc("DEAD END\n");
```
gparmer: If this is an error print out an error report.

src/components/lib/cos_kernel_api.c
phanikishoreg: yeah, sounds good. thank you.

src/kernel/include/shared/cos_types.h
phanikishoreg: Oh ya, MACH! Sure, I'll make it more obvious.

src/kernel/include/shared/cos_types.h
phanikishoreg: Yes, I've changed it to 256. 
I meant I didn't have a valid argument to removing NUM_CPUs, I only could have changed the const from 16 -> 32/64. 

Thank you.

src/components/implementation/no_interface/vkernel/vkernel.c
phanikishoreg: If a Vm at `index` has exited, only then I set initthd = 0. 
And what you mentioned can happen, this would wait for next timer event and wouldn't schedule any other VM. But my concern was, if I have nested for/while here to check for next VM, I might be spending a lot of time here to find a vm to activate. And the periodic timer interval is 100us / 1ms in rump2cos, so it went for simplicity. 

Actually VM is woken up by vkernel scheduler, that means, when it reports that it is rcving, vkernel scheduler is the one responsible to be sending to it (virt-timer event). All I could do was, block for this immediate event and then maybe send to it the next time I index this VM. But that would also mean, I'd have to have another nested loop around cos_sched_rcv to receive all events to get all such blocked events and then mark them blocked for this real timer event and then unblock them at the next real timer event I'd get. All this increasing the scheduler's complexity, so I didn't do it for the first version.

src/components/implementation/no_interface/vkernel/vkernel.c
phanikishoreg: thank you.

src/components/implementation/no_interface/vkernel/vkernel.c
phanikishoreg: Yeah, there is no defined variable for any of the `thd_introspect()` commands. I'll add them, that should improve readability. Thank you.

src/components/implementation/no_interface/vkernel/vkernel.c
phanikishoreg: I did go through this thought process of it being redundant but I wanted to keep the every information I've about the VM and the cinfo was initialized after I've pt, utpt, ct, comp. It was also going to avoid additional indirections that would be a lot more confusing especially ones like `vms_info[x].cinfo.mi.pgtbl_cap`. 
And removing these 4 in vms_info struct would break a bit of consistency. 
So I included them as part of the book-keeping.

src/components/implementation/no_interface/vkernel/vkernel.c
phanikishoreg: yes, vaddr_t will be better. I'll change it, Thank you.

src/components/implementation/no_interface/vkernel/vkernel.c
phanikishoreg: spg = source page, dpg = destination page. ;)

src/components/implementation/no_interface/vkernel/vkernel.c
phanikishoreg: It is kind of an unreachable code and the error-report I had, yeah. There is probably no useful information I could print here as part of error report.

gparmer: I don't think you're right.  `"VKernel: back in initial thread after switching to terminal thread.  ERROR.\n"` sounds much more useful to me than having to grep for `DEAD END`.  One prints a self-describing error, the other requires further work.

`assert(0);` is perhaps the best answer.

src/components/implementation/no_interface/vkernel/vkernel.c
gparmer: since there are no data-structures that given any hint as to what this means, src_pg seems more descriptive.

src/components/implementation/no_interface/vkernel/vkernel.c
gparmer: redundant data is almost never a good idea.  if you're worried about verbosity, then create accessor functions.

src/components/implementation/no_interface/vkernel/vkernel.c
gparmer: I can't see a scenario in which your logic makes sense.  Spinning idly (possibly without delay if all vms fall into this behavior) is almost *never* the correct choice when you trivially have better options.  Your logic on asymptotic complexity (cost of finding a VM) seem trivially wrong to me, so I'm missing something, most likely.  We should talk.

you can pull with this now, but I don't think your decisions are correct.  This requires a talk, likely.

src/kernel/include/shared/cos_types.h
phanikishoreg: I'm moving this macro to vkernel, and will have it as `VM_COUNT` as you initially suggested and also for consistency sake. Else everywhere within vkernel, I'll have to expand VM/vm.

src/components/implementation/no_interface/vkernel/vkernel.c
phanikishoreg: I'm missing the feedback here, can you please elaborate? sorry about that

src/components/implementation/no_interface/vkernel/vkernel.c
phanikishoreg: okay, got it. I'll remove them from the vms_info struct.

src/components/implementation/no_interface/vkernel/vkernel.c
phanikishoreg: thank you, i'll change that.

src/components/implementation/no_interface/vkernel/vkernel.c
phanikishoreg: okay, I understand. thank you.

PR #196, opened 2016-09-19
Bug fixes related to tcap_delegate and a fix for cos_rcv/cos_sched_rcv

phanikishoreg:
### Summary of this PR

Fixed YIELD flag detection in cos_kernel_api and return from syscall handler after TCAP_DELEGATE. 
Also fixed cos_rcv/cos_sched_rcv return value (which was missed in previous PR). 

### Code Quality

Style:

Code Craftsmanship:

### DIFF
--- a/src/components/lib/cos_kernel_api.c
+++ b/src/components/lib/cos_kernel_api.c
@@ -787,7 +787,7 @@ cos_tcap_transfer(arcvcap_t dst, tcap_t src, tcap_res_t res, tcap_prio_t prio)
 int
 cos_tcap_delegate(asndcap_t dst, tcap_t src, tcap_res_t res, tcap_prio_t prio, tcap_deleg_flags_t flags)
 {
-	u32_t yield     = flags & TCAP_DELEG_YIELD;
+	u32_t yield     = (flags && TCAP_DELEG_YIELD);
 	/* top bit is if we are dispatching or not */
 	int prio_higher = (u32_t)(prio >> 32) | (yield << ((sizeof(yield)*8)-1));
 	int prio_lower  = (u32_t)((prio << 32) >> 32);
--- a/src/kernel/capinv.c
+++ b/src/kernel/capinv.c
@@ -483,7 +483,8 @@ cap_thd_switch(struct pt_regs *regs, struct thread *curr, struct thread  *next,
 		assert(!(next->state & THD_STATE_PREEMPTED));
 		next->state &= ~THD_STATE_RCVING;
 		thd_state_evt_deliver(next, &a, &b);
-		__userregs_setretvals(&next->regs, thd_rcvcap_pending_dec(next), a, b);
+		thd_rcvcap_pending_dec(next);
+		__userregs_setretvals(&next->regs, thd_rcvcap_pending(next), a, b);
 	}
 
 	copy_all_regs(&next->regs, regs);
@@ -877,6 +878,7 @@ done:
 	 * Thus the level of indirection here.
 	 */
 	if (!thd_switch) __userregs_set(regs, ret, __userregs_getsp(regs), __userregs_getip(regs));
+	else return ret; /* ret from cap_switch */
 
 	return 0;
 }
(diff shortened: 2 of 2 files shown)


src/components/lib/cos_kernel_api.c
```diff
@@ -787,7 +787,7 @@ cos_tcap_transfer(arcvcap_t dst, tcap_t src, tcap_res_t res, tcap_prio_t prio)
 int
 cos_tcap_delegate(asndcap_t dst, tcap_t src, tcap_res_t res, tcap_prio_t prio, tcap_deleg_flags_t flags)
 {
-	u32_t yield     = flags & TCAP_DELEG_YIELD;
+	u32_t yield     = (flags && TCAP_DELEG_YIELD);
```
gparmer: I'm not sure I can understand why this would be the desirable code.

It is functionally equivalent to `u32_t yield = (flags != 0);`.  If the intention is to see if the correct flag (i.e. bit) in the mask is set, then that is what `&` is made for.

If the previous code really is the problem, then I don't think that you understand the problem.  This feels like a fix for the symptoms, not a fix for the problem.

src/kernel/capinv.c
```diff
@@ -483,7 +483,8 @@ cap_thd_switch(struct pt_regs *regs, struct thread *curr, struct thread  *next,
 		assert(!(next->state & THD_STATE_PREEMPTED));
 		next->state &= ~THD_STATE_RCVING;
 		thd_state_evt_deliver(next, &a, &b);
-		__userregs_setretvals(&next->regs, thd_rcvcap_pending_dec(next), a, b);
+		thd_rcvcap_pending_dec(next);
+		__userregs_setretvals(&next->regs, thd_rcvcap_pending(next), a, b);
```
gparmer: This is fine for now.  I wonder if it is a problem with the API, or with the usage of the API.  Your fix is for the usage of the API.  Food for thought.

src/kernel/capinv.c
```diff
@@ -877,6 +878,7 @@ composite_syscall_handler(struct pt_regs *regs)
 	 * Thus the level of indirection here.
 	 */
 	if (!thd_switch) __userregs_set(regs, ret, __userregs_getsp(regs), __userregs_getip(regs));
+	else return ret; /* ret from cap_switch */
```
gparmer: line up the `return...` and the `__userregs_set...`

gparmer: Shouldn't this logic be:

```
if (thd_switch) return ret;
__userregs_set(regs, ret,  __userregs_getsp(regs), __userregs_getip(regs));
return 0;
```

???

The current conditional structure seems overly complicated and non-uniform (one branch returning, the other falling through, but with no textual indication that it is falling through.

src/kernel/capinv.c
```diff
@@ -1442,6 +1444,7 @@ composite_syscall_slowpath(struct pt_regs *regs, int *thd_switch)
 			n = asnd_process(rthd, thd, tcapdst, tcap_current(cos_info), &tcap_next, yield);
 			if (n != thd) {
 				ret = cap_switch(regs, thd, n, tcap_next, TCAP_TIME_NIL, ci, cos_info);
+				if (unlikely(ret < 0)) cos_throw(err, ret);
```
gparmer: I can't vouch for this right now.  If you're confident in it, then it is fine.

src/components/lib/cos_kernel_api.c
phanikishoreg: Damn! I'm wrong here, yes. It must be `((flags & TCAP_DELEG_YIELD) != 0)`.
I understood the problem but I coded it wrong. We only have 1 bit that designates to yield or not but my code was wrong, that it was not differentiating between different flags (though that worked). But with `&`, yield will be 2 (0010) which will be lost in the next step when we pack it.

I'll fix it to be `((flags & TCAP_DELEG_YIELD) != 0)`

src/kernel/capinv.c
phanikishoreg: For now, its the API return value that is the problem.
But I got your point, will need to think about this too after the deadline.

src/kernel/capinv.c
phanikishoreg: Yeah, your suggestion looks cleaner. I'll change that. Thank you!

src/kernel/capinv.c
phanikishoreg: `cap_switch` return value at this point is same as `cap_thd_switch` which only returns `0` or `1`. This is mainly for unexpected return value for documentation and readability. (even if we extend cap_switch/cap_thd_switch to return errors, this shouldn't need to be revisited). 

That's the main reason to add this.

PR #197, opened 2016-09-21
Bug fixes in cos_tcap_delegate/cos_rcv API, simple working vkernel scheduling with tcaps

phanikishoreg:
### Summary of this PR

1. Fixed a bug in cos_tcap_delegate for setting yield flag. 
2. Fixed return value of cos_rcv from kernel to include pending for sched queue.
3. Working simple scheduling in vkernel using cos_tcap_delegate with fixed budgets for all vms. 
4. Had to eliminate some of the tests that aren't feasible to run vkernel yet. 
5. Replaced the obvious BUG*() from Termination threads in vkernel and micro_booter with a SPIN() to avoid unnecessary dump that could overwrite useful logging information on baremetal.

### Code Quality

Style:

Code Craftsmanship:

- [x] I agree that there is no "throw-away" code, and that code in this PR is of high quality: **any such code that needs further work has FIXME/TODO comments around it**

### DIFF
--- a/src/components/implementation/no_interface/vkernel/vk_types.h
+++ b/src/components/implementation/no_interface/vkernel/vk_types.h
@@ -12,4 +12,7 @@ enum vm_captbl_layout {
 #define VM_COUNT        2	/* virtual machine count */
 #define VM_UNTYPED_SIZE (1<<26) /* untyped memory per vm = 64MB */
 
+#define VM_BUDGET_FIXED 400000
+#define VM_PRIO_FIXED   TCAP_PRIO_MAX
+
 #endif /* VK_TYPES_H */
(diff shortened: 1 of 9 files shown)


src/components/implementation/no_interface/vkernel/vk_types.h
```diff
@@ -12,4 +12,7 @@ enum vm_captbl_layout {
 #define VM_COUNT        2	/* virtual machine count */
 #define VM_UNTYPED_SIZE (1<<26) /* untyped memory per vm = 64MB */
 
+#define VM_BUDGET_FIXED 400000
```
gparmer: Is this every ~200 microseconds?  That is a very small quantum.

phanikishoreg: Yes, it is small. On the HW we're using here, its much lesser: ~117us. This is not going to remain for long, especially after I change the scheduling policy to something like credit-based, I'll change it accordingly.

PR #202, opened 2016-09-28
APIC Timer Support

phanikishoreg:
### Summary of this PR

Added support for APIC Timers, Both ONESHOT and TSC-DEADLINE. If TSC-DEADLINE is not supported by the CPU, it falls back to ONESHOT mode. (PERIODIC mode not supported.)
Qemu-i386 2.0.0 doesn't support TSC-DEADLINE, so it uses ONESHOT mode. 

### Code Quality

Style:

Code Craftsmanship:

### DIFF
--- a/src/platform/i386/lapic.c
+++ b/src/platform/i386/lapic.c
@@ -17,13 +17,10 @@
 #define LAPIC_PERIODIC_MODE    (0x01 << 17)
 #define LAPIC_ONESHOT_MODE     (0x00 << 17)
 #define LAPIC_TSCDEADLINE_MODE (0x02 << 17)
-
 #define LAPIC_MASK (1<<15)
 
-#define IA32_MSR_TSC		0x00000010
 #define IA32_MSR_TSC_DEADLINE	0x000006e0
 
-#define TEST_DEADLINE 3400000
 #define RETRY_ITERS   3
 
 extern int timer_process(struct pt_regs *regs);
@@ -77,22 +74,15 @@ lapic_find_localaddr(void *l)
 
 static void
 lapic_write_reg(u32_t off, u32_t val)
-{
-	*(u32_t *)(lapic + off) = val;
-}
+{ *(u32_t *)(lapic + off) = val; }
 
 static void
 lapic_ack(void)
-{
-	if (lapic) 
-		lapic_write_reg(LAPIC_EOI_REG, 0);
-}
+{ if (lapic) lapic_write_reg(LAPIC_EOI_REG, 0); }
 
 static u32_t
 lapic_read_reg(u32_t off)
-{
-	return *(u32_t *)(lapic + off);
-}
+{ return *(u32_t *)(lapic + off); }
 
 void
 lapic_set_timer(int timer_type, cycles_t deadline)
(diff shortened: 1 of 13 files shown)


src/platform/i386/lapic.c
```diff
@@ -86,26 +87,35 @@ lapic_read_reg(u32_t off)
...
-		printk("Mode not supported\n");
-		assert(0);
-	}
+	assert (timer_type != PERIODIC);
+
+	rdtscll(now);
+	if (deadline <= now) return;
+
+	if (timer_type == ONESHOT) {
+
```
gparmer: empty spaces?  What are these?

src/platform/i386/lapic.c
```diff
@@ -86,26 +87,35 @@ lapic_read_reg(u32_t off)
...
+		lapic_write_reg(LAPIC_INIT_COUNT_REG, (u32_t)(deadline - now));
+
+	} else {
```
gparmer: should be `else if (timer_type == DEADLINE)`

src/platform/i386/lapic.c
```diff
@@ -86,26 +87,35 @@ lapic_read_reg(u32_t off)
...
+		readmsr(IA32_MSR_TSC_DEADLINE, &low, &high);
+		if (!low && !high) {
+			retries --;
+			if (retries > 0) {
+				goto retry;
+			}
+			else {
+				printk("Something wrong.. Cannot program TSC-DEADLINE\n");
+				assert(0);
+			}
```
gparmer: Then after this conditional, 
```
else { 
        assert(0); 
}
```

src/platform/i386/lapic.c
```diff
@@ -86,26 +87,35 @@ lapic_read_reg(u32_t off)
...
-	u32_t high, low;
-	int retries = RETRY_ITERS;
+{
+	u64_t now;
 
-	if (timer_type != TSC_DEADLINE) {
-		printk("Mode not supported\n");
-		assert(0);
-	}
+	assert (timer_type != PERIODIC);
```
gparmer: The problem with putting the assertion here rather than structuring your conditionals like below is 

1. future changes that add more modes will not trigger an assertion, but won't hit one of the conditions.
2. it isn't really documenting the code well.  "we don't do periodic" is not the same as "we do oneshot and deadline"

src/platform/i386/lapic.c
```diff
@@ -86,26 +87,35 @@ lapic_read_reg(u32_t off)
...
+
+	rdtscll(now);
+	if (deadline <= now) return;
+
+	if (timer_type == ONESHOT) {
+
+		lapic_write_reg(LAPIC_INIT_COUNT_REG, (u32_t)(deadline - now));
```
gparmer: Does this really work?  I thought the granularity of the oneshot timer was != cycle.  Don't we need to do the calibration + math crap we had with the hpet?

src/platform/i386/lapic.c
```diff
@@ -0,0 +1,175 @@
...
+#define IA32_MSR_TSC		0x00000010
+#define IA32_MSR_TSC_DEADLINE	0x000006e0
+
+#define TEST_DEADLINE 3400000
+#define RETRY_ITERS   3
+
+extern int timer_process(struct pt_regs *regs);
+
+enum lapic_timer_type {
+	ONESHOT = 0,
```
gparmer: These need to be properly namespaced.  These will conflict with other identifiers trivially.  Should be `LAPIC_ONESHOT` etc.

src/platform/i386/lapic.c
```diff
@@ -0,0 +1,175 @@
...
+	PERIODIC,
+	TSC_DEADLINE,
+};
+
+enum lapic_timer_div_by_config {
+	DIV_BY_2 = 0,
```
gparmer: Same namespacing comment.

PR #203, opened 2016-09-30
Avoid retries in TSC_DEADLINE MSR Write

phanikishoreg:
### Summary of this PR

Intel's specification says writes to TSC_DEADLINE MSR are not serialized and therefore should not be used as serializing instruction. Serialization is only required between writes to TSC_DEADLINE MSR and memory mapped LVT REG, which isn't what I was doing. It introduces unnecessary race bugs and it also triggered the assert() in retry logic - when used in a full system of #rump2cos. 

To fix this, avoid Retries around TSC_DEADLINE MSR writes. (Also, retries weren't with relative times, absolute time retries could cause races especially if they are too close to current TSC). 

### Code Quality

Style:

Code Craftsmanship:

### DIFF
--- a/src/platform/i386/lapic.c
+++ b/src/platform/i386/lapic.c
@@ -23,8 +23,6 @@
 
 #define IA32_MSR_TSC_DEADLINE  0x000006e0
 
-#define RETRY_ITERS            3
-
 extern int timer_process(struct pt_regs *regs);
 
 enum lapic_timer_type {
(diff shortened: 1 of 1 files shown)


gparmer:
Makes sense to me.  The system calls will serialize the system, so unless we program the timer multiple times in a single system call, there is no reason to worry about serializing.

PR #207, opened 2016-10-09
Review request:

phanikishoreg:
### Summary of this PR

Following are the changes:
1. Transient priorities in tcaps using `cos_switch`
2. bug-fixes - mainly budget overflow on transfer
3. setting timers even in the edge cases, disabling timer if switching to INF tcap
4. budget expended logic in cap_switch 
I've tested these on micro_booter, vkernel and rumpkernel environments. 
### Code Quality

Style:

Code Craftsmanship:
- [x] I agree that there is no "throw-away" code, and that code in this PR is of high quality: **some parts of it is really hackish for now**

### DIFF
--- a/src/kernel/capinv.c
+++ b/src/kernel/capinv.c
@@ -496,6 +496,29 @@ cap_thd_switch(struct pt_regs *regs, struct thread *curr, struct thread  *next,
 	return preempt;
 }
 
+static struct thread *
+notify_parent(struct thread *rcv_thd)
+{
+	struct thread *curr_notif = NULL, *prev_notif = NULL, *arcv_notif = NULL;
+	int depth = 0;
+
+	/* hierarchical notifications - upto init (bounded by ARCV_NOTIF_DEPTH) */
+	prev_notif = rcv_thd;
+	curr_notif = arcv_notif = arcv_thd_notif(prev_notif);
+	while (curr_notif && curr_notif != prev_notif) {
+		assert(depth < ARCV_NOTIF_DEPTH);
+
+		thd_rcvcap_evt_enqueue(curr_notif, prev_notif);
+		if (!(curr_notif->state & THD_STATE_RCVING)) break;
+
+		prev_notif = curr_notif;
+		curr_notif = arcv_thd_notif(prev_notif);
+		depth ++;
+	}
+
+	return arcv_notif;
+}
+
 /**
  * Notify the appropriate end-points.
  * Return the thread that should be executed next.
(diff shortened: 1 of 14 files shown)


src/kernel/capinv.c
```diff
@@ -1409,6 +1422,21 @@ composite_syscall_slowpath(struct pt_regs *regs, int *thd_switch)
...
+
+				ret = cap_update(regs, thd, thd_next, tc_curr, tc_next, TCAP_TIME_NIL, ci, cos_info, 0);
+				if (unlikely(ret < 0)) cos_throw(err, ret);
+
+				*thd_switch = 1;
+			}
+
 			break;
 		}
 		case CAPTBL_OP_TCAP_DELEGATE:
```
gparmer: This has always been ridiculous to have in the switch.  Now it is quite complex so should obviously be a separate function.  You make the call when to do this.

src/kernel/capinv.c
```diff
@@ -1449,6 +1477,19 @@ composite_syscall_slowpath(struct pt_regs *regs, int *thd_switch)
 				if (unlikely(ret < 0)) cos_throw(err, ret);
 
 				*thd_switch = 1;
+			} else if(tcap_expended(tcap_current(cos_info))) {
+				struct thread *thd_next;
```
gparmer: Since this logic is identical to the logic above (see 1425), it should be in a function

src/kernel/include/tcap.h
```diff
@@ -190,32 +204,28 @@ tcap_timer_update(struct cos_cpu_local_info *cos_info, struct tcap *next, tcap_t
 
 	/* next == INF? no timer required. */
 	left        = tcap_left(next);
-	if (timeout == TCAP_TIME_NIL && TCAP_RES_IS_INF(left)) return;
+	if (timeout == TCAP_TIME_NIL && TCAP_RES_IS_INF(left)) {
+		if (cos_info->timeout_next) {
+			chal_timer_disable();
+			cos_info->timeout_next = 0;
```
gparmer: wraparound issues, or do we have logic to special-case `0`?

src/platform/i386/lapic.c
```diff
@@ -114,18 +118,31 @@ lapic_find_localaddr(void *l)
...
+}
+
+void
 lapic_set_timer(int timer_type, cycles_t deadline)
 {
 	u64_t now;
 
 	rdtscll(now);
-	if (deadline <= now) return;
+	if (deadline < now || (deadline - now) < LAPIC_TIMER_MIN) deadline = now + LAPIC_TIMER_MIN;
```
gparmer: 👍

gparmer: epic job debugging this.  We should have had assertions in from the start...

src/kernel/include/tcap.h
phanikishoreg: Actually, I think it's okay to get rid of timeout_next variable, it's use is only here for now. I should have the logic inside chal_timer_disable() to not redo if it's already disabled. 

I thought of wraparound issues, but I was thinking it will take too long for a 64bit value to wraparound. But yes, I need to take care of it anyway.

phanikishoreg:
Thank you. I've tested the vkernel and the micro_booter, they don't break. It's only that, after they pull, they'll need to make clean and make. So, that should be okay. But yes, I'll have it in my repo until the deadline and will push things after it. 

Also, I'm implementing `cos_asnd()` with yield, as it sounds like it is very much required for simple vkernel scheduling, if it doesn't want to give out excess budgets, each time. I've it ready. Will push along with these changes and your review comments. But will not do a pull until after the deadline. 

Thank you for the quick review!

gparmer:
Ready to accept the PR whenever you're ready.

PR #208, opened 2016-11-02
Cbuf

ryuxin:
### Summary of this PR

Cbuf code

### Code Quality

Style:

Code Craftsmanship:


src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
```diff
@@ -9,16 +9,27 @@
...
 #include <cmap.h>
 #include <cos_list.h>
 
 #include <stkmgr.h>
 
+/* debug helper */
+#define OP_NUM 10
+/* 0-create 1-collect 2-delete */
+/* 3-retrieve 4-register */
+/* 56-reserve 7-cbuf_alloc_map */
```
gedare: My preference would be an enum or set of macro defines for the distinct values that can be taken on for this "op_num" stuff. OP_NUM here is the total number of cbuf operations, right? I had to infer that.

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
```diff
@@ -9,16 +9,27 @@
...
+int op_nums[OP_NUM];
```
gedare: here op_nums whould really be op_cnts since this is an array of counters. overusing OP_NUM as the max ops and op_nums as the count of how many times operations are invoked is confusing.

gedare: also, if it cannot be negative, I suggest using unsigned here.

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
```diff
@@ -9,16 +9,27 @@
...
+unsigned long long per_total[OP_NUM];
```
gedare: It is not at all clear from the name that this is tracking time. I recommend making the type match with the calculations (u64_t), and adjust the name to be more clear that this is tracking time, e.g. total_tsc_per_op or something similar.

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
```diff
@@ -9,16 +9,27 @@
...
 #include <cos_synchronization.h>
 #include <valloc.h>
+#include <sched.h>
 #include <cos_alloc.h>
 #include <cmap.h>
 #include <cos_list.h>
 
 #include <stkmgr.h>
 
+/* debug helper */
```
gedare: Perhaps these should be ifdef'd on COS_DEBUG?

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
```diff
@@ -9,16 +9,27 @@
...
+#define OP_NUM 10
+/* 0-create 1-collect 2-delete */
+/* 3-retrieve 4-register */
+/* 56-reserve 7-cbuf_alloc_map */
+int op_nums[OP_NUM];
+unsigned long long per_total[OP_NUM];
+
+#define INIT_TARGET_SIZE 4096*2048
```
gedare: Why this number? This is 8 MiB for some reason, but why?

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
```diff
@@ -62,8 +73,8 @@ struct cbuf_maps {
 };
 
 struct cbuf_info {
-	u32_t cbid;
-	int size;
+	int cbid;
+	unsigned long size;
```
gedare: size_t may also be appropriate.

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
```diff
@@ -62,8 +73,8 @@ struct cbuf_maps {
 };
 
 struct cbuf_info {
-	u32_t cbid;
-	int size;
+	int cbid;
```
gedare: Why changing cbid to a signed representation? Do we use negative cbids somewhere? I believe the cbid is really a 30-bit unsigned number.

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
```diff
@@ -73,23 +84,36 @@ struct cbuf_info {
...
+#define CBUF_META_RANGE_HIGH(cmr) (cmr->low_id + (int)(PAGE_SIZE/sizeof(struct cbuf_meta)))
 
 struct cbuf_bin {
-	int size;
+	unsigned long size;
 	struct cbuf_info *c;
 };
 
+struct blocked_thd {
+	unsigned short int thd_id;
```
gedare: isn't there a tid_t type to use?

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
```diff
@@ -73,23 +84,36 @@ struct cbuf_info {
...
+	unsigned long request_size;
+	struct blocked_thd *next, *prev;
+	unsigned long long blk_start;
```
gedare: I guess this is a 'time' field, i suggest using u64_t and saying, blk_start_time in the time.

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
```diff
@@ -73,23 +84,36 @@ struct cbuf_info {
...
+};
+
 struct cbuf_comp_info {
 	spdid_t spdid;
 	struct cbuf_shared_page *csp;
 	vaddr_t dest_csp;
 	int nbin;
 	struct cbuf_bin cbufs[CBUF_MAX_NSZ];
 	struct cbuf_meta_range *cbuf_metas;
+	unsigned long target_size, allocated_size;
```
gedare: This may be me, but I don't like to put multiple fields in the same row in a struct definition. I don't know if this is covered or not in the style guide.

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
```diff
@@ -73,23 +84,36 @@ struct cbuf_info {
...
+	int num_blocked_thds;
+	struct blocked_thd bthd_list;
+	unsigned long long gc_tot, gc_max;
+	unsigned long long blk_tot, blk_max;
```
gedare: time fields again?

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
```diff
@@ -73,23 +84,36 @@ struct cbuf_info {
...
+	int gc_num;
```
gedare: is this a counter of something (I guess garbage-collected cbufs?)

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
gedare: I can't find where allocated_size is initialized.

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
```diff
@@ -190,86 +215,148 @@ cbuf_comp_info_bin_get(struct cbuf_comp_info *cci, int sz)
...
 	return &cci->cbufs[cci->nbin-1];
 }
 
 static int
-cbuf_map(spdid_t spdid, vaddr_t daddr, void *page, int size, int flags)
+cbuf_map(spdid_t spdid, vaddr_t daddr, void *page, unsigned long size, int flags)
 {
-	int off;
-	assert(size == (int)round_to_page(size));
+	unsigned long off;
```
gedare: Do we have off_t available? otherwise, u64_t is appropriate or uintptr_t.

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
```diff
@@ -190,86 +215,148 @@ cbuf_comp_info_bin_get(struct cbuf_comp_info *cci, int sz)
...
 	assert(page);
 	for (off = 0 ; off < size ; off += PAGE_SIZE) {
 		vaddr_t d = daddr + off;
-		if (d != (mman_alias_page(cos_spd_id(), ((vaddr_t)page) + off,
-						spdid, d, flags))) {
-			printd("couldn't alias(%d, %x, %d, %x, %d)\n", cos_spd_id(), ((vaddr_t)page) + off, spdid, d, flags);
-			assert(0); /* TODO: roll back the aliases, etc... */
+		if (unlikely(d != (mman_alias_page(cos_spd_id(), ((vaddr_t)page) + off,
+						   spdid, d, flags)))) {
+			for(d = daddr+off-PAGE_SIZE; d >= daddr; d -= PAGE_SIZE) {
```
gedare: nit: space between for(

gedare: and spaces around the math operators in daddr+off-PAGE_SIZE

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
```diff
@@ -190,86 +215,148 @@ cbuf_comp_info_bin_get(struct cbuf_comp_info *cci, int sz)
...
 	}
 	return 0;
 }
 
+/* map the memory from address p and size sz to the component spdid with 
+ * permission flags. if p is NULL allocate a piece of new memory
+ * return spdid's address to daddr, manager's virtual address to page*/
 static int
-cbuf_alloc_map(spdid_t spdid, vaddr_t *daddr, void **page, int size)
+cbuf_alloc_map(spdid_t spdid, vaddr_t *daddr, void **page, void *p, unsigned long sz, int flags)
```
gedare: It looks like there are no uses where p is actually passed. If it is unused I wouldn't keep it. The function is supposed to alloc and map. If you don't need to alloc, you could just map...

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
```diff
@@ -190,86 +215,148 @@ cbuf_comp_info_bin_get(struct cbuf_comp_info *cci, int sz)
...
 	ret = -1;
-done:
-	*page  = p;
+
+done:	
+	if (page) *page  = new_p;
 	*daddr = dest;
+	rdtscll(end);
+	op_nums[7]++;
+	per_total[7] += (end-start);
```
gedare: I'd put this op_nums and per_total in an inline function that does nothing if not DEBUG mode. e.g. op_end(CBUF_ALLOC_MAP) to increment the op_nums[CBUF_ALLOC_MAP] and per_total[...] (assuming you add enum or macros for CBUF_ALLOC_MAP).

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
```diff
@@ -190,86 +215,148 @@ cbuf_comp_info_bin_get(struct cbuf_comp_info *cci, int sz)
...
+		/* Guarantee atomically read the two words (refcnt and nsent/nrecv).
+		 * Consider this case, c0 sends a cbuf to c1 and frees this 
+		 * this cbuf, but before c1 receives it, the manager comes in 
+		 * and checks c1's refcnt. Now it is zero. But before the manager 
+		 * checks c1's nsent/nrecv, it is preempted by c1. c1 receives 
+		 * this cbuf--increment refcnt and nsent/nrecv. After this, we 
+		 * switch back the manager, who will continues to check c1's 
+		 * nsent/nrecv, now it is 1, which is equal to c0's nsent/nrecv. 
+		 * Thus the manager can collect or unmap this cbuf.*/
+		old = *(unsigned long long *)meta;
```
gedare: This is a bit hackish and works because meta->nfo and meta->snd_rcv are within the same 64-bits in the `struct meta`. Is this implicit requirement checked or enforced anywhere? What prevents someone from adding another field to `struct meta` that would break this code?

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
```diff
@@ -190,86 +215,148 @@ cbuf_comp_info_bin_get(struct cbuf_comp_info *cci, int sz)
...
+	unsigned long long old;
+	struct cbuf_meta *mt, *own_mt = m->m;
+
+	old_nfo = own_mt->nfo;
+	new_nfo = old_nfo | CBUF_INCONSISENT;
+	if (unlikely(!cos_cas(&own_mt->nfo, old_nfo, new_nfo))) {
+		goto done;
+	}
 
+	mt   = (struct cbuf_meta *)(&old);
```
gedare: This is not pointing to an actual `struct cbuf_meta`, but a copy of the first 64-bits of one. It may be worth making a comment about it.

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
```diff
@@ -354,104 +459,192 @@ cbuf_mark_relinquish_all(struct cbuf_comp_info *cci)
...
-		if (!cbi) goto done;
 
-		/* Allocate and map in the cbuf. */
-		cbid = cmap_add(&cbufs, cbi);
+		if (unlikely(!cbi)) goto done;
+		/* Allocate and map in the cbuf. Discard inconsistent cbufs */
+		/* TODO: Find a better way to manage those inconsistent cbufs */
+		do {
+			cbid = cmap_add(&cbufs, cbi);
+			meta = cbuf_meta_lookup(cci, cbid);
```
gedare: Should you cmap_del() in case of inconsistent meta?

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
```diff
@@ -755,18 +1009,97 @@ cbuf_register(spdid_t spdid, long cbid)
...
+	struct cbuf_bin *bin;
+	struct cbuf_info *cbi, *next, *head;
+
+	for(i=cci->nbin-1; i>=0; i--) {
+		bin = &cci->cbufs[i];
+		sz = (int)bin->size;
+		if (!bin->c) continue;
+		cbi = FIRST_LIST(bin->c, next, prev);
+		int i = 0;
+		while (cbi != bin->c) {
```
gedare: you can simplify this a bit with `cbi = bin->c` and then do { ... } while (cbi != bin->c);` to remove the redundant code that calls `cbuf_free_unmap` after the loop ends.

src/components/include/cbuf.h
```diff
@@ -141,64 +125,76 @@
...
 	cbuf_t v;
 	struct {
 		/* 
-		 * cbuf id, aggregate = 1 if this cbuf holds an array of other
-		 * cbufs, tm = 1 if this cbuf is transient (non-persistent)
+		 * cbuf id, aggregate = 1 if this cbuf 
+		 * holds an array of other cbufs
 		 */
-		u32_t aggregate: 1, tm: 1, id:30;
+		u32_t id:31, aggregate: 1;
```
gedare: this makes a `cbuf_t` for a non-aggregate cbuf identically equal to an unsigned cbid (for x86 / little endian). Though if you treat it as a signed integer the equality is not the same (for cbids > 2^24) I guess. I don't know if this equality is a good thing, as it can permit some sloppiness to use cbuf_t in place of cbid or vice-versa, which will be right "most of the time" now.

src/components/include/cbuf.h
```diff
@@ -235,83 +231,94 @@ cbuf_cons(u32_t cbid)
...
 	assert(CBUF_REFCNT(cm));
-	assert(cm->snd_rcv.nsent < CBUF_SENDRECV_MAX);
 
-	CBUF_NSND_INC(cm);
-	if (free) CBUF_REFCNT_ATOMIC_DEC(cm);
-	/* really should deal with this case correctly */
-	assert(!CBUF_RELINQ(cm));     //???
+	int old = cm->snd_rcv.nsent;
+	CBUF_NSND_ATOMIC_INC(cm);
+	if (!(cm->snd_rcv.nsent > old)) printc("spd %d thd %d cb %d sent %d old %d\n", cos_spd_id(), cos_get_thd_id(), cb, cm->snd_rcv.nsent, old);
```
gedare: long line.

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
georgit: Let's be careful here because cbuf_create does return a negative number sometimes. But that might not necessarily mean cbids should be signed. Just those return types, and the variables that handle them.

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
```diff
@@ -190,86 +215,148 @@ cbuf_comp_info_bin_get(struct cbuf_comp_info *cci, int sz)
...
+	mt   = (struct cbuf_meta *)(&old);
 	sent = recvd = 0;
 	do {
 		struct cbuf_meta *meta = m->m;
 
-		if (meta) {
-			if (CBUF_REFCNT(meta)) return 1;
-			sent  += meta->snd_rcv.nsent;
-			recvd += meta->snd_rcv.nrecvd;
+		/* Guarantee atomically read the two words (refcnt and nsent/nrecv).
```
georgit: I thought the style guide said to form comments with /* on a lone line, but I can't find this now. I guess this is fine.

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
```diff
@@ -287,31 +374,52 @@ cbuf_references_clear(struct cbuf_info *cbi)
 	return;
 }
 
-static void
-cbuf_free_unmap(spdid_t spdid, struct cbuf_info *cbi)
+/* Before actually unmap cbuf from a component, we need to atomically
+ * clear the page pointer in the meta, which guarantees that clients 
+ * do not have seg fault. Clients have to check NULL when receive cbuf */
+static int
+cbuf_unmap_prepare(struct cbuf_info *cbi)
```
georgit: How will this affect cbuf_maps structs and cbuf_metas? Let's discuss tomorrow.

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
```diff
@@ -354,104 +459,192 @@ cbuf_mark_relinquish_all(struct cbuf_comp_info *cci)
...
+{
+	struct blocked_thd bthd;
+
+	bthd.thd_id       = cos_get_thd_id();
+	bthd.request_size = request_size;
+	rdtscll(bthd.blk_start);
+	ADD_LIST(&cci->bthd_list, &bthd, next, prev);
+	cci->num_blocked_thds++;
+	cbuf_mark_relinquish_all(cci);
+	CBUF_RELEASE();
```
georgit: So this function requires that a lock be taken first?

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
georgit: Hmm... interesting. So per_total tracks how much time creating a cbuf takes. Is this purely for debugging?

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
```diff
@@ -802,20 +1135,132 @@ stkmgr_grant_stack(spdid_t d_spdid)
...
+	memset(per_total, 0, sizeof(per_total));
+}
+
+
+/* Debug helper functions */
+static int __debug_reference(struct cbuf_info * cbi)
+{
+	struct cbuf_maps *m = &cbi->owner;
+	int sent = 0, recvd = 0;
+	do {
```
gparmer: empty line after variable declarations.

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
```diff
@@ -802,20 +1135,132 @@ stkmgr_grant_stack(spdid_t d_spdid)
...
+		struct cbuf_meta *meta = m->m;
```
gparmer: same

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
```diff
@@ -802,20 +1135,132 @@ stkmgr_grant_stack(spdid_t d_spdid)
...
+		if (CBUF_REFCNT(meta)) return 1;
+		sent  += meta->snd_rcv.nsent;
+		recvd += meta->snd_rcv.nrecvd;
+		m   = FIRST_LIST(m, next, prev);
```
gparmer: if the `=` isn't aligned, then don't put extra spaces.

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
```diff
@@ -802,20 +1135,132 @@ stkmgr_grant_stack(spdid_t d_spdid)
...
+	} while (m != &cbi->owner);
+	if (sent != recvd) return 1;
+	return 0;
```
gparmer: space before the return.

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
```diff
@@ -802,20 +1135,132 @@ stkmgr_grant_stack(spdid_t d_spdid)
...
+	cmap_add(&cbufs, NULL);
+
+	/* debug */
+	memset(op_nums, 0, sizeof(op_nums));
+	memset(per_total, 0, sizeof(per_total));
+}
+
+
+/* Debug helper functions */
+static int __debug_reference(struct cbuf_info * cbi)
```
gparmer: separate line for `static int`

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
```diff
@@ -802,20 +1135,132 @@ stkmgr_grant_stack(spdid_t d_spdid)
...
+		if (CBUF_REFCNT(meta)) return 1;
+		sent  += meta->snd_rcv.nsent;
+		recvd += meta->snd_rcv.nrecvd;
+		m   = FIRST_LIST(m, next, prev);
+	} while (m != &cbi->owner);
+	if (sent != recvd) return 1;
+	return 0;
+}
+
+unsigned long cbuf_debug_cbuf_info(spdid_t spdid, int index, int p)
```
gparmer: same style issues here.

src/components/implementation/no_interface/cbboot/lock.c
```diff
@@ -183,7 +183,7 @@ static inline void __lock_help(struct meta_lock *l, int smp)
 	do {
 		prev_val.v         = l->lock_addr->v;
 		owner              = prev_val.c.owner;
-		assert(owner); /* from lock_help_owner, this must be true */
+		if (owner) continue;
```
gparmer: doesn't this cause a ton of wasted spinning?

gparmer: I guess I've honestly never thought about how `continue` and `do ... while` interact.

src/components/include/cbuf.h
gparmer: this is a great point @gedare.  should we have aggregate as the LSB instead?

src/components/include/cbuf.h
```diff
@@ -224,7 +220,7 @@ cbuf_unpack(cbuf_t cb, u32_t *cbid)
 }
 
 static inline cbuf_t 
-cbuf_cons(u32_t cbid)
+cbuf_cons(int cbid)
```
gparmer: again, I'm not seeing the reasoning behind this.  I'm sure it is there, but I don't understand.

src/components/include/cbuf.h
gparmer: ...and should the printc be paired with some sort of logic?  Error out, handle the situation somehow?  `printc` alone doesn't seem to be that useful.

src/components/include/cbuf.h
```diff
@@ -466,7 +550,44 @@ cbuf_free(cbuf_t cb)
...
 cbuf_is_cbuf(void *buf)
 {
-	return false;   //???
+	/* TODO: After using cbuf to back all memory, this is true */
+	return false;
+}
+
+
+/*===== Debug functions ======*/
+static int cbuf_debug_freelist_num(unsigned long sz)
```
gparmer: style throughout these functions...

src/components/include/cbuf.h
```diff
@@ -466,7 +550,44 @@ cbuf_free(cbuf_t cb)
...
+{
+	struct cbuf_meta *cm, *fl;
+	int ret = 0;
+	sz = nlepow2(round_up_to_page(sz));
+	fl = __cbuf_freelist_get(sz);
+	cm = fl->next;
+	while (cm != fl) {
```
gparmer: maybe better as a `for(...) ;`?

src/components/include/cbuf.h
```diff
@@ -466,7 +550,44 @@ cbuf_free(cbuf_t cb)
...
+		ret++;
+		cm = cm->next;
+	}
+	return ret;
+}
+
+static void cbuf_debug_clear_freelist(unsigned long sz)
```
gparmer: This is a wall of code.

Please add empty lines.

src/components/interface/cbuf_mgr/cbuf_c.c
```diff
@@ -58,121 +59,120 @@ __cbuf_2buf_miss(int cbid, int len)
...
-			} while(unlikely(!cos_dcas(target, *old, *update)));
+			/* own the cbuf we just collected */
+			ret_cm = cbuf_vect_lookup_addr(el.cbid);
+			assert(ret_cm && ret_cm->cbid_tag.cbid == el.cbid);
+		} else {
+			break;
 		}
+	} while(!__cbuf_try_take(ret_cm, flag));
+
+	/* ...add the rest back into freelists. construct a temporary freelist
```
gparmer: style

src/components/interface/cbuf_mgr/cbuf_c.c
```diff
@@ -58,121 +59,120 @@ __cbuf_2buf_miss(int cbid, int len)
...
-	cm   = cbuf_vect_lookup_addr(cbid);
-	assert(cm && CBUF_PTR(cm));
-	assert(cm && CBUF_REFCNT(cm));
+	csp = (struct cbuf_shared_page*)cbuf_map_collect(cos_spd_id());
+again:
+	ret_cm = NULL;
+	cbid = 0;
+
+	/* Do a garbage collection if this not an exact size cbuf*/
+	if (!(flag & CBUF_EXACTSZ)) {
```
gparmer: Please add empty lines.

src/components/interface/cbuf_mgr/cbuf_c.c
```diff
@@ -58,121 +59,120 @@ __cbuf_2buf_miss(int cbid, int len)
...
+		ret_cm = __cbuf_get_collect(csp, size, flag);
+		if (ret_cm) goto done;
+		amnt = cbuf_collect(cos_spd_id(), size);
```
gparmer: line up `=`s

src/components/interface/cbuf_mgr/cbuf_c.c
```diff
@@ -58,121 +59,120 @@ __cbuf_2buf_miss(int cbid, int len)
...
+		ret_cm = __cbuf_get_collect(csp, size, flag);
+		if (!ret_cm) ret_cm = __cbuf_freelist_pop(size, flag);
+		else 	assert(ret_cm && CBUF_PTR(ret_cm));
+	}
+	
+	/* Nothing collected...allocate a new cbuf! */
+	if (!ret_cm) {
+		cbid = cbuf_create(cos_spd_id(), size, cbid*-1);
+		/* We return from being blocked */
+		if (cbid == 0) goto again;
```
gparmer: Use `{}` in both `if` and `else`, or not in either.

src/components/interface/cbuf_mgr/cbuf_c.c
```diff
@@ -58,121 +59,120 @@ __cbuf_2buf_miss(int cbid, int len)
...
 
 CCTOR static void
 cbuf_init(void)
 {
-	//initialize freelist head
-	return ;  //???
+	struct cbuf_freelist *fl = PERCPU_GET(cbuf_alloc_freelists);
+	struct cbuf_meta *m;
+	int i;
+	for(i=0; i<CBUF_MAX_NSZ; i++) {
```
gparmer: spaces around `=` and `;`

src/components/interface/cbuf_mgr/cbuf_meta.h
```diff
@@ -9,41 +9,50 @@
...
-#define CBUF_REFCNT(meta)            (((meta)->nfo>>CBUF_FLAGS_SZ) & 0xFF)
-#define CBUF_REFCNT_ATOMIC_INC(meta) cos_faa((int *)(&((meta)->nfo)), CBUF_INC_UNIT)
-#define CBUF_REFCNT_ATOMIC_DEC(meta) cos_faa((int *)(&((meta)->nfo)), CBUF_DEC_UNIT)
-#define CBUF_PTR(meta)               (((meta)->nfo)&(~CBUF_PTR_MASK))
-#define CBUF_PTR_SET(meta, v)        ((meta)->nfo |= (v))
-#define CBUF_NSND_INC(meta)          ((meta)->snd_rcv.nsent++)
-#define CBUF_NRCV_INC(meta)          ((meta)->snd_rcv.nrecvd++)
-#define CBUF_IS_MAPPED(meta)         (CBUF_PTR(meta) != 0)
-#define CBUF_IS_IN_FREELiST(meta)    ((meta)->next != NULL)
+#define CBUF_PTR_SET(meta, v)         do {   \
```
gparmer: line this up so that it looks like code.

src/components/interface/cbuf_mgr/cbuf_mgr.h
```diff
@@ -13,40 +13,49 @@
...
+vaddr_t cbuf_register(spdid_t spdid, int cbid);
 
-/* Map a cbufp into another component at a given address.
+/* Map a cbuf into another component at a given address.
  * The s_spd that calls this function should ensure the memory is not freed.
  * The d_addr must be alloced with sufficient pages to contain the cbuf.
  */
-vaddr_t cbuf_map_at(spdid_t s_spd, cbuf_t cbid, spdid_t d_spd, vaddr_t d_addr);
-int cbuf_unmap_at(spdid_t s_spd, cbuf_t cbid, spdid_t d_spd, vaddr_t d_addr);
+vaddr_t cbuf_map_at(spdid_t s_spd, int cbid, spdid_t d_spd, vaddr_t d_addr);
```
gparmer: now we're getting rid of the `cbuf_t`s?

src/components/interface/cbuf_mgr/cbuf_vect.h
```diff
@@ -19,15 +19,15 @@
...
 extern cvect_t meta_cbuf;
 static inline int
-__cbuf_vect_expand_rec(struct cvect_intern *vi, const long id, const int depth)
+__cbuf_vect_expand_rec(struct cvect_intern *vi, const int id, const int depth)
 {
 	struct cvect_intern *new;
 
 	if (depth > 1) {
-		long n = id >> (CVECT_SHIFT * (depth-1));
+		int n = id >> (CVECT_SHIFT * (depth-1));
```
gparmer: unsigned?  Shifting with a signed is generally undefined.

src/components/include/cbuf.h
gedare: putting it in LSb makes detecting misuse of the cbuf_t type easier to detect, yes.

src/components/include/cbuf.h
```diff
@@ -343,121 +350,198 @@ PERCPU_EXTERN(cbuf_alloc_freelists);
 static inline struct cbuf_meta *
 __cbuf_freelist_get(int size)
 {
+	/* We have per-core, per-component, per-size free-list */
 	struct cbuf_freelist *fl = PERCPU_GET(cbuf_alloc_freelists);
 	int order = ones(size-1) - PAGE_ORDER;
 	struct cbuf_meta *m;
+	if (!(pow2(size) && size >= PAGE_SIZE)) {
```
gedare: ifdef DEBUG?

src/components/include/cbuf.h
```diff
@@ -343,121 +350,198 @@ PERCPU_EXTERN(cbuf_alloc_freelists);
...
+		printc("cbuf free size %d spd %d\n", size, cos_spd_id());
+	}
 	assert(pow2(size) && size >= PAGE_SIZE);
 	assert(order >= 0 && order < WORD_SIZE-PAGE_ORDER);
 	m = &(fl->freelist_head[order]);
-	if (unlikely(!m->next)) m->next = m;
+	assert(m->next);
```
gedare: what does the last node in the freelist point to for next? I'm not sure why m->next is asserted here.

src/components/include/cbuf.h
```diff
@@ -343,121 +350,198 @@ PERCPU_EXTERN(cbuf_alloc_freelists);
...
+		 * No cbuf leak */
+		if (CBUF_INCONSISENT(cm)) CBUF_FLAG_REM(cm, CBUF_INCONSISENT);
+		goto ret;
+	}
+	r = 1;
+	/* guarantee the order between CAS and set "next".
+	 * If the next is set to NULL before increment refcnt, the
+	 * manager may think this cbuf is garbage and collect it. */
+	/* But there is dependency between those two, we do not 
+	 * need explicit barrier. */
```
gedare: if the "dependency" is through the *cm, a very-smart compiler might be able to optimize and re-order cm->next before the cas...

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
ryuxin: add ifedf for DEBUG

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
ryuxin: fixed

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
ryuxin: fixed

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
ryuxin: this is the default size of cbuf pool for each components.
There is no hard reason for this value, it just cannot be too small.

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
ryuxin: allocated_size is initialized by the memset of the structure.
I will separate it into anther struct

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
ryuxin: I do not see we have off_t.
we can create it if we use it lot

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
georgit: We did not find this. cos_get_thd_id returns an unsigned short int. I always just thought these were ints.

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
ryuxin: p is indeed used

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
ryuxin: fixed

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
```diff
@@ -190,86 +215,148 @@ cbuf_comp_info_bin_get(struct cbuf_comp_info *cci, int sz)
...
+ * However, if clients do not use protocol correctly, this function 
+ * provides no guarantee. cbuf_unmap_prepare takes care of this case.
+ */
 static int
 cbuf_referenced(struct cbuf_info *cbi)
 {
 	struct cbuf_maps *m = &cbi->owner;
-	int sent, recvd;
+	int sent, recvd, ret = 1;
+	unsigned long old_nfo, new_nfo;
```
ryuxin: I change all meta.nfo to be unsigned long

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
ryuxin: change to memcpy

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
ryuxin: yes, we need cmap_del.
But we should record all inconsistent meta first, then del them together.
we cannot do cmap_del in the loop, as we will add to that slot again.

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
ryuxin: This dos not work. Because we will remove the item from the list inside the cbuf_free_unmap and 
modify the header to the list.

src/components/include/cbuf.h
ryuxin: I do not think so.
From my experience, putting it in LSB cause some inconvenience.
I took me much time to figure out the wired cbid value.
although it helps to detect misuse, it is not the only way.

src/components/include/cbuf.h
ryuxin: This is hard.
I do not have good way to handle FAA overflow, because when we know the overflow, it has 
happened. Undo is not always possible.
This is why I have an assertion there.
we can discuss it if we want.

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
ryuxin: I rewrite all those debug stuff.
Your suggestion is great

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
ryuxin: changed the name

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
```diff
@@ -190,86 +215,148 @@ cbuf_comp_info_bin_get(struct cbuf_comp_info *cci, int sz)
...
 {
-	void *p;
 	vaddr_t dest;
 	int ret = 0;
+	void *new_p;
+
+	u64_t start, end;
+	rdtscll(start);
+
+	assert(sz == round_to_page(sz));
```
ryuxin: I use it as debug.
I should check it in the interface, and here it is used to confirm I have check it properly

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
```diff
@@ -190,86 +215,148 @@ cbuf_comp_info_bin_get(struct cbuf_comp_info *cci, int sz)
...
+	if (!p) page_free(new_p, sz/PAGE_SIZE);
 	ret = -1;
-done:
-	*page  = p;
+
+done:	
+	if (page) *page  = new_p;
 	*daddr = dest;
+	rdtscll(end);
+	op_nums[7]++;
```
ryuxin: this is great.
thanks

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
```diff
@@ -190,86 +215,148 @@ cbuf_comp_info_bin_get(struct cbuf_comp_info *cci, int sz)
...
+		 * nsent/nrecv, now it is 1, which is equal to c0's nsent/nrecv. 
+		 * Thus the manager can collect or unmap this cbuf.*/
+		old = *(unsigned long long *)meta;
+		if (unlikely(!cos_dcas(&old, old, old))) {
+			goto unset;
 		}
-
-		m = FIRST_LIST(m, next, prev);
+		if (CBUF_REFCNT(mt)) goto unset;		
+		/* TODO: add per-mapping counter of sent and recv in the manager */
```
ryuxin: no, this is a potential optimization

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
```diff
@@ -287,31 +374,52 @@ cbuf_references_clear(struct cbuf_info *cbi)
...
+	struct cbuf_maps *m = &cbi->owner, *next;
+	struct cbuf_bin *bin;
+	void *ptr = cbi->mem;
+	unsigned long off, size = cbi->size;
+
+	if (cbuf_unmap_prepare(cbi)) return 1;
+
 	/* Unmap all of the pages from the clients */
-	for (off = 0 ; off < cbi->size ; off += PAGE_SIZE) {
+	for (off = 0 ; off < size ; off += PAGE_SIZE) {
```
ryuxin: these two are not exactly the same.
But I agree such function is useful

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
```diff
@@ -321,31 +429,28 @@ cbuf_free_unmap(spdid_t spdid, struct cbuf_info *cbi)
...
 
 	/* deallocate/unlink our data-structures */
-	page_free(ptr, cbi->size/PAGE_SIZE);
+	page_free(ptr, size/PAGE_SIZE);
 	cmap_del(&cbufs, cbi->cbid);
-	cci = cbuf_comp_info_get(spdid);
-	if (unlikely(!cci)) return ;
-	bin = cbuf_comp_info_bin_get(cci, cbi->size);
+	cci->allocated_size -= size;
+	bin = cbuf_comp_info_bin_get(cci, size);
```
ryuxin: I see no way it could fail unless we have some bugs

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
```diff
@@ -354,104 +459,192 @@ cbuf_mark_relinquish_all(struct cbuf_comp_info *cci)
...
+	assert(cci->target_size >= cci->allocated_size);
+
+	if (cci->num_blocked_thds == 0) return;
+	bthd = cci->bthd_list.next;
+	while (bthd != &cci->bthd_list) {
+		next = FIRST_LIST(bthd, next, prev);
+		if (bthd->request_size <= sz) {
+			REM_LIST(bthd, next, prev);
+			cci->num_blocked_thds--;
+			rdtscll(cur);
```
ryuxin: It is not meaningful all the time.
And It depends on how policy use those information

src/components/interface/cbuf_mgr/cbuf_mgr.h
ryuxin: yes, the manager interface should not use cbuf_t

src/components/include/cbuf.h
ryuxin: I remember there is a control flow dependency due to the if statement.
Maybe it is different for different compilers

src/components/include/cbuf.h
ryuxin: this is a circular linked list

src/components/implementation/cbuf_mgr/default/cbuf_mgr.c
ryuxin: in the cbuf_create, it should be int.
I afraid we need to cast a lot if we changed it to be unsigned.

georgit: While we agree in the cbuf_info struct this field should be unsigned, we think it would take a while to fix this throughout the code, while also minimizing casts. For now, Yuxin will attempt to merge as-is, and I will try to fix on my end. This might also be a big enough issue to discuss in person.

src/components/interface/cbuf_mgr/cbuf_vect.h
ryuxin: fixed

PR #213, opened 2016-12-27
Review Request: Default API proposed changes

phanikishoreg:
* API revised for 2nd draft.
* Current thread function is not changed.
* TODO: a way to avoid increase in memory for __thd_init_data.

### Summary of this PR

Review request for New Default API. 
Current thread functions are not changed, they are typed `THD_NORMAL` in the `__thd_init_data` struct and the new typed thread are `THD_AEP`. 
I still need to think about a way to avoid having `rcvcap` and `type` in `__thd_init_data`, I could not find a solution yet. 
**TODO: API definition.**

### Code Quality

Style:

- [+] Comments adhere to the Style Guide (SG)
- [+] Spacing adhere's to the SG 
- [+] Naming adhere's to the SG
- [+] All other aspects of the SG are adhered to, or exceptions are justified in this pull request

Code Craftsmanship: **TODO**

### DIFF
--- a/src/components/include/cos_component.h
+++ b/src/components/include/cos_component.h
@@ -527,8 +527,10 @@ struct cos_array { char *mem; int sz; }; /* TODO: remove */
 #endif
 
 struct __thd_init_data {
-	void *fn;
-	void *data;
+	int     type;
+	void   *fn;
+	capid_t rcv;
+	void   *data;
 };
 
 #endif
--- a/src/components/include/cos_kernel_api.h
+++ b/src/components/include/cos_kernel_api.h
@@ -107,5 +107,63 @@ int cos_hw_detach(hwcap_t hwc, hwid_t hwid);
 void *cos_hw_map(struct cos_compinfo *ci, hwcap_t hwc, paddr_t pa, unsigned int len);
 int cos_hw_cycles_per_usec(hwcap_t hwc);
 
+/* Default API */
+/* Capabilities for Async activation end point */
+struct cos_aep_info {
+	tcap_t    tc;
+	thdcap_t  thd;
+	arcvcap_t rcv;
+};
+
+/* Default Component information */
+struct cos_defci {
+	struct cos_compinfo ci;
+	struct cos_aep_info sched_aep;
+};
+
+/* current component's default component information */
+extern struct cos_defci curr_defci;
+
+/*
+ * cos_defci_init: initialize the current component's global cos_defci struct using the standard boot capabilities layout.
+ */
+void cos_defci_init(void);
+/*
+ * cos_defci_init_ext: initialize the current component's global cos_defci struct using the parameters passed.
+ */
+void cos_defci_init_ext(tcap_t sched_tc, thdcap_t sched_thd, arcvcap_t sched_rcv, pgtblcap_t pgtbl_cap, captblcap_t captbl_cap, compcap_t comp_cap, vaddr_t heap_ptr, capid_t cap_frontier);
+/*
+ * cos_compinfo_get: returns the cos_compinfo pointer that points to the cos_compinfo struct inside cos_defci.
+ */
+struct cos_compinfo *cos_compinfo_get(void);
+/*
+ * cos_sched_aep_get: returns the AEP pointer that points to the current component's scheduler.
+ */
+struct cos_aep_info *cos_sched_aep_get(void);
+
+/*
+ * cos_defci_child_alloc: this is called to create a child component with new initial capabilities like pgtbl, captbl, compcap, etc.
+ *                        if is_sched is set, scheduling end_point is created for the new component, 
+ *                        else current component's scheduler is used as scheduler for the child component.
+ */
+int cos_defci_child_alloc(struct cos_defci *child_defci, vaddr_t entry, vaddr_t heap_ptr, capid_t cap_frontier, int is_sched);
+
+/*
+ * thread function that takes it's async rcv end-point as an argument along with user-passed void *
+ */
+typedef void (*cos_aepthd_fn_t)(arcvcap_t, void *);
+/*
+ * cos_aep_alloc: creates a new async activation end-point which includes thread, tcap and rcv capabilities. 
+ */
+int cos_aep_alloc(struct cos_aep_info *aep, cos_aepthd_fn_t fn, void *data);
+/*
+ * cos_aep_alloc: creates a new async activation end-point, using an existing tcap. 
+ */
+int cos_aep_tcap_alloc(struct cos_aep_info *aep, tcap_t tc, cos_aepthd_fn_t fn, void *data);
+
+/*
+ * cos_defswitch: thread switch api using the default scheduling tcap and rcv.
+ */
+int cos_defswitch(thdcap_t c, tcap_prio_t p, tcap_time_t r, sched_tok_t stok);
 
 #endif /* COS_KERNEL_API_H */
--- a/src/components/include/cos_thd_init.h
+++ b/src/components/include/cos_thd_init.h
@@ -7,7 +7,7 @@
 extern struct __thd_init_data __thd_init_data[COS_THD_INIT_REGION_SIZE];
 
 static inline int
-__init_data_alloc(void *fn, void *data) {
+__init_data_alloc(void *fn, thd_type_t type, capid_t rcv, void *data) {
 	int i, ret, tried = 0;
 
 	assert(fn);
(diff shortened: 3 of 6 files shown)


src/components/include/cos_kernel_api.h
```diff
@@ -107,5 +107,63 @@ int cos_hw_detach(hwcap_t hwc, hwid_t hwid);
 void *cos_hw_map(struct cos_compinfo *ci, hwcap_t hwc, paddr_t pa, unsigned int len);
 int cos_hw_cycles_per_usec(hwcap_t hwc);
 
+/* Default API */
```
gparmer: Might as well document what "default" means.  Otherwise the comment isn't really that useful.

src/components/include/cos_kernel_api.h
```diff
@@ -107,5 +107,63 @@ int cos_hw_detach(hwcap_t hwc, hwid_t hwid);
...
+/* Capabilities for Async activation end point */
+struct cos_aep_info {
+	tcap_t    tc;
+	thdcap_t  thd;
+	arcvcap_t rcv;
+};
```
gparmer: Given my suggestion that we can likely do this without modify existing code, this might grow.

src/components/include/cos_kernel_api.h
```diff
@@ -107,5 +107,63 @@ int cos_hw_detach(hwcap_t hwc, hwid_t hwid);
...
+
+/* Default Component information */
+struct cos_defci {
```
gparmer: I'm not wild about creating the new `ci` namespace, but I don't have any better ideas.

src/components/include/cos_kernel_api.h
```diff
@@ -107,5 +107,63 @@ int cos_hw_detach(hwcap_t hwc, hwid_t hwid);
...
+	struct cos_compinfo ci;
+	struct cos_aep_info sched_aep;
+};
+
+/* current component's default component information */
+extern struct cos_defci curr_defci;
```
gparmer: If this is statically defined in all components that use the `cos_kernel_api`, then we might want to make it in a separate library/`.c` file so that we don't waste the memory for it in components that don't use it.

src/components/include/cos_kernel_api.h
```diff
@@ -107,5 +107,63 @@ int cos_hw_detach(hwcap_t hwc, hwid_t hwid);
...
+
+/*
+ * cos_defci_init: initialize the current component's global cos_defci struct using the standard boot capabilities layout.
+ */
+void cos_defci_init(void);
```
gparmer: I'm not sure we can get rid of `heap_ptr` in the argument list, but I'll withhold judgement to see how you implemented this.

src/components/include/cos_kernel_api.h
```diff
@@ -107,5 +107,63 @@ int cos_hw_detach(hwcap_t hwc, hwid_t hwid);
...
+ */
+void cos_defci_init_ext(tcap_t sched_tc, thdcap_t sched_thd, arcvcap_t sched_rcv, pgtblcap_t pgtbl_cap, captblcap_t captbl_cap, compcap_t comp_cap, vaddr_t heap_ptr, capid_t cap_frontier);
+/*
+ * cos_compinfo_get: returns the cos_compinfo pointer that points to the cos_compinfo struct inside cos_defci.
+ */
+struct cos_compinfo *cos_compinfo_get(void);
+/*
+ * cos_sched_aep_get: returns the AEP pointer that points to the current component's scheduler.
+ */
+struct cos_aep_info *cos_sched_aep_get(void);
```
gparmer: If you're using the `curr_` to describe this component's structures, I'd think you'd want to follow that through.  It might be the case that it would better to have a function to get the current `defci`, and separate functions to get the `aep` and the `compinfo` from *any* `defci`.  These functions feel very specific and it might be better to do something more general here.

src/components/include/cos_kernel_api.h
gparmer: You might actually want to hide this by doing the `extern` in the `.c` file so that it is explicit that this is not part of the ABI.

src/components/include/cos_kernel_api.h
```diff
@@ -107,5 +107,63 @@ int cos_hw_detach(hwcap_t hwc, hwid_t hwid);
...
+
+/*
+ * cos_defci_child_alloc: this is called to create a child component with new initial capabilities like pgtbl, captbl, compcap, etc.
+ *                        if is_sched is set, scheduling end_point is created for the new component,
```
gparmer: grammar

src/components/include/cos_kernel_api.h
```diff
@@ -107,5 +107,63 @@ int cos_hw_detach(hwcap_t hwc, hwid_t hwid);
...
+ * cos_defci_child_alloc: this is called to create a child component with new initial capabilities like pgtbl, captbl, compcap, etc.
+ *                        if is_sched is set, scheduling end_point is created for the new component, 
+ *                        else current component's scheduler is used as scheduler for the child component.
+ */
+int cos_defci_child_alloc(struct cos_defci *child_defci, vaddr_t entry, vaddr_t heap_ptr, capid_t cap_frontier, int is_sched);
+
+/*
+ * thread function that takes it's async rcv end-point as an argument along with user-passed void *
+ */
+typedef void (*cos_aepthd_fn_t)(arcvcap_t, void *);
```
gparmer: Would it be better to pass it a `arcvcap_t`, or a `struct cos_aep_info`?  Do we ever want the other capabilities within the thread?

src/components/include/cos_component.h
```diff
@@ -527,8 +527,10 @@ struct cos_array { char *mem; int sz; }; /* TODO: remove */
 #endif
 
 struct __thd_init_data {
-	void *fn;
-	void *data;
+	int     type;
+	void   *fn;
+	capid_t rcv;
+	void   *data;
```
gparmer: as I said in the email, this needs to go.

src/components/include/cos_thd_init.h
```diff
@@ -20,6 +20,14 @@ __init_data_alloc(void *fn, void *data) {
 
 			assert(__thd_init_data[i].fn == fn);
 			__thd_init_data[i].data = data;
+			__thd_init_data[i].type = type;
+			if (type == THD_AEP) {
```
gparmer: I think you can get rid of the type stuff if you use my solution that doesn't modify existing code.

gparmer:
BTW, in the future, I'm not reviewing code that wasn't written with craftsmanship.  Please keep this in mind when filling out the PR.

src/components/include/cos_kernel_api.h
phanikishoreg: Okay, I'll do that. Thank you.

src/components/include/cos_kernel_api.h
phanikishoreg: Yes

src/components/include/cos_kernel_api.h
phanikishoreg: Yes, I'll make a separate library and remove this extern declaration from the API header.

src/components/include/cos_kernel_api.h
phanikishoreg: I'll be using `cos_get_heap_ptr()` that is used in the `micro_booter` and the `vkernel` `cos_compinfo_init()` for this default initialization.

src/components/include/cos_kernel_api.h
phanikishoreg: Thank you, I'll follow your suggestion.

src/components/include/cos_kernel_api.h
phanikishoreg: I don't see a reason for a non-scheduler AEP to need any other information from that struct. Only scheduler will need the thdcap and tcap, and other AEPs mainly invoke/call scheduler for switching to other threads.

src/components/include/cos_component.h
phanikishoreg: I'm working on this, Thank you.

phanikishoreg:
Because this is mostly API and has no body for the API, I didn't fill the craftsmanship. But yes, I'll do it from the next time no matter. Thank you for the review. I'll close the PR and make a new one soon with API and implementation.

src/components/include/cos_kernel_api.h
gparmer: maybe `defcomp`?

src/components/include/cos_kernel_api.h
gparmer: ah, OK, sounds good.

src/components/include/cos_kernel_api.h
gparmer: OK, sounds reasonable.

src/components/include/cos_kernel_api.h
phanikishoreg: without `info` in it, `defcomp` feels misleading to me given the other namespace is `compinfo`. I'll go with `defcompinfo`, will iterate on name changes if this doesn't sound right.

PR #214, opened 2016-12-29
defcompinfo api and implementation, bugfixes

phanikishoreg:
### Summary of this PR

Implementation for Default API (defcompinfo API), unit test suite for the API, and bugfixes in cos_kernel_api and kernel. 
Removed redundant prio parameter from cos_tcap_alloc / TCAP_ACTIVATE API. 

### Code Quality

Style:

Code Craftsmanship:

### DIFF
--- a/src/components/implementation/tests/unit_defcompinfo/Makefile
+++ b/src/components/implementation/tests/unit_defcompinfo/Makefile
@@ -4,7 +4,7 @@ COMPONENT=unit_defci.o
 INTERFACES=
 DEPENDENCIES=
 IF_LIB=
-ADDITIONAL_LIBS=-lcobj_format -lcos_defcompinfo_api -lcos_kernel_api
+ADDITIONAL_LIBS=-lcobj_format -lcos_defkernel_api -lcos_kernel_api
 
 include ../../Makefile.subsubdir
 MANDITORY_LIB=simple_stklib.o
--- a/src/components/implementation/tests/unit_defcompinfo/Makefile
+++ b/src/components/implementation/tests/unit_defcompinfo/Makefile
@@ -1,0 +1,10 @@
+C_OBJS=unit_defcompinfo.o
+ASM_OBJS=cos_asm_scheduler.o
+COMPONENT=unit_defci.o
+INTERFACES=
+DEPENDENCIES=
+IF_LIB=
+ADDITIONAL_LIBS=-lcobj_format -lcos_defcompinfo_api -lcos_kernel_api
+
+include ../../Makefile.subsubdir
+MANDITORY_LIB=simple_stklib.o
--- a/src/components/implementation/tests/unit_defcompinfo/unit_defcompinfo.c
+++ b/src/components/implementation/tests/unit_defcompinfo/unit_defcompinfo.c
@@ -8,7 +8,7 @@
 #include <string.h>
 #include <cos_component.h>
 #include <cobj_format.h>
-#include <cos_defcompinfo_api.h>
+#include <cos_defkernel_api.h>
 
 #undef assert
 #define assert(node) do { if (unlikely(!(node))) { debug_print("assert error in @ "); *((int *)0) = 0; } } while (0)
@@ -74,16 +74,16 @@ aep_thd_fn(arcvcap_t rcv, void *data)
 static void
 test_aeps(void)
 {
-	int                     i, ret;
-	int                     blocked;
-	cycles_t                cycs;
-	thdid_t                 tid;
-	struct cos_compinfo    *ci    = cos_compinfo_get(cos_defcompinfo_curr_get());
+	int                  i, ret;
+	int                  blocked;
+	cycles_t             cycs;
+	thdid_t              tid;
+	struct cos_compinfo *ci = cos_compinfo_get(cos_defcompinfo_curr_get());
 
 	memset(&test_aep, 0, sizeof(struct cos_aep_info) * TEST_NAEPS);
 
 	printc("Test creating AEPS\n");
-	for (i = 0 ; i < TEST_NAEPS ; i ++) {
+	for (i = 0 ; i < TEST_NAEPS ; i ++ ) {
 		asndcap_t snd;
 
 		printc("\tCreating AEP [%d]\n", i);
@@ -108,7 +108,7 @@ test_childcomps(void)
 	int id, ret;
 
 	printc("Test switching to new components\n");
-	for (id = 0 ; id < CHILD_COMP_COUNT ; id ++) {
+	for (id = 0 ; id < CHILD_COMP_COUNT ; id ++ ) {
 		int      blocked;
 		cycles_t cycs;
 		thdid_t  tid;
@@ -152,11 +152,11 @@ cos_init(void)
 		cos_meminfo_init(&(ci->mi), BOOT_MEM_KM_BASE, COS_MEM_KERN_PA_SZ, BOOT_CAPTBL_SELF_UNTYPED_PT);
 		cos_defcompinfo_init();
 
-		for (id = 0 ; id < CHILD_COMP_COUNT ; id ++) {
+		for (id = 0 ; id < CHILD_COMP_COUNT ; id ++ ) {
 			vaddr_t              vm_range, addr;
 			pgtblcap_t           child_utpt;
-			int                  is_sched       = ((id == CHILD_SCHED_ID) ? 1 : 0);
-			struct cos_compinfo *child_ci       = cos_compinfo_get(&child_defci[id]);
+			int                  is_sched = ((id == CHILD_SCHED_ID) ? 1 : 0);
+			struct cos_compinfo *child_ci = cos_compinfo_get(&child_defci[id]);
 
 			printc("\tCreating new %s component [%d]\n", is_sched ? "scheduler" : "simple", id);
 			child_utpt = cos_pgtbl_alloc(ci);
--- a/src/components/implementation/tests/unit_defcompinfo/unit_defcompinfo.c
+++ b/src/components/implementation/tests/unit_defcompinfo/unit_defcompinfo.c
@@ -23,6 +23,11 @@
 #define CHILD_SCHED_CYCS   TCAP_RES_INF 
 #define CHILD_SCHED_PRIO   TCAP_PRIO_MAX
 
+int is_booter = 1;
+extern vaddr_t cos_upcall_entry;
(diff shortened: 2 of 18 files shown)


src/components/implementation/tests/unit_defcompinfo/Makefile
```diff
@@ -0,0 +1,10 @@
+C_OBJS=unit_defcompinfo.o
+ASM_OBJS=cos_asm_scheduler.o
+COMPONENT=unit_defci.o
+INTERFACES=
+DEPENDENCIES=
+IF_LIB=
+ADDITIONAL_LIBS=-lcobj_format -lcos_defcompinfo_api -lcos_kernel_api
```
gparmer: There is non-consistent naming here.  `cos_kernel_api` and `cos_defcompinfo_api` don't feel like two versions of the same thing...

Perhaps `cos_defkernel_api`?  We need to change the file names accordingly.

src/components/implementation/tests/unit_defcompinfo/unit_defcompinfo.c
```diff
@@ -0,0 +1,238 @@
+/*
+ * Copyright 2016, Phani Gadepalli and Gabriel Parmer, GWU, gparmer@gwu.edu.
+ *
+ * This uses a two clause BSD License.
```
gparmer: Do all of our unit tests use BSD, or do they use GPL?

I know the cos_kernel_api uses BSD, but that's different.

You're using code from the other file, so you have to make sure that the licences don't change without a significant social process.

src/components/implementation/tests/unit_defcompinfo/unit_defcompinfo.c
```diff
@@ -0,0 +1,238 @@
...
+	  ret = vsnprintf(s, len, fmt, arg_ptr);
+	  va_end(arg_ptr);
+	  cos_llprint(s, ret);
+
+	  return ret;
+}
+
+int is_booter = 1;
+extern vaddr_t cos_upcall_entry;
+struct cos_defcompinfo child_defci[CHILD_COMP_COUNT];
```
gparmer: why both `defcompinfo` and `defci` in the same name?

src/components/implementation/tests/unit_defcompinfo/unit_defcompinfo.c
```diff
@@ -0,0 +1,238 @@
...
+}
+
+static void
+test_aeps(void)
+{
+	int                     i, ret;
+	int                     blocked;
+	cycles_t                cycs;
+	thdid_t                 tid;
+	struct cos_compinfo    *ci    = cos_compinfo_get(cos_defcompinfo_curr_get());
```
gparmer: Multiple spaces before `*ci` and after seem superfluous.

src/components/implementation/tests/unit_defcompinfo/unit_defcompinfo.c
```diff
@@ -0,0 +1,238 @@
...
+	printc("Done.\n");
+}
+
+static void
+test_childcomps(void)
+{
+	int id, ret;
+
+	printc("Test switching to new components\n");
+	for (id = 0 ; id < CHILD_COMP_COUNT ; id ++) {
```
gparmer: no space in `id ++`

src/components/implementation/tests/unit_defcompinfo/unit_defcompinfo.c
```diff
@@ -0,0 +1,238 @@
...
+
+	printc("Done.\n");
+}
+
+void
+cos_init(void)
+{
+	int cycs;
+
+	while (!(cycs = cos_hw_cycles_per_usec(BOOT_CAPTBL_SELF_INITHW_BASE))) ;
```
gparmer: Maybe worthwhile changing `cos_hw_cycles_per_usec` to include the loop inside it.

src/components/implementation/tests/unit_defcompinfo/unit_defcompinfo.c
```diff
@@ -0,0 +1,238 @@
...
+		test_aeps();
+
+		/* TEST SWITCHING TO CHILD COMPS */
+		test_childcomps();
+
+		printc("Unit-test done.\n");
+
+		SPIN();
+	} else {
+		struct cos_defcompinfo *defci           = cos_defcompinfo_curr_get();
```
gparmer: don't need the large number of spaces before `=`

src/components/include/cos_defcompinfo_api.h
```diff
@@ -0,0 +1,77 @@
+#ifndef COS_DEFCOMPINFO_API_H
+#define COS_DEFCOMPINFO_API_H
```
gparmer: See comment above about naming this.

src/components/lib/cos_defcompinfo_api.c
```diff
@@ -0,0 +1,143 @@
...
+ * This uses a two clause BSD License.
+ */
+
+#include <cos_defcompinfo_api.h>
+
+static struct cos_defcompinfo curr_defci;
+
+struct cos_defcompinfo *
+cos_defcompinfo_curr_get(void)
+{ return &curr_defci; }
```
gparmer: Is there any way we can add an assertion to check that curr_defci has been initialized?  If so, add a comment above the assertion; something like `/* You must call cos_defcompinfo_init first */`.

src/components/lib/cos_kernel_api.c
```diff
@@ -467,7 +467,7 @@ cos_thd_alloc(struct cos_compinfo *ci, compcap_t comp, cos_thd_fn_t fn, void *da
 
 thdcap_t
 cos_initthd_alloc(struct cos_compinfo *ci, compcap_t comp)
-{ return __cos_thd_alloc(ci, comp, 1); }
+{ return __cos_thd_alloc(ci, comp, 0); }
```
gparmer: What's the story with this?  Did we have a bug?

src/components/implementation/tests/unit_defcompinfo/Makefile
phanikishoreg: Thank you, I'll change them all to `cos_defkernel_api`.

src/components/implementation/tests/unit_defcompinfo/unit_defcompinfo.c
phanikishoreg: It looks like none of the unit tests have a license header in them. I thought that was missing, should have been there for them too. So started the trend with this. 
I also don't have a good idea for what's the right license here.

gparmer: OK, great BSD 2 clause is good.

src/components/implementation/tests/unit_defcompinfo/unit_defcompinfo.c
phanikishoreg: okay, i'll get rid of them. thank you.

src/components/implementation/tests/unit_defcompinfo/unit_defcompinfo.c
phanikishoreg: whoops. will fix it.

src/components/lib/cos_defcompinfo_api.c
phanikishoreg: I think adding comments would be the simplest and right thing to do here. I'll do that, thank you.

src/components/lib/cos_kernel_api.c
phanikishoreg: Yes, we do. upcall entry function expects `arg1` to be `0` for the `initthd`.

src/components/lib/cos_defcompinfo_api.c
gparmer: ....but also the assertion, right?  I was not suggesting to only add a comment.  Please read my statement again.

src/components/lib/cos_kernel_api.c
gparmer: Great, good catch!

src/components/implementation/tests/unit_defcompinfo/unit_defcompinfo.c
phanikishoreg: Okay, I'll do that. I had another idea to call this from upcall entry function, https://github.com/phanikishoreg/composite/commit/3e5c50eeec82c2481c239dbcbd216c1879afd9d0, if that sounds good, I'll incorporate that here.

src/components/implementation/tests/unit_defcompinfo/unit_defcompinfo.c
phanikishoreg: yes, noticed it and fixed in the clean-up commit. thank you.

src/components/implementation/tests/unit_defcompinfo/unit_defcompinfo.c
phanikishoreg: thank you

src/components/lib/cos_defcompinfo_api.c
phanikishoreg: I'm sorry I was wrong. I can neither assert here nor add that comment, because we extract/get the `cos_compinfo` struct from this to call `cos_meminfo_init()`, which is not part of the `cos_defcominfo_init()`.

src/components/implementation/tests/unit_defcompinfo/unit_defcompinfo.c
phanikishoreg: I used `defci` for `defcompinfo` in the variable name all throughout, like `ci` for `compinfo`.

src/components/lib/cos_defcompinfo_api.c
gparmer: damn.  We'd really like the checking so that we can help people use the API correctly.  Perhaps add checks that the `_curr_get` returns an initialized value when it is used within this API?  These checks would likely be added into the child component creation function.  Perhaps elsewhere.

src/components/implementation/tests/unit_defcompinfo/unit_defcompinfo.c
gparmer: Yes.  I know what `defci` stands for.  That's not the question.

phanikishoreg: I'm sorry I didn't get the question.

PR #216, opened 2017-01-05
Review Request: Preemption/Wakeup thread switch and API

phanikishoreg:
### Summary of this PR

Changes for activation of preempted/awoken thread upon cos_rcv from the current (interrupt-handler/AEP) thread & API for waking up a thread.

@ryuxin @georgit - please review. thanks!

### Code Quality

Style:

Code Craftsmanship:

- [x] I agree that there is no "throw-away" code, and that code in this PR is of high quality: **IMP NOTE: Untested. Working on Unit-testing now**


src/kernel/include/thd.h
```diff
@@ -39,6 +39,19 @@ struct rcvcap_info {
...
 };
 
+/*
+ * This is the data structure embedded in the cos_cpu_local_info that
+ * contains information for the thread that was either preempted or woken up
+ * and is waiting to be scheduled instead of the current thread's scheduler
+ * upon RCV syscall. This is mainly to reduce the number of context switches
+ * to schedule the thread that is deemed eligible by the scheduler. 
+ */
+struct next_thdinfo {
```
gparmer: If it is in the `cos_cpu_local_info`, why is it also here?

src/kernel/include/thd.h
```diff
@@ -39,6 +39,19 @@ struct rcvcap_info {
...
+	struct tcap   *tc;
+	struct thread *thd;
+	tcap_prio_t    prio;
+} per_core_nextti[NUM_CPU_COS];
```
gparmer: Ahh, OK, I see.  You're using a pointer in the `cos_cpu_local_info` to this.

Why not just include a field `struct next_thdinfo thd_next;` (notice, it isn't a pointer).

src/kernel/include/thd.h
```diff
@@ -153,6 +166,17 @@ thd_rcvcap_sched(struct thread *t)
 }
 
 static void
+thd_next_thdinfo_update(struct cos_cpu_local_info *cli, struct thread *thd, 
+			struct tcap *tc, tcap_prio_t prio)
+{
+	struct next_thdinfo *nti = (struct next_thdinfo *)cli->next_ti;
```
gparmer: You want this inlined into the per-core structure to avoid the additional TLB and data-cache-line misses for the pointer here.

src/kernel/tcap.c
```diff
@@ -215,6 +215,32 @@ tcap_merge(struct tcap *dst, struct tcap *rm)
...
+	int ret;
+	struct next_thdinfo *nti = (struct next_thdinfo *)(cli->next_ti);
+	tcap_prio_t tmpprio      = tcap_sched_info(tc)->prio;
+
+	if (!nti->tc) {
+		assert(!nti->thd);
+		goto fixup;
+	}
+
+	if (tc == nti->tc && prio >= nti->prio) {
```
gparmer: can avoid `{}`

gparmer:
This looks great.  Simple.

Just some nit-picks.  Keep hammering away at the tests, then see if you can get to `asnd`.

src/kernel/include/thd.h
phanikishoreg: Oh yes, I didn't think in that perspective, I'm sorry. I was thinking in terms of size of `cos_cpu_local_info`. I will change it as you suggested. Thank you.

src/kernel/tcap.c
phanikishoreg: Yes, I'll. Thank you.

phanikishoreg:
Thank you. Code for `asnd` is already in the PR (`asnd_process`), will start testing all the scenarios, yes.

src/kernel/include/thd.h
gparmer: Indeed that is worth worrying about.  However, I think that max 24 bytes (on a 64 bit machine) is fine.  We don't use much stack in the kernel.

gparmer:
Another comment I just remembered:

You should add code into the thread and tcap deactivation paths to invalidate the struct you provided here in the case that the thread/tcap being deleted is the one pointed to there.

gparmer:
Somehow I missed the `asnd` stuff.  Yes, also quite simple!

phanikishoreg:
Thank you, missed that too. I'll add invalidations in thd/tcap deactivation paths.

src/kernel/include/thd.h
phanikishoreg: Got it!

src/components/lib/cos_kernel_api.c
```diff
@@ -649,6 +649,10 @@ int
 cos_thd_switch(thdcap_t c)
 { return call_cap_op(c, 0, 0, 0, 0, 0); }
 
+int
+cos_thd_wakeup(thdcap_t thd, tcap_t tc, tcap_prio_t prio)
+{ return call_cap_op(tc, CAPTBL_OP_TCAP_WAKEUP, thd, (prio << 32) >> 32, prio >> 32, 0); }
```
interwq: nit: use mask to clear the higher 32 bits. It should be a pretty common operation so a macro is probably necessary.

phanikishoreg: Thanks Qi. 
shift operations are pretty much free, and also to maintain consistency, used them to clear the higher 32bits, which is pretty much how it's done throughout `Composite`.

gparmer: @interwq's right about the macro though.  This is becoming quite common in the code, so a macro would make it quite a bit more readable.

interwq: Well, we are talking about 2 shifts (assuming compiler doesn't do something tricky) versus 1 mask instruction (I believe both are close to free instructions in this case). Won't be a big difference but less instructions isn't bad ;)

phanikishoreg: Kk. Will try to do it with this PR, or will add an Issue to clean-up throughout and do it after my vacation.

gparmer:
@georgit Does this work to send you the notification email?

src/components/lib/cos_kernel_api.c
gparmer: BTW, I doubt the shift vs. mask thing is going to generate any different code (assuming unsigned values).  The question is what is easier to read.  If we macro-ify this, then it doesn't really matter, so I don't have a strong opinion either way.

I had figured that the shifts were easier to read before, but with a macro, I don't think it matters.

src/kernel/capinv.c
```diff
@@ -811,6 +821,15 @@ cap_arcv_op(struct cap_arcv *arcv, struct thread *thd, struct pt_regs *regs,
 	}
 
 	next = notify_parent(thd);
+	/* if preempted/awoken thread is waiting, switch to that */
```
georgit: I like the use of comments for large chunks of code.

src/kernel/capinv.c
```diff
@@ -1531,6 +1550,19 @@ composite_syscall_slowpath(struct pt_regs *regs, int *thd_switch)
 
 			break;
 		}
+		case CAPTBL_OP_TCAP_WAKEUP:
+		{
+			struct cap_tcap *tcapwkup = (struct cap_tcap *)ch;
```
georgit: I don't really care for this PR itself, but what does the style guide say for underscores here? I would have expected tcap_wkup.

phanikishoreg: I think the Style guide may not have a strict `grammar` for variable names, except that the variable names should be meaningful and perhaps not too long. I could have also had `tcw`, that would have been less descriptive, but still okay.

gparmer: It doesn't say much about variables, aside from the convention to not use verbs first (which was broken earlier, but for a good reason, I believe).  We aren't going full hungarian here, just encouraging good use of variable names.  I don't have a strong preference personally for `_` or not here.

phanikishoreg:
Thank you all for the review. 
I'll close the PR and make a fresh one incorporating the comments, bug fixes and unit-tests.

PR #218, opened 2017-01-06
Preempted/Awoken thread activation, thd_wakeup API, bug fix, unit test

phanikishoreg:
### Summary of this PR

**Background: TCaps provide mechanism to switch to Receive endpoints and activate schedulers when the budget/timeout expires. Receive endpoints, when they make a system call to receive notifications, kernel mechanism dictates that pending notifications be returned but if there are no pending notifications, mechanism is to switch to the scheduler of that receive endpoints. Now especially in the case of interrupts, where Interrupt threads usually wakeup other threads to process the data, (as an example - network interrupt threads wakeup next network-layers, till the data is delivered to the application), when the interrupt completes its execution, activating its scheduler might be unnecessary if the scheduler decides to schedule the awoken thread. What this means is that, the number of context switches required to schedule the awoken thread is at least 2, which could be avoided with this optimization where the Interrupt thread sets the next thread to be scheduled and kernel mechanism for TCaps can switch to them directly instead of the scheduler switch. This is the idea behind the `cos_thd_wakeup()` API.**

**Secondly, TCaps mechanism to preempt the current thread running, involves comparing the priorities of current and the Interrupt thread. If it deems that the Interrupt thread is of higher priority, it activates it immediately. Going with the above background, if the Interrupt thread completes its execution, it would eventually trigger a awoken thread to be scheduled by it's scheduler which could have been of lower priority than the thread that was preempted by this Interrupt thread. Causing starvation on the medium priority thread. This starvation, as in the case above could vary depending on the number of scheduling susbsystems involved, etc, which makes the system fail to meet predictability guarantees/deadline guarantees on that preempted thread. In order to avoid this, the optimization with `cos_thd_wakeup` or with `cos_asnd` is to allow for preempted thread to be saved on kernel stack, and a priority check to be performed on the thread that is awoken by one of these APIs with the preempted. This guarantees that the higher priority thread at that instant will be eligible when the current highest priority interrupt thread completes execution.** 

**To this end, and for simplicity, we maintain a single preempted/awoken thread in the per-core kernel stack. This if not a complete solution, is a step in the right direction to address the TCap priorities for the threads in the system for their deadline guarantees.**

This PR incorporates review comments from PR: https://github.com/gparmer/composite/pull/216, bug fix for `cos_thd_wakeup`, unit-test cases for both `cos_thd_wakeup` API and preemption using `cos_asnd` which is almost the same path as used in H/w Interrupt thread activation from the kernel.
**Update: added `cos_thd_wakeup` with budget parameter**

(I wonder if the unit-tests should be kept commented for the standard test regime because it is hard to tell if the test was successful without sufficient background about the tests (I've included comments in the test code), but again this is probably the case with `test_thds` and `test_async_endpoints`.)

@gparmer @ryuxin @georgit - please review. Thanks!
I hope this background helps in the review - @ryuxin @georgit. Though all of this is mainly a practice review procedure for all of us. 

### Code Quality

Style:

Code Craftsmanship:


src/components/implementation/tests/micro_booter/mb_tests.c
```diff
@@ -487,6 +491,197 @@ test_budgets(void)
...
+
+	/*
+	 * test case 3: intr_thd = H, preempted_thd = M, worker_thd = L
+	 * expected result:
+	 * - cos_thd_wakeup should not invalidate preempted_thd.
+	 * - cos_rcv from intr_thd should activate preempted_thd.
+	 */
+	PRINTC(" Test - Intr = H, Preempted = M, Worker = L\n");
+	test_wakeup_case(&wat, TEST_PRIO_MED, TEST_PRIO_HIGH, TEST_PRIO_LOW);
+
```
gparmer: Am I missing the logic that tests that the proper sequence of activations/switches actually happens?

src/components/implementation/tests/micro_booter/mb_tests.c
```diff
@@ -487,6 +491,197 @@ test_budgets(void)
...
+	/*
+	 * test case 2: Sender = L, Receiver: H 
+	 * - scheduler here is initthd (this thd)
+	 * expected result:
+	 * - cos_asnd from sender should trigger receiver activation and add sender as wakeup thread.
+	 * - cos_rcv from receiver should activate sender thread.
+	 */
+	PRINTC(" Test - Sender = L, Receiver = H\n");
+	test_preemption_case(&pat, TEST_PRIO_LOW, TEST_PRIO_HIGH);
```
gparmer: You did a great job abstracting away the shared parts of the code, but I'm missing the code that checks that the proper control flow for each case actually happens.

src/kernel/tcap.c
```diff
@@ -215,6 +218,30 @@ tcap_merge(struct tcap *dst, struct tcap *rm)
 	return 0;
 }
 
+int
+tcap_wakeup(struct tcap *tc, tcap_prio_t prio, struct thread *thd, struct cos_cpu_local_info *cli)
+{
```
gparmer: I think that we also need the `budget`.  As the code stands, we're setting the next timer interrupt to fire corresponding to the budget of the associated tcap.  But we want the scheduler to be able to instead control when the next timer interrupt fires, so long as it is less than the tcap's budget.

This is, in the current design, the difference between a tcap associated with a rcv end-point, and one that is used for the scheduling decisions.  This design allows the scheduler to not have to constantly `tcap_transfer`, and spend time managing tcaps.  So, we'd like to continue to support that, which is what wakeup is all about.

What do you think?

src/components/implementation/tests/micro_booter/mb_tests.c
phanikishoreg: I'm verifying it manually through the logs printed, which is why I was wondering if this should be part of the standard test regime. Each case has it's own sequence, so verifying right sequence might get complicated, I'll think about it. Probably maintaining a flag in exec_cluster to indicate if the thread activated is indeed expected.

src/components/implementation/tests/micro_booter/mb_tests.c
phanikishoreg: I'll think about automating verification, at this point it's manually done through the logs printed.

src/kernel/tcap.c
phanikishoreg: Yes, It makes sense. `cos_thd_wakeup` as it is now, doesn't control the next timer interrupt, which is one main difference from `cos_switch` and can be a limitation as you suggested. 
Implementing it could be simple, bookkeeping the budget in `next_thdinfo` and converting it to timeout(`tcap_time_t`) when `awoken thread` it is activated from `cos_rcv`.

phanikishoreg: Updated PR. I'll try and add a test case for the budget test in `cos_thd_wakeup()`.

gparmer:
OK, I've been thinking about your API for tracking the sequence of operations.  It is functionally fine, but is getting into the territory of opaqueness.  Remember that these micro/unit tests are going to be one of the best forms of documentation for how to use the system, so we want this all to be relatively clear.  My main issue them them now is that they are not self-documenting.

I'd suggest adding functions similar to `seq_expected_order(struct exec_cluster, int)` to set the expected order of the thread, and change `seq_check` to `seq_check_order`.  We really want to make sure that people understand that the "random" integers are not just integers, rather the ordering of threads executing in sequence.

phanikishoreg:
Thank you. I've updated it with the API and budget test for wakeup.

PR #230, opened 2017-02-17
Voter

georgit:
### Summary of this PR

I'm actually going to submit this PR now even if no one looks at it until Sat/Sun/Mon.

[It appears this PR thinks it also has all the data from my `quarantine`/`cbuf_mgr` changes. Please ignore the bulk of this. You only want to be looking at `fork_tests/ping`, `f_t/pong`, and `f_t/voter`.]

Here are some of the key questions I have:

- Do my structs follow what we talked about on Tue? In particular:
- Should an nmod component be able to send to more than one nmod component at a time? I think the answer we agreed to during our last meeting was yes, but so far we can only send one set of data to multiple receivers.
- Does a replica need a flag for whether or not it has already sent in data? Could a replica do `write(A); write(B);` and have data B override data A? We agreed that during a write, a replica should be blocked so it doesn't spam the system, but does this cover all cases?
- What exactly does the thread that starts out in the voter _do_? Currently, I believe it exits out after cos_init, and then it is the threads from ping and pong that pass data along as they call the read/write functions. Is that safe? Would there be any value in periodically waking up the voter so it can assess the situation? My intuition is no but wanted to check.
- Currently I am ignoring cbufs and just passing around an int. This is easy to fix - I just wanted to use the simplest structure for data I could think of to start.

@jcmarsh @gparmer

### Code Quality

Style:

Code Craftsmanship:


src/components/implementation/fork_tests/voter/voter.c
```diff
@@ -0,0 +1,290 @@
...
+
+struct channel {
+	struct nmod_comp *A, *B;	// come up with better terms, they're symmetric
+	int timer; // is this an int, figure it out later
+	int A2B_data, have_A2B_data;	// make these arrays
+	int B2A_data, have_B2A_data;
+};
+
+/* number of components and component array */
+int n_comps = 2;
```
gparmer: should be a `#define`?

src/components/implementation/fork_tests/voter/voter.c
```diff
@@ -0,0 +1,290 @@
...
+};
+
+struct nmod_comp {
+	struct replica_info *replicas;
+	int nreplicas;
+};
+
+struct channel {
+	struct nmod_comp *A, *B;	// come up with better terms, they're symmetric
+	int timer; // is this an int, figure it out later
```
gparmer: line up comments....use proper comments

src/components/implementation/fork_tests/voter/voter.c
```diff
@@ -0,0 +1,290 @@
...
+	int A2B_data, have_A2B_data;	// make these arrays
+	int B2A_data, have_B2A_data;
+};
+
+/* number of components and component array */
+int n_comps = 2;
+struct nmod_comp components[2];
+
+/* number of channels and channels array. Symmetric. */
+int n_channels = 1;
```
gparmer: same

src/components/implementation/fork_tests/voter/voter.c
```diff
@@ -0,0 +1,290 @@
...
+struct channel channels[1];
```
gparmer: if they are `#defines`, then you can use them in the array index here....

src/components/implementation/fork_tests/voter/voter.c
```diff
@@ -0,0 +1,290 @@
...
+
+int period = 100;
+
+/* map of spd to replica type */
+replica_type map[250];
```
gparmer: #define

src/components/implementation/fork_tests/voter/voter.c
```diff
@@ -0,0 +1,290 @@
...
+	}
+}
+
+void cos_init(void)
+{
+	int i, j;
+	for (i = 0; i < n_comps; i++) {
+		printc("Setting up nmod component %d\n", i);
+		components[i].nreplicas = 1; // for now, no fork
+		components[i].replicas = (struct replica_info*) malloc(sizeof(struct replica_info) * components[i].nreplicas);
```
gparmer: likely want to do static allocation here, not malloc.  Keep the voter simple.  dynamic allocation is moving in the opposite direction.

src/components/implementation/fork_tests/voter/voter.c
```diff
@@ -0,0 +1,290 @@
...
+		for (j = 0; j < components[i].nreplicas; j++) {
+			components[i].replicas[j].spdid = 0;
+			components[i].replicas[j].blocked = 0;
+		}
+	}
+
+	channels[0].A = &components[0];
+	channels[0].B = &components[1];
+
+	for (i = 0; i < 250; i++) map[i] = none;
```
gparmer: `none` is non-standard, and I can't see where this is defined.  This is making it quite unreadable.

src/components/implementation/fork_tests/voter/voter.c
```diff
@@ -0,0 +1,290 @@
...
+		printc("Examing channel %d between nmodcomps with replicas %d and %d \n", i, channels[i].A->replicas[0].spdid, channels[i].B->replicas[0].spdid);
+
+		printc("channel A2B_data present %d (%d)\n", channels[i].have_A2B_data, channels[i].A2B_data);
+		printc("channel B2A_data present %d (%d)\n", channels[i].have_B2A_data, channels[i].B2A_data);
+
+		if (channels[i].have_A2B_data) {
+			rcv_comp = channels[i].B;
+			found = 1;
+		}
+		else if (channels[i].have_B2A_data) {
```
gparmer: possible for both to be true here????  I'm not sure.

src/components/implementation/fork_tests/voter/voter.c
```diff
@@ -0,0 +1,290 @@
...
+			rcv_comp = channels[i].A;
+			found = 1;
+		}
+
+		if (found) {
```
gparmer: change to `if (!found) continue;` to avoid the nesting.  It is making it a little more imposing then it needs to be.

src/components/implementation/fork_tests/voter/voter.c
```diff
@@ -0,0 +1,290 @@
...
+struct channel channels[1];
+
+int period = 100;
+
+/* map of spd to replica type */
+replica_type map[250];
+
+void cos_fix_spdid_metadata(spdid_t o_spd, spdid_t f_spd) { }
+
+struct nmod_comp *get_component(spdid_t spdid) {
```
gparmer: use another map for this?  O(n*m) isn't great.  Also will simpilify.

src/components/implementation/fork_tests/voter/voter.c
```diff
@@ -0,0 +1,290 @@
...
+		}
+		return NULL;
+	}
+	else if (type == pong) {
+		for (i = 0; i < components[1].nreplicas; i++) {
+			if (components[1].replicas[i].spdid == spdid) return &components[1];
+		}
+		return NULL;
+	}
+	else {
```
gparmer: `else` not needed.

src/components/implementation/fork_tests/voter/voter.c
```diff
@@ -0,0 +1,290 @@
...
+	struct nmod_comp *component = NULL;
+	int i;
+	replica_type type = map[spdid];
+	if (type == none) return NULL;
+
+	component = get_component(spdid);
+	if (!component) return NULL;
+
+	for (i = 0; i < component->nreplicas; i++) {
+		if (component->replicas[i].spdid == spdid) return &component->replicas[i];
```
gparmer: again, replace all of this with a map from spdid to replica (and perhaps have poi8nters in the replicas to the components etc...)

src/components/implementation/fork_tests/voter/voter.c
```diff
@@ -0,0 +1,290 @@
...
+}
+
+int block_replica(struct replica_info *replica) {
+	if (!replica->thread_id) BUG();
+	printc("Blocking thread %d\n", replica->thread_id);
+	replica->blocked = 1;
+	return sched_block(cos_spd_id(), 0);
+}
+
+int nwrite(spdid_t spdid, replica_type to, int data) {
```
gparmer: we might want the second argument to nwrite be the channel id.  Then we should have O(1) (with the above maps).

Will of course have to validate that the calling component is allowed to use that channel for communication (i.e. that it is A or B).

src/components/implementation/fork_tests/voter/voter.c
```diff
@@ -0,0 +1,290 @@
...
+			ret = block_replica(replica);
+			if (ret < 0) printc("Error\n");
+			/* On wakeup */	
+			return 0;
+		}
+	}
+	else if (component == c->B) {
+		if (c->have_B2A_data) {
+			printc("This replica has already sent some data down the channel\n");
+			return -1; // whatever, shouldn't happen if blocked
```
gparmer: I think I'm not understanding these cases.

src/components/implementation/fork_tests/voter/voter.c
```diff
@@ -0,0 +1,290 @@
...
+		else {
+			c->have_A2B_data = 1;
+			c->A2B_data = data;
+			ret = block_replica(replica);
+			if (ret < 0) printc("Error\n");
+			/* On wakeup */	
+			return 0;
+		}
+	}
+	else if (component == c->B) {
```
gparmer: we can chat about how to remove most of these nested conditionals

src/components/implementation/fork_tests/voter/voter.c
```diff
@@ -0,0 +1,290 @@
...
+int nread(spdid_t spdid, replica_type from, int data) {
+	struct replica_info *replica = get_replica(spdid);
+	struct nmod_comp *component = get_component(spdid);
+	struct nmod_comp *from_comp = (from == ping) ? &components[0] : (from == pong) ? & components[1] : NULL;
+	struct channel *c = get_channel(component, from_comp);
+	int ret;
+	if (!replica) BUG();
+	if (!c) BUG();
+
+	if (component == c->A) {
```
gparmer: Yuck.  Lets figure out a way to remove all of this horrible rudundancy!!!

src/components/implementation/fork_tests/voter/voter.c
georgit: Wait, how are you also viewing the channel commits? I didn't include those in the PR??? Must be a github feature. But this is fantastic, you basically seem to have done this PR and the next one at once.

gedare: @georgit when you push commits to a branch that is part of a pull request, the PR is automatically updated to include those commits. (Corrolary, use separate feature branches for separate PRs. :))

src/components/implementation/fork_tests/voter/voter.c
gedare: That comment is so meta considering the nmr theme. ;)

src/components/implementation/fork_tests/voter/voter.c
georgit: Yes! I was wondering if we wanted nmod_comp id's. Had not considered channel id's.

src/components/implementation/fork_tests/voter/voter.c
georgit: Well, parts of this code are horribly redundant. I managed to take out one redundant piece of code before pushing for the week. Some remain.

src/components/implementation/fork_tests/voter/voter.c
georgit: For the record, it is defined in interfaces/voter/voter.h but it is a terrible choice of name. I want to change this whole thing because interfaces shouldn't have information about ping/pong/none. This should likely be in some separate config file.

src/components/implementation/fork_tests/voter/voter.c
georgit: Yes it should. I kind of did this, then realized why it was wrong, but didn't fix it, and now it happens in three places.

src/components/implementation/fork_tests/voter/voter.c
```diff
@@ -89,17 +92,20 @@ int nwrite(spdid_t spdid, replica_type to, size_t sz) {
...
-		assert(replica->buf_write);
 		ret = block_replica(replica);
 		if (ret < 0) printc("Error\n");
 		/* On wakeup */	
 		return 0;
 	}
 }
 
-int nread(spdid_t spdid, replica_type from, size_t sz) {
+size_t nread(spdid_t spdid, replica_type from, size_t sz) {
```
jcmarsh: What is size_t sz intended for? Would a comment block describing parameters / return value / side effects be appropriate here?

src/components/interface/voter/voter.h
```diff
@@ -8,7 +8,7 @@ typedef enum {
 } replica_type;
 
 // to be fleshed out at a later time - long term solution is to have data in cbufs
-int nread(spdid_t spdid, replica_type from, size_t sz);
+size_t nread(spdid_t spdid, replica_type from, size_t sz);
 int nwrite(spdid_t spdid, replica_type to, size_t sz);
```
jcmarsh: Is `nwrite` finished? If so, I think I am confused...

georgit: I think the confusion might be that nwrite no longer takes in data. That's because after I changed data from an int to an actual buffer, I had it mapped into both the voter and ping/pong's address space. One write buffer for ping and one for pong. So ping/pong now just write into their buffer and only tell the voter how many bytes they have written. If the confusion is for the int return value, that's just success/fail for me. Does that help?

src/components/implementation/fork_tests/voter/voter.c
georgit: A comment would be helpful and it's super hypocritical of me to not include one - you're right! Although I will put it in voter.h.

size_t sz is intended to tell the voter the maximum bytes the calling component _thinks_ it can read. Then the size_t return value is how many bytes were actually returned. I think this is inspired by POSIX? I've seen it somewhere before, and I believe Gabe suggested this unless I misunderstood.

jcmarsh: Okay, I think we are on the same page for it. The voter will not read more that sz bytes. As of right now, sz isn't being used though.

src/components/interface/voter/voter.h
jcmarsh: Yes, that helps! Thanks.

src/components/implementation/fork_tests/voter/voter.c
gparmer: `replica_type` is not correct for two reasons.

First, if you're going to provide a new type (i.e. with `typedef`), it should be `replica_t`.

Second, we are supposed to be reading and writing to channels, not specific replicas.  Why would the component being replicated even know it is replicated?  Ideally, we want to hide that from it.  There should be no hint about replicas anywhere in the API.

We read and write to *channels*, so that variables should reflect that.

src/components/implementation/fork_tests/voter/voter.c
```diff
@@ -176,7 +182,7 @@ void monitor(void) {
 			matches[j]++;						// for our own match. Is this how voting works? Do we even know anymore...
 			for (k = j + 1; k < snd_comp->nreplicas; k++) {
 				/* compare j to k  */
-				if (memcmp(snd_comp->replicas[j].buf_write, snd_comp->replicas[k].buf_write, 1) == 0) {
+				if (snd_comp->replicas[j].sz_write == snd_comp->replicas[k].sz_write && memcmp(snd_comp->replicas[j].buf_write, snd_comp->replicas[k].buf_write, snd_comp->replicas[j].sz_write) == 0) {
 					matches[j]++;
```
gparmer: is `matches` global data?  If so, that is broken.

Shouldn't `matches` just be in the `nmod_comp` structure, not wherever this is?

Shouldn't all of this code be gone (I'm assuming it is in the periodic thread which may be incorrect).

gparmer: The iteration of `j` through 9 doesn't give me much confidence.

georgit:
Pushing some stuff. Mainly I'm concerned my solution, even if it worked, is inelegant. I can't help but feel like tracking all of these fields is not the way to go about the problem.

PR #231, opened 2017-03-06
Bugfixes in schedlib, timer reprogramming fix, and ps submodule url to https

phanikishoreg:
### Summary of this PR

1. Schedlib: Scheduler library is a new user-level scheduling abstraction. It provides standard API for users and backend API for new scheduling policy developers. A fixed-priority module has been implemented, which has periodic timer interval of 100us. Interval calculation logic is broken in that it doesn't actually take the offset for calculation. And also the `cos_defswitch` api has been used incorrectly by passing in the absolute timer value instead of `tcap_time_t` value. 
Moreover, a unit-test suite is available for testing the scheduling library and fp scheduling module. The test has been improved for making threads work (do computation) and then yield. Immediately yielding causes some issues with Interval calculations as most of the time is spent in context-switches and in the kernel (non-preemptive mode). 

2. Timer reprogramming: LAPIC timers are used for Oneshot/TSC-Deadline for the TCap mechanism. In the past, we used HPET timers which were really pathetic especially for use with Oneshots, as it took forever (~10us) to program them. While debugging some of the issues in it especially the inaccuracies in the timer conversions, I had to remove a check that would avoid timer reprogramming if the new timer being set is close to already programmed one. But now, given LAPIC timer programming is about 400cycles, and is way better in accuracies, we are adding those checks back to ensure that timer is not reprogrammed if it's close to a previously programmed one. 

3. Minor: Update for `ps` submodule `url` which is currently `git:`/`ssh` based, that requires users to have their public-keys registered with github, to `https:` that doesn't have any such requirements. 

### Code Quality

Style:

Code Craftsmanship:

### DIFF
--- a/src/components/implementation/no_interface/scheddev/unit_schedlib.c
+++ b/src/components/implementation/no_interface/scheddev/unit_schedlib.c
@@ -17,7 +17,7 @@
 #define PRINT_FN prints
 #define debug_print(str) (PRINT_FN(str __FILE__ ":" STR(__LINE__) ".\n"))
 #define BUG() do { debug_print("BUG @ "); *((int *)0) = 0; } while (0);
-#define SPIN() do { while (1) ; } while (0)
+#define SPIN(iters) do { if (iters > 0) { for (; iters > 0 ; iters -- ) ; } else { while (1) ; } } while (0)
 
 static void
 cos_llprint(char *s, int len)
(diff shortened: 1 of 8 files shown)


src/components/implementation/no_interface/scheddev/unit_schedlib.c
```diff
@@ -49,11 +49,15 @@ printc(char *fmt, ...)
...
 
 #define N_TESTTHDS 8
+#define WORKITERS  10000
 
 void
 test_thd_fn(void *data)
 {
+	int workiters = WORKITERS * ((int)data);
 	while (1) {
+		SPIN(workiters);
```
gparmer: With this, how are you testing if the timeouts are happening correctly?

phanikishoreg: I'm mainly checking the time between 2 timer interrupts. 
They are consistent and about the 100us or 1000us periods I tested in the schedlib.

PR #232, opened 2017-03-09
Bugfixes in timer-driver and schedlib sched_loop 

phanikishoreg:
### Summary of this PR

HPET timer configuration wasn't done right. Instead of using AND, we have had XOR for a while there. I'm still not sure how it worked fine for Oneshots (I need to think about how many times we XOR in that to see that it really worked because of that) but because we use it for initial Periodic timers, it was okay. Now, I had to fix bugs to make it work for reprogramming of periodic timers. 

LAPIC timer divide configuration was not set to right numbers, which is a bug but didn't impact much as we calibrate oneshots dynamically and use. So whichever divider it used, we were still programming oneshots using conversions required for such frequency!

Schedlib - sched_loop should consider child schedulers where there are events other than scheduling ones, mainly activations from the parent schedulers if they decide to use notification-based activations as opposed to simply context-switching to them. Also, if pending = 0, sched loop still need to process that one event that caused it to unblock, so fixed that as well. 

### Code Quality

Style:

Code Craftsmanship:

### DIFF
--- a/src/components/implementation/no_interface/scheddev/sl.c
+++ b/src/components/implementation/no_interface/scheddev/sl.c
@@ -234,9 +234,10 @@ sl_sched_loop(void)
 		sl_cs_enter();
 
 		do {
-			thdid_t  tid;
-			int      blocked;
-			cycles_t cycles;
+			thdid_t        tid;
+			int            blocked;
+			cycles_t       cycles;
+			struct sl_thd *t;
 
 			/*
 			 * a child scheduler may receive both scheduling notifications (block/unblock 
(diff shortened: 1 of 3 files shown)


src/components/implementation/no_interface/scheddev/sl.c
```diff
@@ -228,26 +228,33 @@ sl_init(void)
...
+			int      blocked;
+			cycles_t cycles;
+
+			/*
+			 * a child scheduler may receive both scheduling notifications (block/unblock 
+			 * states of it's child threads) and normal notifications (mainly activations from
+			 * it's parent scheduler).
+			 */
+			pending = cos_sched_rcv(BOOT_CAPTBL_SELF_INITRCV_BASE, &tid, &blocked, &cycles);
+			if (tid) {
```
gparmer: `if (!tid) continue;`

gparmer: small fixup to make the common path clear

phanikishoreg: Got it, Will fix it. Thank you!

phanikishoreg:
Fixed the code and PR auto-updated.

PR #240, opened 2017-05-31
Change the src Makefile to use PLATFORM_ID instead of PLATFORM

Others:
### Summary of this PR
The src Makefile creates a temporary file that stores the name of
the platform the build is running on. (It does this when you run
`make config`.) This file was historically called PLATFORM, which
worked fine. However, there is a folder in the src directory
called `platform`. On case insensitive file systems this causes an
obvious problem. To fix this, I modified the Makefile to use
PLATFORM_ID instead.

### Code Quality

Style:

Code Craftsmanship:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ tags
 src/kernel/include/shared/cpu_ghz.h
 src/Makefile.cosconfig
 src/PLATFORM
+src/PLATFORM_ID
 modules.order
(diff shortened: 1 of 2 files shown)


gparmer:
Merging this now, and we can change the name later if we need.  Not that big a deal either way.  Perhaps just making it a dot file.

PR #241, opened 2017-06-01
Bugfixes and changes in kernel and schedlib

phanikishoreg:
### Summary of this PR

This PR includes bugfixes in either the virtualization test or the hierarchical test debugging. The description of the commits could be used as a change/bugfix list. Most of the fixes are in the kernel in this PR, there are a few more that I'll continue to work on and push.

Remaining (major ones):
1. scheduler thread activation on timeouts.
2. hpet repurposing and hpet first period for scheduler period alignment
3. 64bit introspect especially to get hpet first period.
4. tcap uid only for schedulers and non schedulers use their parents for scheduler offsets. This is not part of my fixes for paper. 

5. EDF module & budget server implementation for AEPs
6. Additional api in defkernel_api lib, like cos_deftransfer, cos_defdelegate, etc.
7. Different sl_thread types
8. Wakeup logic in timeout module for threads that need wakeup.
9. hierarchical scheduling
10. scheduler period alignment
11. priority setting with tcap/scheduler module integration.
12. (spinner library)
13. support for updating priorities (especially) for dynamic priority-based schedulers.

### Code Quality

Style:

Code Craftsmanship:


gparmer:
@Others Do a quick code review of this?

src/components/implementation/no_interface/scheddev/sl.c
```diff
@@ -17,21 +17,24 @@ struct sl_global sl_global_data;
  * These functions are removed from the inlined fast-paths of the
  * critical section (cs) code to save on code size/locality
  */
-void
+int
```
Others: Maybe add some documentation explaining exactly what this return value means

src/components/implementation/no_interface/scheddev/sl.c
```diff
@@ -17,21 +17,24 @@ struct sl_global sl_global_data;
...
 	/* recursive locks are not allowed */
 	assert(csi->s.owner != t->thdcap);
 	if (!csi->s.contention) {
 		csi->s.contention = 1;
-		if (!ps_cas(&g->lock.u.v, cached->v, csi->v)) return;
+		if (!ps_cas(&g->lock.u.v, cached->v, csi->v)) return 1;
 	}
 	/* Switch to the owner of the critical section, with inheritance using our tcap/priority */
-	cos_defswitch(csi->s.owner, t->prio, g->timeout_next, tok);
+	if ((ret = cos_defswitch(csi->s.owner, t->prio, g->timeout_next, tok))) return ret;
```
Others: Looks like unnecessary parenthesis here. Also, an equals sign in an if statement looks like code smell to me... Can that be avoided here?

src/components/implementation/no_interface/scheddev/sl.c
```diff
@@ -248,18 +249,24 @@ sl_sched_loop(void)
...
 			if (blocked) sl_mod_block(sl_mod_thd_policy_get(t));
 			else         sl_mod_wakeup(sl_mod_thd_policy_get(t));
+
+			sl_cs_exit();
 		} while (pending);
 
+		if ((ret = sl_cs_enter_sched()) == -EBUSY) continue;
 		/* If switch returns an inconsistency, we retry anyway */
 		sl_cs_exit_schedule_nospin();
 	}
```
Others: The usage of the `ret` variable in this function makes my head spin. Combined with the use of equals in if statements, it's really confusing. Is there any refactoring that could be done to make it more clear what's going on? The variable could at least be renamed...

src/components/implementation/no_interface/scheddev/sl.h
```diff
@@ -85,32 +85,48 @@ sl_cs_owner(void)
 { return sl__globals()->lock.u.s.owner == sl_thd_curr()->thdcap; }
 
 /* ...not part of the public API */
-void sl_cs_enter_contention(union sl_cs_intern *csi, union sl_cs_intern *cached, thdcap_t curr, sched_tok_t tok);
+int sl_cs_enter_contention(union sl_cs_intern *csi, union sl_cs_intern *cached, thdcap_t curr, sched_tok_t tok);
```
Others: Shouldn't the return type be on its own line?

src/components/implementation/no_interface/scheddev/sl.h
```diff
@@ -85,32 +85,48 @@ sl_cs_owner(void)
...
-	if (!ps_cas(&sl__globals()->lock.u.v, cached.v, csi.v)) goto retry;
+	if (!ps_cas(&sl__globals()->lock.u.v, cached.v, csi.v)) return 1;
+
+	return 0;
+}
+
+/* Enter into scheduler cs from a non-sched thread context */
+static inline void
+sl_cs_enter(void)
+{ while (sl_cs_enter_nospin()) ; }
```
Others: I like this refactoring, it makes the code a lot more clear! 👍

src/components/implementation/no_interface/scheddev/sl.h
```diff
@@ -85,32 +85,48 @@ sl_cs_owner(void)
...
+
+/* Enter into scheduler cs from scheduler thread context */
+static inline int
+sl_cs_enter_sched(void)
+{
+	int ret;
+
+	while ((ret = sl_cs_enter_nospin())) {
```
Others: Unnecessary parens here. Also this code seems overcomplicated. Couldn't it just early return on -EBUSY?

src/kernel/capinv.c
```diff
@@ -592,21 +571,20 @@ cap_update(struct pt_regs *regs, struct thread *thd_curr, struct thread *thd_nex
 		}
 	}
 
-	if (intr_context || switch_away) {
+	if (budget_expired) notify_parent(tcap_rcvcap_thd(tc_curr));
+	if (timer_intr_context || switch_away) {
 		thd_next = notify_process(thd_next, thd_curr, tc_next, tc_curr, &tc_next, 1);
-		if (thd_next == thd_curr && tc_next == tc_curr) return (intr_context ? 1 : 0);
+		if (thd_next == thd_curr && tc_next == tc_curr) return (timer_intr_context ? 1 : 0);
```
Others: Isn't `a ? 1 : 0` the same as `a` for boolean variables? Even if it isn't, the ternary here is confusing imho.

src/kernel/capinv.c
```diff
@@ -644,13 +623,16 @@ cap_thd_op(struct cap_thd *thd_cap, struct thread *thd, struct pt_regs *regs,
 		arcv_cap = (struct cap_arcv *)captbl_lkup(ci->captbl, arcv);
 		if (!CAP_TYPECHK_CORE(arcv_cap, CAP_ARCV)) return -EINVAL;
 
-		rcvt = arcv_cap->thd;
+		rcvt     = arcv_cap->thd;
```
Others: Nitpick: this spacing change seems pointless.

src/kernel/capinv.c
```diff
@@ -662,8 +644,9 @@ cap_thd_op(struct cap_thd *thd_cap, struct thread *thd, struct pt_regs *regs,
 
 		tcap_cap = (struct cap_tcap *)captbl_lkup(ci->captbl, tc);
 		if (!CAP_TYPECHK_CORE(tcap_cap, CAP_TCAP)) return -EINVAL;
-		tcap = tcap_cap->tcap;
+		tcap     = tcap_cap->tcap;
```
Others: Again, this spacing change is pointless and unnecessary.

src/kernel/include/thd.h
```diff
@@ -192,9 +193,32 @@ thd_rcvcap_evt_dequeue(struct thread *head) { return list_dequeue(&head->event_h
...
+static int
+thd_rcvcap_all_pending(struct thread *t)
+{
+	int pending = t->rcvcap.pending;
+
+	/* receive all pending */
+	t->rcvcap.pending = 0;
+	thd_rcvcap_all_pending_set(t, 0);
+
+	return ((pending<<1) | (list_first(&t->event_head) != NULL));
```
Others: This line feels a little arcane. Would a comment documenting it be helpful?

src/platform/linker/Makefile
```diff
@@ -7,6 +7,7 @@ LDFLAGS=-m elf_i386
 PRODUCTS=cos_linker gen_client_stub
 
 default: $(PRODUCTS)
+	@echo
```
Others: What does this change actually do?

src/components/implementation/no_interface/scheddev/sl_policy.h
```diff
@@ -1,23 +0,0 @@
...
-
-static inline struct sl_thd *
-sl_mod_thd_get(struct sl_thd_policy *tp)
-{ return &tp->thd; }
-
-static inline struct sl_thd_policy *
-sl_mod_thd_policy_get(struct sl_thd *t)
-{ return ps_container(t, struct sl_thd_policy, thd); }
-
-#endif /* SL_POLICY_H */
```
Others: Nice cleanup here!

src/kernel/capinv.c
```diff
@@ -438,21 +438,16 @@ cap_move(struct captbl *t, capid_t cap_to, capid_t capin_to,
...
-		if (!(curr->state & THD_STATE_SUSPENDED))
-			assert(!(curr->state & (THD_STATE_RCVING)) && !(curr->state & THD_STATE_PREEMPTED));
-		__userregs_set(regs, 0, __userregs_getsp(regs), __userregs_getip(regs));
-		return 0;
-	}
-
 	assert(next_ci && curr && next);
+	if (unlikely(curr == next)) return thd_switch_update(curr, regs, 1);
+
 	/* FIXME: trigger fault for the next thread, for now, return error */
```
Others: Is this FIXME still relevant? / Could it be easily fixed in this PR, or is fine to be left for later?

src/kernel/capinv.c
```diff
@@ -624,15 +602,16 @@ static int
...
-	sched_tok_t usr_counter   = (((sched_tok_t)__userregs_get3(regs) << 16) >> 16) | ((sched_tok_t)__userregs_getop(regs) << 16); /* op holds MSB of counter */
-	tcap_time_t timeout = (tcap_time_t)__userregs_get4(regs);
-	struct tcap *tcap   = tcap_current(cos_info);
+	struct thread *next       = thd_cap->t;
+	capid_t arcv              = (__userregs_get1(regs) << 16) >> 16;
+	capid_t tc                = __userregs_get1(regs) >> 16;
+	u32_t prio_higher         = __userregs_get3(regs);
+	u32_t prio_lower          = __userregs_get2(regs);
+	tcap_prio_t prio          = (tcap_prio_t)(prio_higher >> 16) << 32 | (tcap_prio_t)prio_lower;
+	sched_tok_t usr_counter   = (((sched_tok_t)__userregs_get3(regs) << 16) >> 16)
```
Others: What's the point of shifting something 16 one way, then immediately 16 the other way? Wouldn't a bit mask be a better fit here?

src/kernel/capinv.c
```diff
@@ -863,7 +854,8 @@ cap_introspect(struct captbl *ct, capid_t capid, u32_t op, unsigned long *retval
 	if (unlikely(!ch)) return -EINVAL;
 
 	switch(ch->type) {
-	case CAP_THD: return thd_introspect(((struct cap_thd*)ch)->t, op, retval);
+	case CAP_THD:  return thd_introspect(((struct cap_thd*)ch)->t, op, retval);
+	case CAP_TCAP: return tcap_introspect(((struct cap_tcap*)ch)->tcap, op, retval);
```
Others: I think there should be an explicit default case here, even if it does nothing.

src/kernel/include/thd.h
```diff
@@ -441,6 +465,42 @@ thd_preemption_state_update(struct thread *curr, struct thread *next, struct pt_
 	memcpy(&curr->regs, regs, sizeof(struct pt_regs));
 }
 
+static inline void 
+thd_rcvcap_pending_deliver(struct thread *thd, struct pt_regs *regs) {
+	unsigned long a = 0, b = 0;
```
Others: Are there any better names for `a` and `b` here?

src/components/implementation/no_interface/scheddev/sl.c
phanikishoreg: Makes sense, I'll add some comments.

src/components/implementation/no_interface/scheddev/sl.c
phanikishoreg: You'll need parenthesis around assignment and conditional, GCC spits out warnings if you don't.
And I don't think this code breaks CSG style.

src/components/implementation/no_interface/scheddev/sl.h
phanikishoreg: Declaration style.

src/components/implementation/no_interface/scheddev/sl.h
phanikishoreg: GCC warning! 
Can't do early return, retry is required for non zero return, only skip on ebusy!

src/kernel/capinv.c
phanikishoreg: Kk. But Ternary here is the most trivial and common one.

src/kernel/capinv.c
phanikishoreg: They are both ​doing the same. Also this is consistent with most bit operations in cos.

src/kernel/include/thd.h
phanikishoreg: Self documentary I think!

src/components/implementation/no_interface/scheddev/sl.c
phanikishoreg: It could be done in two lines and I think it's really preference. I'm open to change if that's concerning!

src/kernel/capinv.c
phanikishoreg: CSG requirement, I was just cleaning up.

src/platform/linker/Makefile
phanikishoreg: Just so there is no make error when there is nothing to do in PRODUCTS

Others:
Turns out it was git confusion on my end, looks like ppos was never broken. I apologize for the confusion.

src/components/implementation/no_interface/scheddev/sl.c
Others: Ah I understand now thanks.

src/components/implementation/no_interface/scheddev/sl.c
Others: I stand by this comment, I find this function really confusing.

src/components/implementation/no_interface/scheddev/sl.h
Others: Ah, right!

src/kernel/capinv.c
Others: Do you think a change is necessary here, or no?

src/kernel/capinv.c
Others: But it isn't aligning it with anything... Shouldn't it just be one space?

src/kernel/include/thd.h
Others: I respectfully disagree, this line definitely needs some comments.

src/kernel/capinv.c
phanikishoreg: No.

src/kernel/capinv.c
phanikishoreg: It is to an assignment 2 lines above in the same block.

src/kernel/include/thd.h
phanikishoreg: Hmm

src/kernel/capinv.c
Others: But they're split by an if statement, and the upper one isn't aligned with anything else... I think the extra spacing adds nothing here.

src/platform/linker/Makefile
Others: Awesome!

src/components/implementation/no_interface/scheddev/sl.h
gparmer: I forget positive reinforcement in code reviews.  Thanks for this @Others!b

src/components/implementation/no_interface/vkernel/micro_booter.c
```diff
@@ -62,7 +62,7 @@ dom0_io_fn(void *id)
 {
 	arcvcap_t rcvcap = dom0_vio_rcvcap((unsigned int)id);
 	while (1) {
-		int pending = cos_rcv(rcvcap);
+		int pending = cos_rcv(rcvcap, 0, NULL);
```
gparmer: Why `int pending =` at all?  Is there a compiler error otherwise?

src/components/include/cos_kernel_api.h
```diff
@@ -79,14 +79,21 @@ int cos_thd_switch(thdcap_t c);
 int cos_thd_wakeup(thdcap_t thd, tcap_t tc, tcap_prio_t prio, tcap_res_t res);
 #define CAP_NULL 0
 sched_tok_t cos_sched_sync(void);
+/*
+ * returns 0 on success and errno on failure:
+ * -EBUSY: if rcv has pending notifications and if current thread is the thread associated with rcv.
+ * -EAGAIN: if stok is outdated
+ * -EPERM: if tcap is not active (has no budget left)
+ * -EINVAL: any other error
+ */
```
gparmer: :+1:

src/kernel/capinv.c
gparmer: I think we have two options:

1. What you have.
2. `return !!timer_intr_context;`

Which do you think is more clear?

src/kernel/capinv.c
gparmer: the argument for doing it this ways is that the # of bits removed is easier to discern from `>> <<` than a mask (people don't think in hex).  Neither is correct; that is the argument.

The "correct" way is to use a macro/function, but we're customizing each system call so much that even this would be a mess.

src/kernel/capinv.c
gparmer: This is a grey area in the CSG.  General rule:  if it looks better, do it.  That argument could be made here.  The converse argument could be made as well.  The author wins these arguments ;-)

src/kernel/capinv.c
gparmer: says the person coming from PLs with required exhaustive pattern matching ;-)  I'd tend to agree, but I don't know if it is worth making this change without making a comparable change across the code-base.

src/kernel/include/tcap.h
```diff
@@ -269,4 +269,14 @@ tcap_higher_prio(struct tcap *a, struct tcap *c)
 	return ret;
 }
 
+static inline int
+tcap_introspect(struct tcap *t, unsigned long op, unsigned long *retval)
+{
+	switch(op) {
+	case TCAP_GET_BUDGET: *retval = t->budget.cycles; break;
+	default: return -EINVAL;
```
gparmer: if you're putting these all on the same line, then at least line up the "bodies" of each case/default so that it is easier to visually parse what's happening.

src/kernel/include/thd.h
gparmer: I think that what's making it slightly confusing is the `!=` and conditional.

Better might be `return ((pending << 1) | list_empty(&t->event_head));`  Should we add that function?

Notice the extra spaces around the binary operator change here as well (`pending<<1` vs. `pending << 1`).

src/kernel/include/thd.h
```diff
@@ -192,9 +193,32 @@ thd_rcvcap_evt_dequeue(struct thread *head) { return list_dequeue(&head->event_h
...
+
+	return ((pending<<1) | (list_first(&t->event_head) != NULL));
+}
+
 static int
 thd_rcvcap_pending(struct thread *t)
-{ return t->rcvcap.pending || list_first(&t->event_head) != NULL; }
+{
+	if (t->rcvcap.pending) return t->rcvcap.pending;
+	return list_first(&t->event_head) != NULL;
```
gparmer: also `!list_empty(...)`?

src/kernel/include/thd.h
```diff
@@ -441,6 +465,42 @@ thd_preemption_state_update(struct thread *curr, struct thread *next, struct pt_
...
+	if (all_pending) {
+		__userregs_setretvals(regs, thd_rcvcap_all_pending(thd), a, b);
+	} else {
+		thd_rcvcap_pending_dec(thd);
+		__userregs_setretvals(regs, thd_rcvcap_pending(thd), a, b);
+	}
+}
+
+static int
+thd_switch_update(struct thread *thd, struct pt_regs *regs, int issame)
```
gparmer: I don't know 1. where all this is called from, nor 2. how many times it is called.  If it is on a fast-path, and not called that often, `inline` might be appropriate.  Up to you.

src/components/implementation/no_interface/scheddev/sl.c
phanikishoreg: Will separate conditional and assignment here.

src/kernel/capinv.c
phanikishoreg: I think I should just do `return timer_intr_context;` as it's a binary value telling if it's in the timer interrupt context, I didn't think of that before! Thank you @Others for bringing it up though it's really not part of my code change!

src/kernel/include/thd.h
phanikishoreg: Got it. This is called from `kernel/capinv.c:cap_thd_switch()`, I should use `inline` here, it's on a fast-path.

src/components/implementation/no_interface/vkernel/micro_booter.c
phanikishoreg: Not really, I must have added it for prints. Got rid of them (in 2 places).

phanikishoreg:
Updated the PR with review fixes. Thank you!

PR #247, opened 2017-06-26
Fix hardcoded stack size and alignment

Others:
### Summary of this PR

Thread stack size is supposed to be determined by a constant in
consts.h. A lot of the code in the codebase relies on this,
including the code for fetching the a thread's id. However, the
assembly code that was really allocating the stack was using a
hardcoded value.

This mismatch led to invalid thread id's being returned, and
broke increasing stack sizes to avoid stack overflows.

To fix this, I made the assembly dependent on #defines, and made
sure those defines were available for those files. I also fixed
how those defines were structured in the first place.

### Code Quality

Style:

Code Craftsmanship:

### DIFF
--- a/src/components/include/cos_asm_simple_stacks.h
+++ b/src/components/include/cos_asm_simple_stacks.h
@@ -1,13 +1,17 @@
 #ifndef COS_ASM_SIMPLE_STACKS_H
 #define COS_ASM_SIMPLE_STACKS_H
 
+#ifndef MAX_STACK_SZ_BYTE_ORDER
+#error "Missing MAX_STACK_SZ_BYTE_ORDER, try including consts.h"
+#endif
+
 #define COS_ASM_GET_STACK                   \
 	movl $cos_static_stack, %esp;	    \
 	movl %eax, %edx;		    \
 	andl $0xffff, %eax;		    \
-	shl $12, %eax;                      \
+	shl $MAX_STACK_SZ_BYTE_ORDER, %eax;                      \
 	addl %eax, %esp;		    \
-	shr $12, %eax;			    \
+	shr $MAX_STACK_SZ_BYTE_ORDER, %eax;			    \
 	shr $16, %edx;			    \
 	pushl %edx;			    \
 	pushl %eax;
--- a/src/kernel/include/shared/consts.h
+++ b/src/kernel/include/shared/consts.h
@@ -44,20 +44,25 @@ struct pt_regs {
 #endif
 #define PAGE_ORDER 12
 #ifndef __KERNEL__
-#define PAGE_SIZE (1<<PAGE_ORDER)
+#define PAGE_SIZE (1 << PAGE_ORDER)
 #endif
 
 #define MAX_SERVICE_DEPTH 31
 #define MAX_NUM_THREADS 64
+
 /* Stacks are 2 * page_size (expressed in words) */
-#define MAX_STACK_SZ    (PAGE_SIZE/4) /* a page */
-#define COS_STACK_SZ    (MAX_STACK_SZ*4)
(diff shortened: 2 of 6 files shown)


Others:
I think this is ready for a review @gparmer. Do I need someone else to look through it too?

src/components/include/cos_asm_simple_stacks.h
```diff
@@ -1,13 +1,17 @@
...
+#ifndef MAX_STACK_SZ_BYTE_ORDER
+#error "Missing MAX_STACK_SZ_BYTE_ORDER, try including consts.h"
+#endif
+
 #define COS_ASM_GET_STACK                   \
 	movl $cos_static_stack, %esp;	    \
 	movl %eax, %edx;		    \
 	andl $0xffff, %eax;		    \
-	shl $12, %eax;                      \
+	shl $MAX_STACK_SZ_BYTE_ORDER, %eax;                      \
```
base0x10: Consistent spacing with the rest of the file

src/components/include/cos_asm_simple_stacks.h
```diff
@@ -1,13 +1,17 @@
...
 	addl %eax, %esp;		    \
-	shr $12, %eax;			    \
+	shr $MAX_STACK_SZ_BYTE_ORDER, %eax;			    \
```
base0x10: Consistent spacing with the rest of the file

src/components/include/cos_asm_simple_stacks.h
```diff
@@ -1,13 +1,17 @@
 #ifndef COS_ASM_SIMPLE_STACKS_H
 #define COS_ASM_SIMPLE_STACKS_H
 
+#ifndef MAX_STACK_SZ_BYTE_ORDER
+#error "Missing MAX_STACK_SZ_BYTE_ORDER, try including consts.h"
```
zacharied: I'm not one to complain about error checking, but this style of error reporting seems unconventional compared to the rest of Composite's codebase.

Others: I don't think there is a better way in this case... Any suggestions?

src/kernel/include/shared/consts.h
```diff
@@ -44,20 +44,25 @@ struct pt_regs {
...
 #define PAGE_ORDER 12
 #ifndef __KERNEL__
-#define PAGE_SIZE (1<<PAGE_ORDER)
+#define PAGE_SIZE (1 << PAGE_ORDER)
 #endif
 
 #define MAX_SERVICE_DEPTH 31
 #define MAX_NUM_THREADS 64
+
 /* Stacks are 2 * page_size (expressed in words) */
```
base0x10: I know that you didn't make it incorrect, but you should either make the comment accurate or remove it

src/components/include/cos_asm_simple_stacks.h
gparmer: Generally, this is a good way to check errors (a static assertion).  However, in this case, won't this just result a compile-time error if it is not defined?  So you add this, it gives a compile-time error; you don't, and it does the same?  If that's the case, I'd always err on less code = better.

I'm going to likely apply this anyway, since you're rushed, but another PR to change this (if you agree) would be good.

src/components/include/cos_asm_simple_stacks.h
gparmer: Yeah, fix your `\` alignment.  Same deal as above.

src/kernel/include/shared/consts.h
gparmer: +1 @base0x10 

Comments fall out of coherence with code easily.  Removing them if this is likely to happen is often the right move.

PR #249, opened 2017-06-27
Implement sl_lock, a library for locking built on sl

Others:
### Summary of this PR
sl_lock provides a few basic primitives for locking. I think it
might be useful for projects other than cFE2cos, for which it was
initially developed.

### Code Quality

Style:

Code Craftsmanship:

### DIFF
--- a/src/components/implementation/no_interface/scheddev/sl.c
+++ b/src/components/implementation/no_interface/scheddev/sl.c
@@ -104,11 +104,10 @@ done:
 }
 
 void
-sl_thd_yield(thdid_t tid)
+sl_thd_yield_cs_exit(thdid_t tid)
 {
 	struct sl_thd *t = sl_thd_curr();
 
-	sl_cs_enter();
 	if (tid) {
 		struct sl_thd *to = sl_thd_lkup(tid);
 
@@ -118,8 +117,13 @@ sl_thd_yield(thdid_t tid)
 		sl_mod_yield(sl_mod_thd_policy_get(t), NULL);
 		sl_cs_exit_schedule();
 	}
+}
 
-	return;
+void
+sl_thd_yield(thdid_t tid)
+{
+	sl_cs_enter();
+	sl_thd_yield_cs_exit(tid);
 }
 
 static struct sl_thd *
--- a/src/components/implementation/no_interface/scheddev/sl.c
+++ b/src/components/implementation/no_interface/scheddev/sl.c
@@ -259,8 +259,8 @@ sl_sched_loop(void)
 			if (unlikely(t == sl__globals()->idle_thd)) continue;
 
 			/*
-			 * receiving scheduler notifications is not in critical section mainly for 
-			 * 1. scheduler thread can often be blocked in rcv, which can add to 
+			 * receiving scheduler notifications is not in critical section mainly for
+			 * 1. scheduler thread can often be blocked in rcv, which can add to
 			 *    interrupt execution or even AEP thread execution overheads.
 			 * 2. scheduler events are not acting on the sl_thd or the policy structures, so
 			 *    having finer grained locks around the code that modifies sl_thd states is better.
--- a/src/components/implementation/no_interface/scheddev/sl.h
+++ b/src/components/implementation/no_interface/scheddev/sl.h
@@ -83,8 +83,10 @@ static inline thdid_t
 sl_thdid(void)
 {
 	thdid_t tid = cos_thdid();
+
 	assert(tid != 0);
 	assert(tid < MAX_NUM_THREADS);
+	
 	return tid;
 }
 
--- a/src/components/implementation/no_interface/scheddev/sl.h
+++ b/src/components/implementation/no_interface/scheddev/sl.h
@@ -73,11 +73,29 @@ sl_thd_setprio(struct sl_thd *t, tcap_prio_t p)
 
 static inline struct sl_thd *
 sl_thd_lkup(thdid_t tid)
-{ return sl_mod_thd_get(sl_thd_lookup_backend(tid)); }
+{
+	assert(tid != 0);
+	if (tid > MAX_NUM_THREADS) {
+		return NULL;
+	}
+	return sl_mod_thd_get(sl_thd_lookup_backend(tid));
+}
+
+static inline thdid_t
+sl_thd_curr_id(void)
+{
+	thdid_t tid = cos_thdid();
+	assert(tid != 0);
+	assert(tid < MAX_NUM_THREADS);
+	return tid;
+}
+
 
 static inline struct sl_thd *
 sl_thd_curr(void)
-{ return sl_thd_lkup(cos_thdid()); }
+{
+	return sl_thd_lkup(sl_thd_curr_id());
+}
 
 /* are we the owner of the critical section? */
 static inline int
@@ -94,7 +112,7 @@ sl_cs_owner(void)
  * @ret:
  *     (Caller of this function should retry for a non-zero return value.)
  *     1 for cas failure or after successful thread switch to thread that owns the lock.
- *     -ve from cos_defswitch failure, allowing caller for ex: the scheduler thread to 
+ *     -ve from cos_defswitch failure, allowing caller for ex: the scheduler thread to
  *     check if it was -EBUSY to first recieve pending notifications before retrying lock.
  */
 int sl_cs_enter_contention(union sl_cs_intern *csi, union sl_cs_intern *cached, thdcap_t curr, sched_tok_t tok);
@@ -174,9 +192,22 @@ retry:
 	if (!ps_cas(&sl__globals()->lock.u.v, cached.v, 0))    goto retry;
 }
 
+static inline microsec_t
+sl_cyc2usec(cycles_t cyc)
+{ return cyc / sl__globals()->cyc_per_usec; }
+
+static inline microsec_t
+sl_usec2cyc(microsec_t usec)
+{ return usec * sl__globals()->cyc_per_usec; }
+
 static inline cycles_t
 sl_now(void)
 { return ps_tsc(); }
+
+static inline microsec_t
+sl_now_usec(void)
+{ return sl_cyc2usec(sl_now()); }
+
 
 /*
  * Do a few things: 1. take the critical section if it isn't already
@@ -313,14 +344,6 @@ sl_timeout_oneshot(cycles_t absolute_us)
 static inline void
 sl_timeout_relative(cycles_t offset)
 { sl_timeout_oneshot(sl_now() + offset); }
-
-static inline microsec_t
-sl_cyc2usec(cycles_t cyc)
-{ return cyc / sl__globals()->cyc_per_usec; }
-
-static inline microsec_t
-sl_usec2cyc(microsec_t usec)
-{ return usec * sl__globals()->cyc_per_usec; }
 
 void sl_thd_param_set(struct sl_thd *t, sched_param_t sp);
 
--- a/src/components/implementation/no_interface/scheddev/sl.h
+++ b/src/components/implementation/no_interface/scheddev/sl.h
@@ -75,14 +75,12 @@ static inline struct sl_thd *
 sl_thd_lkup(thdid_t tid)
 {
 	assert(tid != 0);
-	if (tid > MAX_NUM_THREADS) {
-		return NULL;
-	}
+	if (unlikely(tid > MAX_NUM_THREADS)) return NULL;
 	return sl_mod_thd_get(sl_thd_lookup_backend(tid));
 }
 
 static inline thdid_t
-sl_thd_curr_id(void)
+sl_thdid(void)
 {
 	thdid_t tid = cos_thdid();
 	assert(tid != 0);
@@ -93,9 +91,7 @@ sl_thd_curr_id(void)
 
 static inline struct sl_thd *
 sl_thd_curr(void)
-{
-	return sl_thd_lkup(sl_thd_curr_id());
-}
+{ return sl_thd_lkup(sl_thdid()); }
 
 /* are we the owner of the critical section? */
 static inline int
@@ -310,6 +306,7 @@ void sl_thd_block(thdid_t tid);
 /* wakeup a thread that has (or soon will) block */
 void sl_thd_wakeup(thdid_t tid);
 void sl_thd_yield(thdid_t tid);
+void sl_thd_yield_cs_exit(thdid_t tid);
 
 /* The entire thread allocation and free API */
 struct sl_thd *sl_thd_alloc(cos_thd_fn_t fn, void *data);
--- a/src/components/implementation/no_interface/scheddev/sl_lock.c
+++ b/src/components/implementation/no_interface/scheddev/sl_lock.c
@@ -1,0 +1,110 @@
+#include "sl_lock.h"
+
+#include "sl.h"
+
+void sl_lock_init(struct sl_lock *lock) {
+    assert(lock);
+
+    *lock = (struct sl_lock) {
+        .holder = 0
+    };
+}
+
+thdid_t sl_lock_holder(struct sl_lock *lock) {
+    assert(lock);
+    return lock->holder;
+}
+
+// If timeout is 0, ignore the timeout
+int sl_lock_timeout_elapsed(microsec_t start_time, microsec_t timeout) {
+    if (timeout == 0 || (sl_now_usec() - start_time) <= timeout) {
+        return 0;
+    }
+    return 1;
+}
+
+int sl_lock_lock_internal_no_cs(struct sl_lock *lock, microsec_t timeout) {
+    assert(lock);
+
+    microsec_t start_time = sl_now_usec();
+    while (lock->holder != 0 && !sl_lock_timeout_elapsed(start_time, timeout)) {
+        thdid_t holder = lock->holder;
+
+        /*
+         * If we are preempted after the exit, and the holder is no longer holding
+         * the critical section, then we will yield to them and possibly waste a
+         * time-slice.  This will be fixed the next iteration, as we will see an
+         * updated value of the holder, but we essentially lose a timeslice in the
+         * worst case.  From a real-time perspective, this is bad, but we're erring
+         * on simplicity here.
+         */
+        sl_cs_exit();
+        sl_thd_yield(holder);
+        sl_cs_enter();
+    }
+
+    int took_lock = 0;
+    if (lock->holder == 0) {
+        lock->holder = sl_thd_curr_id();
+        took_lock = 1;
+    }
+    return took_lock;
+}
+
+
+int sl_lock_lock_internal(struct sl_lock *lock, microsec_t timeout) {
+    sl_cs_enter();
+    int result = sl_lock_lock_internal_no_cs(lock, timeout);
+    sl_cs_exit();
+    return result;
+}
+
+void sl_lock_lock(struct sl_lock *lock) {
+    assert(lock);
+
+    // Lock internally, which should always succeed since timeout is 0
+    assert(sl_lock_lock_internal(lock, 0) == 1);
+}
+
+void sl_lock_lock_no_cs(struct sl_lock *lock) {
+    assert(lock);
+
+    // Lock internally, which should always succeed since timeout is 0
+    assert(sl_lock_lock_internal_no_cs(lock, 0) == 1);
+}
+
+
+int sl_lock_timed_lock(struct sl_lock *lock, microsec_t timeout) {
+    assert(lock);
+
+    if (timeout == 0) {
+        return 0;
+    }
+    return sl_lock_lock_internal(lock, timeout);
+}
+
+int sl_lock_timed_lock_no_cs(struct sl_lock *lock, microsec_t timeout) {
+    assert(lock);
+
+    if (timeout == 0) {
+        return 0;
+    }
+    return sl_lock_lock_internal_no_cs(lock, timeout);
+}
+
+
+void sl_lock_unlock(struct sl_lock *lock) {
+    assert(lock);
+    sl_cs_enter();
+
+    sl_lock_unlock_no_cs(lock);
+
+    sl_cs_exit();
+}
+
+void sl_lock_unlock_no_cs(struct sl_lock *lock) {
+    assert(lock);
+
+    assert(lock->holder == sl_thd_curr_id());
+    lock->holder = 0;
+}
--- a/src/components/implementation/no_interface/scheddev/sl_lock.c
+++ b/src/components/implementation/no_interface/scheddev/sl_lock.c
@@ -4,14 +4,12 @@
 void
 sl_lock_init(struct sl_lock *lock)
 {
-    assert(lock);
     *lock = SL_LOCK_STATIC_INIT();
 }
 
 thdid_t
 sl_lock_holder(struct sl_lock *lock)
 {
-    assert(lock);
     return lock->holder;
 }
 
@@ -20,8 +18,6 @@ sl_lock_timed_take(struct sl_lock *lock, microsec_t max_wait_time)
 {
     int result;
     cycles_t deadline = sl_now() + sl_usec2cyc(max_wait_time);
-
-    assert(lock);
 
     sl_cs_enter();
     while (lock->holder != 0 && sl_now() < deadline) {
--- a/src/components/implementation/no_interface/scheddev/sl_lock.c
+++ b/src/components/implementation/no_interface/scheddev/sl_lock.c
@@ -15,23 +15,8 @@ sl_lock_holder(struct sl_lock *lock)
     return lock->holder;
 }
 
-
-void
-sl_lock_lock(struct sl_lock *lock)
-{
-    assert(lock);
-
-    sl_cs_enter();
-    while (lock->holder != 0) {
-        sl_thd_yield_cs_exit(lock->holder);
-        sl_cs_enter();
-    }
-    lock->holder = sl_thdid();
-    sl_cs_exit();
-}
-
 int
-sl_lock_timed_lock(struct sl_lock *lock, microsec_t max_wait_time)
+sl_lock_timed_take(struct sl_lock *lock, microsec_t max_wait_time)
 {
     int result;
     cycles_t deadline = sl_now() + sl_usec2cyc(max_wait_time);
@@ -47,21 +32,9 @@ sl_lock_timed_lock(struct sl_lock *lock, microsec_t max_wait_time)
     if (lock->holder == 0) {
         lock->holder = sl_thdid();
         result = 1;
-    }else {
+    } else {
         result = 0;
     }
     sl_cs_exit();
     return result;
-}
-
-void
-sl_lock_unlock(struct sl_lock *lock)
-{
-    assert(lock);
-
-    sl_cs_enter();
-    assert(lock->holder);
-    assert(lock->holder == sl_thdid());
-    lock->holder = 0;
-    sl_cs_exit();
 }
--- a/src/components/implementation/no_interface/scheddev/sl_lock.c
+++ b/src/components/implementation/no_interface/scheddev/sl_lock.c
@@ -1,6 +1,5 @@
-#include "sl_lock.h"
-
 #include "sl.h"
+#include "sl_lock.h"
 
 void
 sl_lock_init(struct sl_lock *lock)
@@ -35,11 +34,11 @@ int
 sl_lock_timed_lock(struct sl_lock *lock, microsec_t max_wait_time)
 {
     int result;
+    cycles_t deadline = sl_now() + sl_usec2cyc(max_wait_time);
 
     assert(lock);
 
     sl_cs_enter();
-    cycles_t deadline = sl_now() + sl_usec2cyc(max_wait_time);
     while (lock->holder != 0 && sl_now() < deadline) {
         sl_thd_yield_cs_exit(lock->holder);
         sl_cs_enter();
--- a/src/components/implementation/no_interface/scheddev/sl_lock.c
+++ b/src/components/implementation/no_interface/scheddev/sl_lock.c
@@ -2,109 +2,67 @@
 
 #include "sl.h"
 
-void sl_lock_init(struct sl_lock *lock) {
+void
+sl_lock_init(struct sl_lock *lock)
+{
     assert(lock);
-
-    *lock = (struct sl_lock) {
-        .holder = 0
-    };
+    *lock = SL_LOCK_STATIC_INIT();
 }
 
-thdid_t sl_lock_holder(struct sl_lock *lock) {
+thdid_t
+sl_lock_holder(struct sl_lock *lock)
+{
     assert(lock);
     return lock->holder;
 }
 
-// If timeout is 0, ignore the timeout
-int sl_lock_timeout_elapsed(microsec_t start_time, microsec_t timeout) {
-    if (timeout == 0 || (sl_now_usec() - start_time) <= timeout) {
-        return 0;
-    }
-    return 1;
-}
 
-int sl_lock_lock_internal_no_cs(struct sl_lock *lock, microsec_t timeout) {
+void
+sl_lock_lock(struct sl_lock *lock)
+{
     assert(lock);
 
-    microsec_t start_time = sl_now_usec();
-    while (lock->holder != 0 && !sl_lock_timeout_elapsed(start_time, timeout)) {
-        thdid_t holder = lock->holder;
+    sl_cs_enter();
+    while (lock->holder != 0) {
+        sl_thd_yield_cs_exit(lock->holder);
+        sl_cs_enter();
+    }
+    lock->holder = sl_thdid();
+    sl_cs_exit();
+}
 
-        /*
-         * If we are preempted after the exit, and the holder is no longer holding
-         * the critical section, then we will yield to them and possibly waste a
-         * time-slice.  This will be fixed the next iteration, as we will see an
-         * updated value of the holder, but we essentially lose a timeslice in the
-         * worst case.  From a real-time perspective, this is bad, but we're erring
-         * on simplicity here.
-         */
-        sl_cs_exit();
-        sl_thd_yield(holder);
+int
+sl_lock_timed_lock(struct sl_lock *lock, microsec_t max_wait_time)
+{
+    int result;
+
+    assert(lock);
+
+    sl_cs_enter();
+    cycles_t deadline = sl_now() + sl_usec2cyc(max_wait_time);
+    while (lock->holder != 0 && sl_now() < deadline) {
+        sl_thd_yield_cs_exit(lock->holder);
         sl_cs_enter();
     }
 
-    int took_lock = 0;
     if (lock->holder == 0) {
-        lock->holder = sl_thd_curr_id();
-        took_lock = 1;
+        lock->holder = sl_thdid();
+        result = 1;
+    }else {
+        result = 0;
     }
-    return took_lock;
-}
-
-
-int sl_lock_lock_internal(struct sl_lock *lock, microsec_t timeout) {
-    sl_cs_enter();
-    int result = sl_lock_lock_internal_no_cs(lock, timeout);
     sl_cs_exit();
     return result;
 }
 
-void sl_lock_lock(struct sl_lock *lock) {
+void
+sl_lock_unlock(struct sl_lock *lock)
+{
     assert(lock);
 
-    // Lock internally, which should always succeed since timeout is 0
-    assert(sl_lock_lock_internal(lock, 0) == 1);
-}
-
-void sl_lock_lock_no_cs(struct sl_lock *lock) {
-    assert(lock);
-
-    // Lock internally, which should always succeed since timeout is 0
-    assert(sl_lock_lock_internal_no_cs(lock, 0) == 1);
-}
-
-
-int sl_lock_timed_lock(struct sl_lock *lock, microsec_t timeout) {
-    assert(lock);
-
-    if (timeout == 0) {
-        return 0;
-    }
-    return sl_lock_lock_internal(lock, timeout);
-}
-
-int sl_lock_timed_lock_no_cs(struct sl_lock *lock, microsec_t timeout) {
-    assert(lock);
-
-    if (timeout == 0) {
-        return 0;
-    }
-    return sl_lock_lock_internal_no_cs(lock, timeout);
-}
-
-
-void sl_lock_unlock(struct sl_lock *lock) {
-    assert(lock);
     sl_cs_enter();
-
-    sl_lock_unlock_no_cs(lock);
-
-    sl_cs_exit();
-}
-
-void sl_lock_unlock_no_cs(struct sl_lock *lock) {
-    assert(lock);
-
-    assert(lock->holder == sl_thd_curr_id());
+    assert(lock->holder);
+    assert(lock->holder == sl_thdid());
     lock->holder = 0;
+    sl_cs_exit();
 }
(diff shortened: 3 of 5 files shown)


gparmer:
@Others please @ mention people.

@phanikishoreg

src/components/implementation/no_interface/scheddev/sl.h
```diff
@@ -73,11 +73,29 @@ sl_thd_setprio(struct sl_thd *t, tcap_prio_t p)
 
 static inline struct sl_thd *
 sl_thd_lkup(thdid_t tid)
-{ return sl_mod_thd_get(sl_thd_lookup_backend(tid)); }
+{
+	assert(tid != 0);
+	if (tid > MAX_NUM_THREADS) {
```
gparmer: I'd also de-emphasize the error case (which is the prominent part of the function right now) by putting the `return NULL;` on the same line as the `if`, and not use `{}`.  This is a common pattern for error paths to de-emphasize them, and make the main path easier to read.

src/components/implementation/no_interface/scheddev/sl.h
```diff
@@ -73,11 +73,29 @@ sl_thd_setprio(struct sl_thd *t, tcap_prio_t p)
...
+		return NULL;
+	}
+	return sl_mod_thd_get(sl_thd_lookup_backend(tid));
+}
+
+static inline thdid_t
+sl_thd_curr_id(void)
```
gparmer: Can likely just be `sl_thdid` as the `void` argument makes it clear it is for this thread.  That would also be consistent with the cos call for the same.

src/components/implementation/no_interface/scheddev/sl.h
```diff
@@ -73,11 +73,29 @@ sl_thd_setprio(struct sl_thd *t, tcap_prio_t p)
...
+{
+	thdid_t tid = cos_thdid();
+	assert(tid != 0);
```
gparmer: de-emphasize the error checking and use a single assert with `&&`?

Up to you.  I understand why you did it this way.

src/components/implementation/no_interface/scheddev/sl.h
```diff
@@ -73,11 +73,29 @@ sl_thd_setprio(struct sl_thd *t, tcap_prio_t p)
...
+	assert(tid < MAX_NUM_THREADS);
+	return tid;
+}
+
 
 static inline struct sl_thd *
 sl_thd_curr(void)
-{ return sl_thd_lkup(cos_thdid()); }
+{
```
gparmer: inconsistent to make this change, but then use the compact style later.  I'd prefer the compact style for single-line functions.

src/components/implementation/no_interface/scheddev/sl.h
```diff
@@ -174,10 +192,23 @@ sl_cs_exit(void)
 	if (!ps_cas(&sl__globals()->lock.u.v, cached.v, 0))    goto retry;
 }
 
+static inline microsec_t
```
gparmer: Just a note that the math in these is expensive, and prone to wrap-around issues.  We don't want to use it much.  This code is perfectly good, this is just a note to you for the future.

Concretely:  we want to keep things in cycles, in cycles where possible.

gparmer: Ah, I see, these just got moved from below.

src/components/implementation/no_interface/scheddev/sl_lock.c
```diff
@@ -0,0 +1,110 @@
+#include "sl_lock.h"
+
```
gparmer: no space

src/components/implementation/no_interface/scheddev/sl_lock.c
```diff
@@ -0,0 +1,110 @@
...
+#include "sl.h"
+
+void sl_lock_init(struct sl_lock *lock) {
+    assert(lock);
+
+    *lock = (struct sl_lock) {
```
gparmer: This should use your static initializer.

`#define SL_LOCK_STATIC_INIT() (struct sl_lock){ .holder = 0 }`

Casts don't have a space between themselves and what they modify.

src/components/implementation/no_interface/scheddev/sl_lock.c
```diff
@@ -0,0 +1,110 @@
+#include "sl_lock.h"
+
+#include "sl.h"
+
+void sl_lock_init(struct sl_lock *lock) {
```
gparmer: void on separate line.

src/components/implementation/no_interface/scheddev/sl_lock.c
```diff
@@ -0,0 +1,110 @@
...
+    assert(lock);
+
+    *lock = (struct sl_lock) {
+        .holder = 0
+    };
+}
+
+thdid_t sl_lock_holder(struct sl_lock *lock) {
```
gparmer: type on separate line throughout....

src/components/implementation/no_interface/scheddev/sl_lock.c
```diff
@@ -0,0 +1,110 @@
...
+    assert(lock);
+    return lock->holder;
+}
+
+// If timeout is 0, ignore the timeout
```
gparmer: /* */

src/components/implementation/no_interface/scheddev/sl_lock.c
gparmer: { on separate line.

src/components/implementation/no_interface/scheddev/sl_lock.c
gparmer: { separate line throughout

src/components/implementation/no_interface/scheddev/sl_lock.c
```diff
@@ -0,0 +1,110 @@
...
+int sl_lock_timeout_elapsed(microsec_t start_time, microsec_t timeout) {
```
gparmer: it is ambiguous here if timeout is relative to start-time, or if it is it is an absolute time.  Often you "set a timeout for 2 seconds from now", often conflating timeout with "deadline".

call it `delay` instead?

src/components/implementation/no_interface/scheddev/sl_lock.c
```diff
@@ -0,0 +1,110 @@
...
+    if (timeout == 0 || (sl_now_usec() - start_time) <= timeout) {
```
gparmer: Just `return !(...);`

Idiomatic c

src/components/implementation/no_interface/scheddev/sl_lock.c
```diff
@@ -0,0 +1,110 @@
...
+        return 0;
+    }
+    return 1;
+}
+
+int sl_lock_lock_internal_no_cs(struct sl_lock *lock, microsec_t timeout) {
```
gparmer: The verbs should likely be `take` and `release` instead of `lock` and `unlock` to avoid this type of naming.

gparmer: What is the point of the return value?

Do you only have it because cFE requires it?

src/components/implementation/no_interface/scheddev/sl_lock.c
```diff
@@ -0,0 +1,110 @@
...
+    assert(lock);
+
+    microsec_t start_time = sl_now_usec();
```
gparmer: declarations at top

gparmer: The only reason to use `start_time` and `delay` in the timing APIs is to avoid races that plague solely relative time APIs.  Not taking the start time from the user makes this point moot.  So I don't understand why `sl_lock_timeout_elapsed` exists.  If we don't care about that race (which we can talk about), then we should just calculate the deadline immediately (`microsec_t deadline = sl_now_usec() + timeout;`), and just compare that to now_us.

src/components/implementation/no_interface/scheddev/sl_lock.h
```diff
@@ -0,0 +1,24 @@
+#ifndef SL_LOCK_H
+#define SL_LOCK_H
+
+#include <cos_defkernel_api.h>
+
+struct sl_lock {
+    // will be 0 if no one holds the lock
+    thdid_t holder;
```
gparmer: should be `volatile`

src/components/implementation/no_interface/scheddev/sl_lock.c
```diff
@@ -0,0 +1,110 @@
...
+
+        /*
+         * If we are preempted after the exit, and the holder is no longer holding
+         * the critical section, then we will yield to them and possibly waste a
+         * time-slice.  This will be fixed the next iteration, as we will see an
+         * updated value of the holder, but we essentially lose a timeslice in the
+         * worst case.  From a real-time perspective, this is bad, but we're erring
+         * on simplicity here.
+         */
+        sl_cs_exit();
```
gparmer: strange to take the lock, only to release it again in `yield`.  This can lead to the case you point out.  Is there no `yield_no_cs` which would solve this problem?

src/components/implementation/no_interface/scheddev/sl_lock.c
```diff
@@ -0,0 +1,110 @@
...
+        sl_thd_yield(holder);
+        sl_cs_enter();
+    }
+
+    int took_lock = 0;
```
gparmer: decl at top

src/components/implementation/no_interface/scheddev/sl_lock.c
```diff
@@ -0,0 +1,110 @@
...
+    if (lock->holder == 0) {
```
gparmer: Not maintaining error free path at same level of indentation.

```
if (lock->holder) return 0;
lock->holder = sl_thd_curr_id();

return 1;
```

src/components/implementation/no_interface/scheddev/sl.h
```diff
@@ -73,11 +73,29 @@ sl_thd_setprio(struct sl_thd *t, tcap_prio_t p)
...
+}
+
+static inline thdid_t
+sl_thd_curr_id(void)
+{
+	thdid_t tid = cos_thdid();
+	assert(tid != 0);
+	assert(tid < MAX_NUM_THREADS);
+	return tid;
+}
```
phanikishoreg: This function doesn't seem to have any relation to SL library. You should either have a wrapper around `sl_thd_curr` so you validate that this thread is in fact a `sl_thd` or you just don't need this function. Using `cos_thdid()` directly and validating after it should be fine!

Style: Declarations, "space", body, "space", return.

src/components/implementation/no_interface/scheddev/sl_lock.c
```diff
@@ -0,0 +1,110 @@
...
+    }
+    return took_lock;
+}
+
+
+int sl_lock_lock_internal(struct sl_lock *lock, microsec_t timeout) {
+    sl_cs_enter();
+    int result = sl_lock_lock_internal_no_cs(lock, timeout);
+    sl_cs_exit();
+    return result;
```
gparmer: space before `return`

src/components/implementation/no_interface/scheddev/sl.h
```diff
@@ -73,11 +73,29 @@ sl_thd_setprio(struct sl_thd *t, tcap_prio_t p)
...
+
 
 static inline struct sl_thd *
 sl_thd_curr(void)
-{ return sl_thd_lkup(cos_thdid()); }
+{
+	return sl_thd_lkup(sl_thd_curr_id());
+}
```
phanikishoreg: Why this change?

src/components/implementation/no_interface/scheddev/sl.h
```diff
@@ -174,10 +192,23 @@ sl_cs_exit(void)
...
+{ return usec * sl__globals()->cyc_per_usec; }
+
 static inline cycles_t
 sl_now(void)
 { return ps_tsc(); }
 
+static inline microsec_t
+sl_now_usec(void)
+{ return sl_cyc2usec(sl_now()); }
+
```
phanikishoreg: Perhaps all these changes should be a separated from the locking commit!

src/components/implementation/no_interface/scheddev/sl_lock.c
phanikishoreg: function definition style. 
```
void
sl_lock_init(struct sl_lock *lock)
{
...
}
```

This applies to all functions in this file.

src/components/implementation/no_interface/scheddev/sl_lock.c
```diff
@@ -0,0 +1,110 @@
...
+}
+
+void sl_lock_lock(struct sl_lock *lock) {
+    assert(lock);
+
+    // Lock internally, which should always succeed since timeout is 0
+    assert(sl_lock_lock_internal(lock, 0) == 1);
```
gparmer: We really don't want all of the code including the timeout calculations to be run for the fast path of taking a lock.  This needs to be really fast.

I don't see a way to share the implementations while still making it efficient (without lots of `ALWAYS_INLINE` hackery.

Since the the non-deadline lock will have a very small body (3 lines of code?), perhaps it is OK to just have the logic here.

src/components/implementation/no_interface/scheddev/sl_lock.c
```diff
@@ -0,0 +1,110 @@
...
+}
+
+void sl_lock_lock_no_cs(struct sl_lock *lock) {
+    assert(lock);
+
+    // Lock internally, which should always succeed since timeout is 0
+    assert(sl_lock_lock_internal_no_cs(lock, 0) == 1);
```
gparmer: you cannot have calls to essential logic inside of `assert`ions.  Assertions are often removed from the generated object for final code (using a `#define`), so that would get rid of the logic you need.

src/components/implementation/no_interface/scheddev/sl_lock.c
```diff
@@ -0,0 +1,110 @@
+#include "sl_lock.h"
+
+#include "sl.h"
+
+void sl_lock_init(struct sl_lock *lock) {
+    assert(lock);
+
+    *lock = (struct sl_lock) {
+        .holder = 0
+    };
```
phanikishoreg: I'm not sure if there is a rule for this init style. I'd keep it consistent with the rest `sl` library struct init!

src/components/implementation/no_interface/scheddev/sl_lock.c
```diff
@@ -0,0 +1,110 @@
...
+
+    microsec_t start_time = sl_now_usec();
+    while (lock->holder != 0 && !sl_lock_timeout_elapsed(start_time, timeout)) {
+        thdid_t holder = lock->holder;
+
+        /*
+         * If we are preempted after the exit, and the holder is no longer holding
+         * the critical section, then we will yield to them and possibly waste a
+         * time-slice.  This will be fixed the next iteration, as we will see an
+         * updated value of the holder, but we essentially lose a timeslice in the
```
phanikishoreg: Perhaps a FIXME or a TODO in the comment will help if you say you'll fix in the next iteration.

src/components/implementation/no_interface/scheddev/sl_lock.c
phanikishoreg: Declarations should be first, then body and then `return`.

src/components/implementation/no_interface/scheddev/sl_lock.c
```diff
@@ -0,0 +1,110 @@
...
+
+    // Lock internally, which should always succeed since timeout is 0
+    assert(sl_lock_lock_internal_no_cs(lock, 0) == 1);
+}
+
+
+int sl_lock_timed_lock(struct sl_lock *lock, microsec_t timeout) {
+    assert(lock);
+
+    if (timeout == 0) {
```
gparmer: This check is redundant with the logic in `sl_lock_lock_internal`, I believe.

If you have this function, it should be:
```
int
sl_lock_timed_lock(struct sl_lock *lock, microsec_t timeout)
{ return sl_lock_lock_internal(lock, timeout); }
```
...which kinda just makes me wonder why we have this function in the first place (if my assertion above about redundancy is true).

src/components/implementation/no_interface/scheddev/sl_lock.c
```diff
@@ -0,0 +1,110 @@
...
+        return 0;
+    }
+    return sl_lock_lock_internal(lock, timeout);
+}
+
+int sl_lock_timed_lock_no_cs(struct sl_lock *lock, microsec_t timeout) {
```
gparmer: ditto.

src/components/implementation/no_interface/scheddev/sl_lock.c
phanikishoreg: don't mix declarations with body. I think you'll need to check for such in the rest.

src/components/implementation/no_interface/scheddev/sl_lock.c
```diff
@@ -0,0 +1,110 @@
...
+        lock->holder = sl_thd_curr_id();
+        took_lock = 1;
+    }
+    return took_lock;
+}
+
+
+int sl_lock_lock_internal(struct sl_lock *lock, microsec_t timeout) {
+    sl_cs_enter();
+    int result = sl_lock_lock_internal_no_cs(lock, timeout);
```
gparmer: define above...

src/components/implementation/no_interface/scheddev/sl_lock.c
```diff
@@ -0,0 +1,110 @@
...
+    if (timeout == 0) {
+        return 0;
+    }
+    return sl_lock_lock_internal(lock, timeout);
+}
+
+int sl_lock_timed_lock_no_cs(struct sl_lock *lock, microsec_t timeout) {
+    assert(lock);
+
+    if (timeout == 0) {
```
gparmer: ditto from above.

src/components/implementation/no_interface/scheddev/sl_lock.c
```diff
@@ -0,0 +1,110 @@
...
+        return 0;
+    }
+    return sl_lock_lock_internal_no_cs(lock, timeout);
+}
+
+
+void sl_lock_unlock(struct sl_lock *lock) {
+    assert(lock);
+    sl_cs_enter();
```
gparmer: earlier you used a different style, and I like it better (iwth the CS operations without new-line separators from the called function).  That makes the intent of this function much more clear.

src/components/implementation/no_interface/scheddev/sl_lock.c
phanikishoreg: `assert` only to validate. Don't call functions or do useful work in `assert`. `assert`s could be compiled out.

src/components/implementation/no_interface/scheddev/sl_lock.c
phanikishoreg: separate asserts and function calls.

src/components/implementation/no_interface/scheddev/sl_lock.c
```diff
@@ -0,0 +1,110 @@
...
+
+    sl_lock_unlock_no_cs(lock);
+
+    sl_cs_exit();
+}
+
+void sl_lock_unlock_no_cs(struct sl_lock *lock) {
```
gparmer: try to have called functions appear before those that call them.

src/components/implementation/no_interface/scheddev/sl_lock.c
```diff
@@ -0,0 +1,110 @@
...
+    // Lock internally, which should always succeed since timeout is 0
+    assert(sl_lock_lock_internal_no_cs(lock, 0) == 1);
+}
+
+
+int sl_lock_timed_lock(struct sl_lock *lock, microsec_t timeout) {
+    assert(lock);
+
+    if (timeout == 0) {
+        return 0;
```
phanikishoreg: Most POSIX API I've used actually do a `infinite` retry if `timeout` is `0`. Perhaps follow that convention?

src/components/implementation/no_interface/scheddev/sl_lock.c
```diff
@@ -0,0 +1,110 @@
...
+    return sl_lock_lock_internal_no_cs(lock, timeout);
+}
+
+
+void sl_lock_unlock(struct sl_lock *lock) {
+    assert(lock);
+    sl_cs_enter();
+
+    sl_lock_unlock_no_cs(lock);
+
```
phanikishoreg: blank-lines seem unnecessary here.

src/components/implementation/no_interface/scheddev/sl_lock.h
```diff
@@ -0,0 +1,24 @@
...
+
+void sl_lock_lock(struct sl_lock *lock);
+void sl_lock_lock_no_cs(struct sl_lock *lock);
+
+int sl_lock_timed_lock(struct sl_lock *lock, microsec_t timeout);
+int sl_lock_timed_lock_no_cs(struct sl_lock *lock, microsec_t timeout);
+
+void sl_lock_unlock(struct sl_lock *lock);
+void sl_lock_unlock_no_cs(struct sl_lock *lock);
+
```
phanikishoreg: are all API here exposed? if not, perhaps separating out would be better.

src/components/implementation/no_interface/scheddev/sl_lock.h
```diff
@@ -0,0 +1,24 @@
+#ifndef SL_LOCK_H
+#define SL_LOCK_H
+
+#include <cos_defkernel_api.h>
```
phanikishoreg: doesn't seem to be used here. including where it's used is sufficient.

src/components/implementation/no_interface/scheddev/sl_lock.c
gparmer: This is a great point.

Many APIs have a trylock which essentially does what the current implementation is doing:  take the lock if it is uncontended, and return otherwise with a notification that the lock wasn't taken.

So we have

- trylock
- lock with no qualifications
- lock, and contend for max some period of time

If these can be abstracted into a single *static inline* function, then we can likely generate efficient code for it.

src/components/implementation/no_interface/scheddev/sl_lock.h
gparmer: I'm a bit concerned about the no_cs functions here.  In the end, these APIs do end up taking/releasing the CS.  Will a client using this API expect that?  anyway, the main reason we added the no_cs functions to sl in the first place was for this API, so I don't see a reason the no_cs functions should be public at all.

Opinions?

src/components/implementation/no_interface/scheddev/sl_lock.c
Others: The return value indicates if the lock was successfully taken or not. Which is important if you're using timed_lock

src/components/implementation/no_interface/scheddev/sl_lock.c
Others: I have to handle the case where the timeout is 0, ergo we can wait forever. Breaking the functionality up makes it more clear what's going on imho. But if you think this should be changed, I can move it around?

src/components/implementation/no_interface/scheddev/sl_lock.c
Others: Nope, it's not redundant. Here if the timeout is zero, we don't wait at all. If the timeout passed to `sl_lock_lock_internal` is zero, we can wait forever. (I use it as a sentinel value.)

src/components/implementation/no_interface/scheddev/sl.h
Others: My idea was that users should remain on the same level of abstraction while using the library. Dipping into `cos_thdid()` feels like a pretty low level thing to do. Plus, this way you don't have to have a couple lines of safety checks after every call...

src/components/implementation/no_interface/scheddev/sl.h
Others: Uses the safety checks in `sl_thd_curr_id`

src/components/implementation/no_interface/scheddev/sl.h
Others: This here is just a reordering

src/components/implementation/no_interface/scheddev/sl_lock.c
Others: Doing infinite retry if timeout is 0 is exactly what `sl_lock_lock_internal` does.

src/components/implementation/no_interface/scheddev/sl_lock.h
Others: I need the `no_cs` functions for the cFE. So I don't really have a choice but to expose them :/

src/components/implementation/no_interface/scheddev/sl.h
Others: I'm going to leave this as is. If one of the asserts fails, it's useful to know which one.

Others:
Okay, I've fixed the above problems and made `sl_lock_lock` faster

Others:
@gparmer @phanikishoreg thoughts?

src/components/implementation/no_interface/scheddev/sl.c
```diff
@@ -104,11 +104,10 @@ sl_thd_wakeup(thdid_t tid)
 }
 
 void
-sl_thd_yield(thdid_t tid)
+sl_thd_yield_cs_exit(thdid_t tid)
```
phanikishoreg: Why do we need this API? 
I'm trying to understand what's the motivation for letting user of this API handle sched locking, and expect only to do unlocking here.

src/components/implementation/no_interface/scheddev/sl_lock.c
```diff
@@ -2,109 +2,67 @@
...
-         * on simplicity here.
-         */
-        sl_cs_exit();
-        sl_thd_yield(holder);
+    sl_cs_enter();
+    while (lock->holder != 0) {
+        sl_thd_yield_cs_exit(lock->holder);
         sl_cs_enter();
     }
-
```
phanikishoreg: Can't we simplify to? 
```
do {
     sl_cs_enter();
     if (lock->holder) sl_thd_yield_cs_exit(lock->holder);
     else sl_cs_exit();
} while (lock->holder);
```
I could be wrong with both but can't we also just do:
(with cs_enter() and exit() from within yield())
```
while (lock->holder) sl_thd_yield(lock->holder);
```

src/components/implementation/no_interface/scheddev/sl.c
Others: I need this functionality to avoid a race in sl_lock_lock, where a timeslice could sometimes be wasted.

src/components/implementation/no_interface/scheddev/sl.h
phanikishoreg: My bad!
I was mostly referring to the change in style/formatting.

src/components/implementation/no_interface/scheddev/sl_lock.c
Others: I like my current code better than the first one. The second version has a subtle race, where we get preempted after reading the holder, and the lock as released. Then we waste a timeslice yielding to no one.

src/components/implementation/no_interface/scheddev/sl_lock.c
```diff
@@ -2,109 +2,67 @@
...
-}
-
-
-int sl_lock_timed_lock(struct sl_lock *lock, microsec_t timeout) {
-    assert(lock);
-
-    if (timeout == 0) {
-        return 0;
+    sl_cs_enter();
+    cycles_t deadline = sl_now() + sl_usec2cyc(max_wait_time);
```
phanikishoreg: Is this absolute time from when the function was called? or is it relative to when you can successfully acquire the lock? I feel like it's the former, in which case, `deadline` calc should be before acquiring the lock.

Others: Right, of course. I'll fix that!

src/components/implementation/no_interface/scheddev/sl_lock.c
Others: This is actually fixed, the diff just hasn't updated correctly

src/components/implementation/no_interface/scheddev/sl_lock.c
phanikishoreg: Thank you. I see the race you're talking about in second one!

src/components/implementation/no_interface/scheddev/sl_lock.h
phanikishoreg: Wait, what's going on? You said you need `no_cs` for cFE, but you got rid of them, which is fine, but just curious about what's happening!

src/components/implementation/no_interface/scheddev/sl.h
gparmer: Noone would or should do those safety checks.  You have to design the system so that there are no stack overflows (which is how this value could be wrong).  If you want to check for stack overflows, then you need to do so using structured techniques (red zones), not hacks like this.

You can't add checks to all of your code for every piece of undefined behavior in C.  It isn't reasonable, nor possible.  So I don't agree with the justification about checks.  In reality, I doubt anyone will use this API, instead sticking with the consistent `cos_thdid`, but I'm not opposed to it. since it will get compiled away as dead code if it isn't used.

src/components/implementation/no_interface/scheddev/sl.h
```diff
@@ -73,11 +73,25 @@ sl_thd_setprio(struct sl_thd *t, tcap_prio_t p)
...
+	assert(tid != 0);
+	assert(tid < MAX_NUM_THREADS);
+	return tid;
+}
+
 
 static inline struct sl_thd *
 sl_thd_curr(void)
-{ return sl_thd_lkup(cos_thdid()); }
+{ return sl_thd_lkup(sl_thdid()); }
```
gparmer: Aren't the assertions better done in the backend?  Only there will it know how many thread contexts it actually has which may be less than `MAX_NUM_THREADS`?  Given how this is compiled (across .o files), we can't assume the compiler will remove redundant checks.

src/components/implementation/no_interface/scheddev/sl.h
```diff
@@ -73,11 +73,25 @@ sl_thd_setprio(struct sl_thd *t, tcap_prio_t p)
...
+	return sl_mod_thd_get(sl_thd_lookup_backend(tid));
+}
+
+static inline thdid_t
+sl_thdid(void)
+{
+	thdid_t tid = cos_thdid();
+	assert(tid != 0);
+	assert(tid < MAX_NUM_THREADS);
+	return tid;
```
gparmer: blank lines above `return` and after definitions.

src/components/implementation/no_interface/scheddev/sl.h
gparmer: You're saying that `sl_now_usec` existed before in the code?  Why is it not removed in another place in this commit?

src/components/implementation/no_interface/scheddev/sl_lock.h
```diff
@@ -0,0 +1,23 @@
...
+#define SL_LOCK_H
+
+#include <cos_kernel_api.h>
+
+struct sl_lock {
+    // will be 0 if no one holds the lock
+    volatile thdid_t holder;
+};
+
+#define SL_LOCK_STATIC_INIT() (struct sl_lock) { .holder = 0 }
```
gparmer: didn't fix style here.

src/components/implementation/no_interface/scheddev/sl_lock.c
```diff
@@ -0,0 +1,67 @@
...
+sl_lock_init(struct sl_lock *lock)
+{
+    assert(lock);
+    *lock = SL_LOCK_STATIC_INIT();
+}
+
+thdid_t
+sl_lock_holder(struct sl_lock *lock)
+{
+    assert(lock);
```
gparmer: I'm not sold on all of these `assert(lock)`s.  They only check if the client passes in `NULL`, which is a bug that would be near trivial to diagnose with `objdump`.  They have a very real cost (binary size and branch target buffer capacity), and don't provide a real benefit.

src/components/implementation/no_interface/scheddev/sl_lock.c
```diff
@@ -0,0 +1,67 @@
...
+    return lock->holder;
+}
+
+
+void
+sl_lock_lock(struct sl_lock *lock)
```
gparmer: I recommended changing the name.  You didn't respond, and just did another PR without acknowledging the suggestion.

So now I'm making it mandatory:  Change the name as described previously.

src/components/implementation/no_interface/scheddev/sl_lock.c
```diff
@@ -0,0 +1,67 @@
...
+    assert(lock);
+
+    sl_cs_enter();
+    while (lock->holder != 0) {
+        sl_thd_yield_cs_exit(lock->holder);
+        sl_cs_enter();
+    }
+    lock->holder = sl_thdid();
+    sl_cs_exit();
+}
```
gparmer: looks great!

src/components/implementation/no_interface/scheddev/sl_lock.c
gparmer: It looks like all the comments on the previous PR died?

`sl_lock_take` and `sl_lock_release`

src/components/implementation/no_interface/scheddev/sl_lock.c
```diff
@@ -0,0 +1,67 @@
...
+int
+sl_lock_timed_lock(struct sl_lock *lock, microsec_t max_wait_time)
+{
+    int result;
+    cycles_t deadline = sl_now() + sl_usec2cyc(max_wait_time);
+
+    assert(lock);
+
+    sl_cs_enter();
+    while (lock->holder != 0 && sl_now() < deadline) {
```
gparmer: Good job converting to cycles so that this loop is cheaper.

src/components/implementation/no_interface/scheddev/sl_lock.c
```diff
@@ -0,0 +1,67 @@
...
+        sl_thd_yield_cs_exit(lock->holder);
+        sl_cs_enter();
+    }
+
+    if (lock->holder == 0) {
+        lock->holder = sl_thdid();
+        result = 1;
+    }else {
```
gparmer: `}else` -> `} else`

src/components/implementation/no_interface/scheddev/sl_lock.c
```diff
@@ -0,0 +1,67 @@
...
+    return result;
+}
+
+void
+sl_lock_unlock(struct sl_lock *lock)
+{
+    assert(lock);
+
+    sl_cs_enter();
+    assert(lock->holder);
```
gparmer: this is redundant with the next check...  Only need the next line, not this one.

src/components/implementation/no_interface/scheddev/sl_lock.h
```diff
@@ -0,0 +1,23 @@
...
+
+void sl_lock_init(struct sl_lock *lock);
+
+thdid_t sl_lock_holder(struct sl_lock *lock);
+
+void sl_lock_lock(struct sl_lock *lock);
```
gparmer: These are common operations, and I think that we want to give the compiler the opportunity to optimize them.  I think `sl_lock_take` and `sl_lock_release` should be `static inline` in the header.  I see little to no downside to doing this, and it enables inlining.

src/components/implementation/no_interface/scheddev/sl.h
Others: This safety checks would've saved me a ridiculous amount of time working on the CFE. I know we can't guard against every problem, but this has a good chance of saving someone time in the future.

gparmer: "f you want to check for stack overflows, then you need to do so using structured techniques (red zones), not hacks like this."

gparmer: As I said, the function is fine, but I doubt it will be used often.

src/components/implementation/no_interface/scheddev/sl_lock.c
Others: I left it as two checks because they diagnose two different failure states. In the first state, the lock is not taken, whereas in the second someone else holds the lock. This has already helped me debug real code, so I'd like to leave it as-is if possible.

src/components/implementation/no_interface/scheddev/sl_lock.c
Others: Whoops!

src/components/implementation/no_interface/scheddev/sl_lock.c
Others: There are already 80 comments on this PR. I'm sorry I forgot to reply to the previous comment, and if it is mandatory for inclusion I will change the names of the functions.

src/components/implementation/no_interface/scheddev/sl_lock.c
Others: Shouldn't `assert` be compiled out in production code? I'm happy to remove these if that is not the case.

src/components/implementation/no_interface/scheddev/sl_lock.c
gparmer: Assertions aren't there to help you debug.  They are there to fail early when there is a problem and give you a starting point for debugging.  They can be completely disabled when building, which points out their position as not really being a debugging aid.  I see no reason to incur overhead that is strictly redundant that all execution on the system will pay, to save yourself one recompile to add a few checks to see why it failed if it did.  That is not a good trade-off.

src/components/implementation/no_interface/scheddev/sl.h
Others: I guess I made a mistake when replying to @phanikishoreg earlier, my apologies. This is just a small helper method I added, and is the only code change unrelated to locks in this commit. (Although it is/was used in the library I think.

src/components/implementation/no_interface/scheddev/sl.h
Others: This is the case where safety checks would've really helped me with the cFE, so I'm reluctant to change this.

src/components/implementation/no_interface/scheddev/sl_lock.c
Others: I'm confused by your point about building without assertions. If you would build without them, then aren't they only useful for debugging? I'm going to go ahead and remove this one though, you're right that it is redundant.

src/components/implementation/no_interface/scheddev/sl_lock.c
gparmer: It was hard to leave 80 comments.  It is hard to take the time to do that, and then have it take up *more* time the next time to send out a reminder.  Read the post on SWEng in research.  Even if code review comments are accidentally ignored (which is the case here), it is always hard for someone who puts in the time for the review to see it as anything other than you valuing your time other theirs.

Yes, code reviews suck for everyone.  But they are the only way to try and add some quality assurance and taste in the fundamental aspects of the system.  They're also a concrete way to do some training and to show more ways to think about code construction.  So it is important that they be taken seriously.

OK, I'm done ;-)

src/components/implementation/no_interface/scheddev/sl_lock.h
Others: Shoot, another instance I missed.

src/components/implementation/no_interface/scheddev/sl_lock.h
Others: 👍

PR #258, opened 2017-07-06
Wakeup threads from timeout module

phanikishoreg:
### Summary of this PR

* Fixed style in heap library and added `heap_init` for initializing stack allocated heap structs.
* Moved `fprr` into a module directory under `scheddev`. TODO: fix dependency warnings from composite compilation system. Perhaps having more intelligent logic to detect `no_interface` relative to `src` and not current directory. 
* added waking up of threads from the timeout module. timeout module now has a min-heap sorted based on wakeup times. Every `schedule` (`sl_cs_exit_schedule_nospin()` call) checks for threads that could have expired their timeouts and wakes them up. A unit test that wakes up high priority thread when it's timeout expires. (very basic test done for now, on qemu and only tested to see if high prio thread runs after it blocks itself, it does! TODO: thorough testing. ). 

**UPDATE**
- Reverted folder structure changes. Need to spend some time in doing it the right way. Perhaps how Gabe suggested, having it in `components/lib/sl/` and linking to each module from the user. Not sure if that will work, because of interdependencies between sl and modules. [Big TODO]
- Included other review fixes for comments from Gabe. 

**UPDATE - 2** (based on review comments and f2f discussion).
- Mainly added `sl_thd_block_timeout` that blocks for absolute timeout and returns number of cycles elapsed between timeout and actual wakeup, and `sl_thd_block_periodic` that blocks for until next period (which could be in the past, if not used correctly - care to be taken not to interleave `block_timeout` and `block_periodic` unless `timeout` is within the next period.) and returns number of periods elapsed between timeout and actual wakeup (+1). 
- Removed timeout module and included timeout api as part of sl library.
- Moved timeout specifics out of the scheduling policy module.

@gparmer @Others - please review. 

### Code Quality

Style:

Code Craftsmanship:


src/components/implementation/no_interface/scheddev/fprr/Makefile
```diff
@@ -0,0 +1,9 @@
+#C_OBJS=unit_schedlib.o
```
gparmer: I see.  You're trying to create a directory hierarchy in scheddev for the different scheduling modules.

This seems to be a "one-off" solution for the scheddev component, though.  If you're going so far as to break apart different modules, should we just do that in the `components/lib/` directory so that we can include them in each component separately?

Still might be missing something big here, so I'll read on.

src/components/implementation/no_interface/scheddev/fprr/sl_mod_fprr.c
```diff
@@ -35,6 +38,7 @@ void
 sl_mod_block(struct sl_thd_policy *t)
 {
 	ps_list_rem_d(t);
+	sl_timeout_mod_block(sl_mod_thd_get(t), 0);
```
gparmer: Is this just setting a timeout for this thread to "no timeout"?

This doesn't seem like the type of thing that we want in a scheduling module.  Shouldn't it be called for *every* scheduling module?  There will be one, sometime in the future, that forgets to call this...

src/components/implementation/no_interface/scheddev/fprr/sl_mod_fprr.c
```diff
@@ -70,10 +74,32 @@ sl_mod_thd_delete(struct sl_thd_policy *t)
...
+		ps_list_head_append_d(&threads[t->priority], t);
+		sl_thd_setprio(sl_mod_thd_get(t), t->priority);
+
+		break;
+	}
+	case SCHEDP_WINDOW:
+	{
+		struct sl_thd *td = sl_mod_thd_get(t);
+
+		assert(v >= SL_FPRR_US_MIN && v <= SL_FPRR_US_MAX);
```
gparmer: Why the limitation on `MAX`?

src/components/implementation/no_interface/scheddev/fprr/sl_thd.h
```diff
@@ -22,6 +22,9 @@ struct sl_thd {
 	thdcap_t       thdcap;
 	tcap_prio_t    prio;
 	struct sl_thd *dependency;
+
+	cycles_t       period, wakeup_cycs;
+	int            wakeup_idx;
```
gparmer: comment for the `wakeup_idx`?  Not sure what it is.  I assume some sort of heap index?

src/components/implementation/no_interface/scheddev/fprr/sl_timer_mod_periodic.c
```diff
@@ -0,0 +1,78 @@
+#include <sl.h>
+#include <sl_consts.h>
+#include <sl_plugins.h>
+#include <heap.h>
+
+#define SL_TIMEOUT_MOD_MAX_THDS MAX_NUM_THREADS
+
+struct wakeup_heap {
+	struct heap  h;
+	void        *data[SL_TIMEOUT_MOD_MAX_THDS];
```
gparmer: We likely want to replace all of this with rb-tree with nodes internal to the threads.  This heap is fast, but the static memory allocation is not great.

TODO: in the future.

src/components/implementation/no_interface/scheddev/fprr/sl_timer_mod_periodic.c
```diff
@@ -0,0 +1,78 @@
...
+	char         p; /* pad. TODO: use alignment */
```
gparmer: What is this?  why?

src/components/implementation/no_interface/scheddev/fprr/sl_timer_mod_periodic.c
```diff
@@ -0,0 +1,78 @@
...
+} wakeup_heap;
+
+static struct heap *wkh = (struct heap *)&wakeup_heap;
```
gparmer: Can't you just pass in `&wakeup_heap.heap` instead of this?

src/components/implementation/no_interface/scheddev/fprr/sl_timer_mod_periodic.c
```diff
@@ -0,0 +1,78 @@
...
+	} while (heap_size(wkh));
+}
+
+void
+sl_timeout_mod_block(struct sl_thd *t, cycles_t wakeup)
+{
+	assert(t && t->wakeup_idx == -1); /* not already in heap */
+	assert(heap_size(wkh) < SL_TIMEOUT_MOD_MAX_THDS);
+
+	if (!wakeup) t->wakeup_cycs += t->period; /* implicit wakeup = task period */
```
gparmer: No check if `t->period` has actually been set?

src/components/implementation/no_interface/scheddev/fprr/sl_timer_mod_periodic.c
```diff
@@ -0,0 +1,78 @@
...
+	else         t->wakeup_cycs  = wakeup;
+
+	heap_add(wkh, t);
+}
+
+void
+sl_timeout_mod_wakeup_expired(cycles_t now)
+{ __sl_timeout_mod_wakeup_expired(now); }
```
gparmer: If there's no reason to have the `__` version, get rid of it.

src/components/lib/heap.c
```diff
@@ -13,16 +13,18 @@
...
-//#define HEAP_TRACE_DEBUG
+#undef HEAP_TRACE_DEBUG
 #ifdef HEAP_TRACE_DEBUG
 #define debug(format, ...) printd(format, ## __VA_ARGS__)
 #else 
 #define debug(format, ...)
 #endif
 
-static inline void swap_entries(void *arr[], int a, int b, update_fn_t u)
+static inline void
```
gparmer: Thanks for fixing all of this old code!

gparmer:
Thanks for doing this @phanikishoreg!  It is almost there.  My main question is about how we want to structure the code.  Perhaps moving all of it into `components/lib/sl/` and then just adding `-lsl -lfprr -lthdstatic -lperiodic -lsl` to the makefiles for the compoents we want to use them from?  Unfortunately, the order of that list matters, and if there are circular dependencies between libraries, they will need to be included twice in the list.

src/components/implementation/no_interface/scheddev/fprr/Makefile
phanikishoreg: Yes, I can do that. I was thinking that it would be time consuming and will first push my current changes. I can undo the directory changes for now if this is not acceptable.

src/components/implementation/no_interface/scheddev/fprr/sl_mod_fprr.c
phanikishoreg: It is setting timeout to the next period, if second parameter is 0, it uses thread's period as implicit timeout. I should add a comment here, I'll. I have a comment in the `sl_timeout_mod_block` function but not here. 

I did this in the module because in my understanding module should have more information about the threads and have wakeup/deadline etc intelligence built into it and not the sl core library.

src/components/implementation/no_interface/scheddev/fprr/sl_mod_fprr.c
phanikishoreg: My bad, period doesn't need to have that check. It was mainly for budget as budget is `unsigned long`, I will get rid of this for Period.

src/components/implementation/no_interface/scheddev/fprr/sl_thd.h
phanikishoreg: Yes, it is heap index. I'll comments, thank you.

src/components/implementation/no_interface/scheddev/fprr/sl_timer_mod_periodic.c
phanikishoreg: Got it.

src/components/implementation/no_interface/scheddev/fprr/sl_timer_mod_periodic.c
phanikishoreg: This was from the example I picked up from another usage of heap library, I didn't change it. I added a TODO for using alignments. I'm stupid, trying to get changes in quick! 

This padding is not required at all here, I will remove it.

src/components/implementation/no_interface/scheddev/fprr/sl_timer_mod_periodic.c
phanikishoreg: Thank you for catching it. It is a merge mistake.

src/components/implementation/no_interface/scheddev/fprr/sl_timer_mod_periodic.c
phanikishoreg: okay, I'll get rid of it.

phanikishoreg:
Thank you @gparmer for the review. I'll fix, update PR with feedback commit and notify you. I'm not sure if restructuring code will consume more time, I'll it to issues for now?

src/components/implementation/no_interface/scheddev/fprr/sl_mod_fprr.c
gparmer: Do we always want to set the timer when we block a thread?   If so, I don't see why this would be in scheduling-specific policy code.

src/components/implementation/no_interface/scheddev/fprr/sl_timer_mod_periodic.c
gparmer: yes, `&wakeup_heap.h`...sorry I got that wrong.

If you want shorthand, use a macro like `SL_TIMER_HEAP()`.  Using another variable just uses memory and creates another structure that might get corrupted.

src/components/implementation/no_interface/scheddev/fprr/sl_mod_fprr.c
phanikishoreg: Yes, we do. (Except for AEPs as they're explicitly woken up using `asnd`s.). Your point is valid and I think I understand that now. This code should really be in `sl_thd_block()` and not here as we always use it's period as timeout and sl_thd has that information. Also, none of this is module specific, you're right. Thank you!

src/components/implementation/no_interface/scheddev/fprr/sl_timer_mod_periodic.c
phanikishoreg: Kk, I'll fix it. Thank you.

phanikishoreg:
@gparmer I've updated the PR with the fixes for your feedback. Thank you.

src/components/implementation/no_interface/scheddev/sl_mod_fprr.c
```diff
@@ -70,10 +72,32 @@ sl_mod_thd_delete(struct sl_thd_policy *t)
...
+	{
+		assert(v < SL_FPRR_NPRIOS);
+		ps_list_rem_d(t); 	/* if we're already on a list, and we're updating priority */
+		t->priority = v;
+		ps_list_head_append_d(&threads[t->priority], t);
+		sl_thd_setprio(sl_mod_thd_get(t), t->priority);
+
+		break;
+	}
+	case SCHEDP_WINDOW:
```
gparmer: This is broken as it is bunding the scheduling policy with the timeout policy.

However, I'm not convinced we'll *ever* use a different timeout module.  I'm not sure why I thought this in the initial design.  It seems like the timeout stuff should just be a core part of the scheduling library.  In that case, `SCHEDP_WINDOW` should likely be interpreted by the wakeup/timeout policy, perhaps in addition to the code here.

src/components/implementation/no_interface/scheddev/sl_plugins.h
```diff
@@ -48,4 +48,8 @@ void sl_mod_init(void);
 void sl_timeout_mod_expended(cycles_t now, cycles_t oldtimeout);
 void sl_timeout_mod_init(void);
 
+/* API for thread wakeup using timeout module */
```
gparmer: I just realized that you didn't add a new API for blocking on a timeout.  Am I correct in interpreting this as you not supporting oneshot timeout requests?  I believe this is what cFE will be doing.  I just assumed that there was a `sl_block_timeout(..., timeout)` where timeout is `0` if we want to use the periodic timer (that we will check is already set up), or `X` > 0 for a timeout at an absolute time `X`.

How does this differ from your code?

src/components/implementation/no_interface/scheddev/sl_timer_mod_periodic.c
```diff
@@ -1,6 +1,54 @@
...
+
+void
+sl_timeout_mod_block(struct sl_thd *t, cycles_t wakeup)
+{
+	assert(t && t->wakeup_idx == -1); /* not already in heap */
+	assert(heap_size(SL_TIMEOUT_HEAP()) < SL_TIMEOUT_MOD_MAX_THDS);
+
+	if (!wakeup) {
+		assert(t->period);
+		t->wakeup_cycs += t->period; /* implicit wakeup = task period */
```
gparmer: perhaps an assertion for overflow?

phanikishoreg: Yeah, I have a couple of `/*FIXME: wraparound*/`, I guess asserting here would help to address those comments eventually. I'll add that. Thanks.

src/components/implementation/no_interface/scheddev/sl_plugins.h
phanikishoreg: Yes, you're right. I do not have oneshot timeout requests exposed. 
Can I just extend `sl_thd_block()` to include a parameter for `timeout`? The Timeout API for oneshot `block` requests is `void sl_timeout_mod_block(struct sl_thd *t, cycles_t wakeup);` is already implemented. But I think I need to expose it through `sl_thd_XX` API.

src/components/implementation/no_interface/scheddev/sl_timer_mod_periodic.c
gparmer: yeah, I don't want anyone debugging a wraparound issue ;-)

src/components/implementation/no_interface/scheddev/sl_plugins.h
gparmer: blocking will be done very commonly, and very uncommonly will the timeouts be used.  No, I'd rather not make the common-case API more complex (the "dependency" argument is already confusing enough).  So `sl_thd_block_timeout(...)`, though verbose, is likely better.

src/components/implementation/no_interface/scheddev/sl_mod_fprr.c
gparmer: Is there any way that we can separate the scheduling policy from the timeout stuff, or are they by their nature intertwined?

phanikishoreg: I'm sorry I need to think about this a bit more. Only thing timeout specific is initialising `wakeup_cycs` of the thread and perhaps moving that to timeout module makes sense because timeout module can then do synchronize that for all threads.

src/components/implementation/no_interface/scheddev/sl_plugins.h
phanikishoreg: Gotcha.

phanikishoreg:
@gparmer I've updated the PR with the feedback from you based on the review comments and the discussion we had on friday. Please review. Thank you!

src/components/implementation/no_interface/scheddev/sl_mod_fprr.c
```diff
@@ -7,6 +7,8 @@
 #define SL_FPRR_HIGHEST 0
 #define SL_FPRR_LOWEST  (SL_FPRR_NPRIOS-1)
 
+#define SL_FPRR_US_MIN  1000
```
gparmer: I'm not sure what this is, but consider if it should be `#define SL_FPRR_US_MIN SL_PERIOD_US`...

src/components/implementation/no_interface/scheddev/sl.h
```diff
@@ -174,11 +175,177 @@ sl_cs_exit(void)
...
+		/* FIXME: logic for wraparound in current tsc */
+		if (tp->timeout_cycs > now) break;
+
+		th = heap_highest(SL_TIMEOUT_HEAP());
+		assert(th && th == tp);
+		th->timeout_idx = -1;
+
+		assert(th->wakeup_cycs == 0);
+		th->wakeup_cycs = now;
+		sl_thd_wakeup_no_cs(th);
```
gparmer: This means that timeouts will wakeup threads blocked on non-timeouts, right?  Didn't we decide that we didn't want this?

Perhaps a thread execution state for timeouts is getting too complicated.  Perhaps we just need a flag for if a thread is in the blocked state, and it is doing so for a timeout, or if it is woken up (before blocking, in woken) by a timer.  If you need to discuss this, lets chat.

src/components/implementation/no_interface/scheddev/sl.c
```diff
@@ -51,6 +51,28 @@ sl_cs_exit_contention(union sl_cs_intern *csi, union sl_cs_intern *cached, sched
...
+ *	    0 if it successfully blocked in this call.
+ */
+int
+sl_thd_block_no_cs(struct sl_thd *t, int is_timeout)
+{
+	assert(t);
+
+	if (unlikely(t->state == SL_THD_WOKEN)) {
+		t->state = SL_THD_RUNNABLE;
+		sl_cs_exit();
```
phanikishoreg: Yes, big mistake. I'll correct it. Thank you. I'll also add an assertion in `sl_cs_exit`.

src/components/implementation/no_interface/scheddev/sl.c
```diff
@@ -61,21 +83,66 @@ sl_thd_block(thdid_t tid)
 
 	sl_cs_enter();
 	t = sl_thd_curr();
-	if (unlikely(t->state == SL_THD_WOKEN)) {
-		t->state = SL_THD_RUNNABLE;
+	if (sl_thd_block_no_cs(t, 0)) {
```
phanikishoreg: Yes, `block_type` seems to be much better. My main concern is passing `thd_state` as a parameter to this function, it is kinda needing the caller to know the `thd_states`. Instead I chose this.

src/components/implementation/no_interface/scheddev/sl.c
```diff
@@ -61,40 +83,103 @@ sl_thd_block(thdid_t tid)
...
+	int is_timeout = abs_timeout ? 1 : 0;
+
+	/* TODO: dependencies not yet supported */
+	assert(!tid);
 
 	sl_cs_enter();
-	t = sl_thd_lkup(tid);
-	if (unlikely(!t)) goto done;
+	t = sl_thd_curr();
+	if (sl_thd_block_no_cs(t, is_timeout)) {
```
phanikishoreg: Thank you.

src/components/implementation/no_interface/scheddev/sl.c
```diff
@@ -61,40 +83,103 @@ sl_thd_block(thdid_t tid)
...
+		sl_cs_exit();
+		return 0;
+	}
+	if (is_timeout) sl_timeout_block(t, abs_timeout);
+	sl_cs_exit_schedule();
+
+	/* FIXME: possible race around these accesses */
+	if (t->wakeup_cycs > t->timeout_cycs) jitter = t->wakeup_cycs - t->timeout_cycs;
```
phanikishoreg: Yes, re-taking the CS will fix it. But wasn't sure if it's worth that extra overhead and if this could cause interference on other higher prio tasks.

src/components/implementation/no_interface/scheddev/sl.c
```diff
@@ -61,40 +83,103 @@ sl_thd_block(thdid_t tid)
...
+
+	return jitter;
+}
+
+unsigned
+sl_thd_block_periodic(thdid_t tid)
+{
+	unsigned jitter = 0;
```
phanikishoreg: Kk, will keep that in mind for style.

`cycles_t` as return type could be misleading and a misnomer as I'm returning `number of periods` elapsed in through this API, is also one of the reasons I introduced a separate API `sl_thd_block_periodic()`. If their return types can be same, I could use `sl_thd_block_timeout()` with `timeout == 0` to be periodic, which was my originally how I thought I could do it. Because I don't somehow like that I could block indefinitely in two different ways with the current API, one through `sl_thd_block()` and other through `sl_thd_block_timeout(...,0)`.

src/components/implementation/no_interface/scheddev/sl.c
```diff
@@ -61,40 +83,103 @@ sl_thd_block(thdid_t tid)
...
+	tcap_t         tcap;
+	tcap_prio_t    prio;
+	sl_thd_state   prev_state;
+
+	sl_cs_enter();
+	t = sl_thd_lkup(tid);
+	if (unlikely(!t)) goto done;
+
+	prev_state = t->state;
+	if(sl_thd_wakeup_no_cs(t)) goto done;
```
phanikishoreg: Thank you.

src/components/implementation/no_interface/scheddev/sl.c
```diff
@@ -61,40 +83,103 @@ sl_thd_block(thdid_t tid)
...
+	if (prev_state == SL_THD_BLOCKED_TIMEOUT) sl_timeout_remove(t);
```
phanikishoreg: I totally missed that, I like this suggestion. Will change it.

src/components/implementation/no_interface/scheddev/sl.h
```diff
@@ -174,11 +175,177 @@ sl_cs_exit(void)
...
+/* Timeout and wakeup functionality */
+#define SL_TIMEOUT_HEAP()       (&timeout_heap.h)
+
+struct timeout_heap {
+	struct heap  h;
+	void        *data[SL_MAX_NUM_THDS];
+} timeout_heap;
+
+/* wakeup any blocked threads! */
+static void
```
phanikishoreg: Got it, I'll move them to `.c`.

src/components/implementation/no_interface/scheddev/sl.h
```diff
@@ -174,11 +175,177 @@ sl_cs_exit(void)
...
+	if (!heap_size(SL_TIMEOUT_HEAP())) return;
+
+	do {
+		struct sl_thd *tp, *th;
+
+		tp = heap_peek(SL_TIMEOUT_HEAP());
+		assert(tp);
+
+		/* FIXME: logic for wraparound in current tsc */
+		if (tp->timeout_cycs > now) break;
```
phanikishoreg: Got it. Will modify to use `likely()`.

I didn't `inline` this mainly because I didn't see any optimizations that compiler could do with this loop (dynamic loop condition check), and also don't think it will `inline` this body for that loop.

src/components/implementation/no_interface/scheddev/sl.h
phanikishoreg: `sl_thd_block()` which is used for blocking forever (waiting for a wakeup event), does not add that thread to the timeout heap. Timeout heap only contains threads that have requested for a timeout wakeup. So it will not wakeup a thread that's blocked without timeout. 

I should however add an assert for checking the thread state is `BLOCKED_TIMEOUT`, that will also serve as documentation.

src/components/implementation/no_interface/scheddev/sl.h
```diff
@@ -174,11 +175,177 @@ sl_cs_exit(void)
...
+{
+	/* FIXME: logic for wraparound in either timeout_cycs */
+	return ((struct sl_thd *)a)->timeout_cycs <= ((struct sl_thd *)b)->timeout_cycs;
+}
+
+static inline void
+__sl_timeout_update_idx(void *e, int pos)
+{ ((struct sl_thd *)e)->timeout_idx = pos; }
+
+static inline void
```
phanikishoreg: I understand what you mean here: `init` is not called often and it's not on the `fast-path`, so I should move it to `.c` and omit `inline`. Perhaps GCC will be smart enough to inline as part of it's optimizations even otherwise. 

Will change it. Thank you.

src/components/implementation/no_interface/scheddev/sl_mod_fprr.c
phanikishoreg: Kk. I'll make the names explicit here and for lets say `SL_FPRR_PERIOD_US_MIN`, I could use `SL_PERIOD_US`. 
I'm also going to change the PRIO related defines to `SL_FPRR_PRIO_HIGHEST` and `SL_FPRR_PRIO_LOWEST` for better clarity.

src/components/implementation/no_interface/scheddev/sl.h
phanikishoreg: Based on your suggestion below, it should leave this in `.h` by making it `inline`. Looks like `init` is the only inappropriate one to be in `.h`, will change that.

src/components/implementation/no_interface/scheddev/sl.h
phanikishoreg: Also I don't think that the TIMEOUT state is complicated and the changes are local to `sl_thd_block/wakeup`, so I'd still go with current implementation. 

I also don't want to say NO to the flag idea especially if you think we might extend this BLOCKED state further to other types which at this point I don't have a clue of. 
Like `BLOCKED | TIMEOUT`, if we were to have `BLOCKED | SOMETHINGELSE`, I don't forsee that.

Others:
@phanikishoreg The merging of my PR is what caused the merge conflict here. I looked at it before getting my code merged, and it looks like an easy fix. Basically I reordered some functions around code you deleted, so it's basically just fixing the ordering again.

src/components/implementation/no_interface/scheddev/sl.h
```diff
@@ -174,10 +193,163 @@ sl_cs_exit(void)
...
+ * scheduler chooses to run this thread, we will run the dependency
+ * instead (note that "dependency" is transitive).
+ */
+void sl_thd_block(thdid_t tid);
+/*
+ * if abs_timeout == 0, block forever = sl_thd_block()
+ *
+ * @returns: 0 if the thread is woken up by external events before timeout.
+ *	     +ve - number of cycles elapsed from abs_timeout before the thread
+ *		   was woken up by Timeout module.
```
Others: I think the documentation here could talk a little about the second argument, noting that it's an absolute timeout. It's obvious from the function signature, but I think it'd be nice to note.

phanikishoreg:
@gparmer I've updated with new commits for fixing most issues, except for critical-section around return value computation in sl_thd_block_timeout/sl_thd_block_periodic API. 

@Others Wierd, I got an email notification but I'm not able to find your comment in this PR. Reply: Yes, I've taken care of the merge conflict, which is like you said reordering some declarations in `sl.h`. Thank you!

src/components/implementation/no_interface/scheddev/sl.h
```diff
@@ -174,10 +193,163 @@ sl_cs_exit(void)
...
+ */
+cycles_t sl_thd_block_timeout(thdid_t tid, cycles_t abs_timeout);
+/*
+ * blocks for on periodic wakeup based on task period.
```
Others: I think this comment should be rewritten. It doesn't really explain the semantics of the function very well, and I had to look at the `@returns` annotation bellow + the code  to understand what this function does.

src/components/implementation/no_interface/scheddev/sl.h
```diff
@@ -174,10 +193,163 @@ sl_cs_exit(void)
...
+ *
+ * @returns: 0 if the thread is woken up by external events before timeout.
+ *           +ve - number of periods elapsed. (1 if it wokeup exactly at timeout = next period)
+ */
+unsigned int sl_thd_block_periodic(thdid_t tid);
+int  sl_thd_block_no_cs(struct sl_thd *t, sl_thd_state block_type);
+/* wakeup a thread that has (or soon will) block */
+void sl_thd_wakeup(thdid_t tid);
+int  sl_thd_wakeup_no_cs(struct sl_thd *t);
+void sl_thd_yield(thdid_t tid);
```
Others: IMHO there should be a space between `sl_thd_wakeup_no_cs` and `sl_thd_yield`

Others:
@phanikishoreg I deleted my comment because you pushed a merge commit while I was writing it :P Sorry for any confusion.

src/components/implementation/no_interface/scheddev/sl.h
phanikishoreg: Agreed, will rephrase that!

src/components/implementation/no_interface/scheddev/sl.h
phanikishoreg: Hmm, block/wakeup/yield are actions on threads and we've had them grouped until now. 
But yeah, now that we have so many versions of each, we might as well have space between each action. Will do!

src/components/implementation/no_interface/scheddev/sl.h
phanikishoreg: I'll do that. :)

phanikishoreg:
@Others Got it. I'm mad at github, they could have notified me of a delete! (Not that I like them spamming my inbox, but they all get grouped by gmail anyway :P). Haha.

src/components/implementation/no_interface/scheddev/sl.c
gparmer: OK, I was worried about incorrectness here.  The only data-structure we're accessing is the current thread, and if we're executing here, that thread structure should be valid.  Additionally, since we're only *accessing* memory (not changing it), it seems we're likely to avoid races.

Should be OK, but the analysis to determine that should be taken and is pretty nuanced.

src/components/implementation/no_interface/scheddev/sl.c
gparmer: Ahhh, I see.  I missed that you were returning the number of periods.

In that case, it should be an `unsigned int`, right?

src/components/implementation/no_interface/scheddev/sl.h
gparmer: I don't think that most of the functions are on the fast paths.  It seems like the only one that is called very frequently is the expired function.  Please think about the logic here.

Most of the sl functions are *not* inlined.  The only place that most of these timeout functions are used is in those functions.  So why would they be here and visible to the user?  If the sl functions are in a .c file, then putting these functions here as opposed to in the .c file (assuming they are only used there) has no impact on optimization at all.  So we're just exposing functionality and APIs to clients that should be hidden for no benefit.

What is the justification for any of the main timeout functions (from `sl_timeout_wakeup_expired` all the way down to `sl_timeout_init`) having bodies in the .h file?

src/components/implementation/no_interface/scheddev/sl.h
gparmer: If a thread has set up a periodic timer, how will the timer not wake it up when it is blocked on a mutex?  It doesn't seem to matter at all to me if `sl_thd_block` puts a thread on the timer queue if we support an API that always keeps it there.  What am I missing?

gparmer:
Added another much shorter round of feedback, but I don't see how to avoid more changes.

src/components/implementation/no_interface/scheddev/sl.h
phanikishoreg: If a thread wants to be blocked on a mutex, I think it would either use `sl_thd_block()` (if it wants to block indefinitely) or `sl_thd_block_timeout()` (if it wants to block until a timeout or if timeout == 0, it blocks indefinitely). 

I don't understand why a thread will use periodic timer when it wants to block on a mutex. Also, if it did use it, I'm thinking it is intending to wakeup when it period arrives even when it's waiting for mutex. It can then use the return value to find out if it was woken up by period expiry or by an external event and retry mutex-take. I thought we decided on this.

src/components/implementation/no_interface/scheddev/sl.h
phanikishoreg: Kk. I misunderstood, `static inline` or not, these functions are basically private to sl library now. So I should have put them in `sl.c`.

src/components/implementation/no_interface/scheddev/sl.c
phanikishoreg: Kk, I think this is actually safe. But I'll think about this again and get back.

src/components/implementation/no_interface/scheddev/sl.c
phanikishoreg: Yes, fixed that.

src/components/implementation/no_interface/scheddev/sl.h
Others: The current `sl_lock` mutex implementation uses explicit yields, not blocking. (This is for priority inheritance + ease of implementation.) I don't understand how/why you'd use the blocking mechanism for mutexes.

phanikishoreg: directed yields are not really doing priority inheritance but yes I think you're right in terms of ease of implementation. The way i'd think of implementing mutexes using locks: If a thread tries to take mutex that's already taken, we'd put that as `contender` in the lock data structure and block it either indefinitely or until a timeout (timed_mutex) based on caller request. And the thread that has successfully taken the mutex would wakeup the `contender` when it does a `mutex_release`. 

Yielding to a thread for mutex release may not be good idea for many reasons (especially in terms of number of context switches). 
One reason I said it's not priority inheritance in directed-yielding is because, we're not changing the priority of the thread that we'd run on yield. Imagine that the lock is taken by a low-prio thread and a high-prio thread is now trying to take it but will yield to that low-prio thread for releasing the lock. But in the meantime if there was a mid-prio interrupt, 1. it can execute because the current thread execution still has low-prio 2. when mid-prio thread is done with it's work and blocks, scheduler would again run the high-prio thread because it's still in the run-queue (for doing a yield). This is unnecessary overhead and could be a potential livelock.

src/components/implementation/no_interface/scheddev/sl.c
phanikishoreg: Yes, I think this computation is totally safe. 
A thread has to wakeup and resume execution at this point before it can call `block_timeout` or `block_periodic` api again, so neither of `timeout_cycs` or `wakeup_cycs` of that thread will be overwritten in the meantime, so there is absolutely no race-condition possibility.

src/components/implementation/no_interface/scheddev/sl.h
gparmer: Yes, the directed yields that @Others is doing *are* implementing PI for mutexes.  The way that they are not PI is only WRT tcaps (which is an issue).

Generally, using this mechanism causes many context switches as we go along a yield chain to the thread that holds the lock.  In the worst case, this overhead can mount (as @phanikishoreg pointed out).

Thus, you'd typically want to use the blocking support with a dependency (not supported currently) which makes the scheduler aware of the dependency.

So that's why 1. the directed yields are PI, and 2. why you don't want to do that long term.

src/components/implementation/no_interface/scheddev/sl.c
gparmer: Is "sporadically returning negative jitter values" totally safe?

src/components/implementation/no_interface/scheddev/sl.h
gparmer: @phanikishoreg we should chat about the timeout + block interactions.  slack/github isn't going to solve this.

src/components/implementation/no_interface/scheddev/sl.c
phanikishoreg: No. why would there be "negative jitter values"? In my opinion, there is no race condition possible and `timeout_cycs` and `wakeup_cycs` reflect on the last `block_timeout` or `block_periodic` call values, so they will be valid.

phanikishoreg:
Updated with new commit for feedback changes.

PR #262, opened 2017-07-19
Critical-section in `sl_thd_alloc` and `sl_thd_free`.

phanikishoreg:

* cFE unit-tests have thread deallocation tests and found out that thread was being woken up from timeout-queue when it is being deleted and also more subtle issues on deleted threads. To avoid races, and to cleanup the thread (except the actual system resource) use critical-section and delete it from timeout queue and scheduling policy module. This is what @Others and I've debugged recently.
* It would be good if `sl_thd_alloc` is also atomic w.r.t sl library, so that there will not be races in backend allocations when users dynamically allocate more threads. Added that.
* removed assert when there is nothing to schedule from `sl_mod_schedule`.
* also added basic unit-tests for `sl_thd_free`. (idle thread runs after all test threads are deleted in the system.)

**Mature and almost ready to be pulled but need to discuss on the fundamental idea about a thread freeing itself. In my opinion, a thread should not be allowed to run after it deletes itself. But I see two things that makes me feel that I could be wrong: 1. an unused `SL_THD_DYING` state. 2. A comment in `sl_thd_free: /* TODO: add logic for the graveyard to delay this deallocation if t == current */`, that I got rid of!**

Review: @gparmer @Others 

### Intent for your PR

Choose one (Mandatory):

### Code Quality

### Testing

- unit_schedtest with tests added for `sl_thd_free`.

### DIFF
--- a/src/components/implementation/tests/unit_schedtests/unit_schedlib.c
+++ b/src/components/implementation/tests/unit_schedtests/unit_schedlib.c
@@ -84,11 +84,14 @@ test_blocking_directed_yield(void)
 	sl_thd_param_set(high, sph.v);
 }
 
+#define TEST_ITERS 1000
+
 void
 test_high_wakeup(void *data)
 {
-	unsigned toggle  = 0;
-	struct sl_thd *t = data;
+	unsigned int toggle = 0, iters = 0;
+	struct sl_thd *t    = data;
+	cycles_t start      = sl_now();
 
 	while (1) {
 		cycles_t timeout = sl_now() + sl_usec2cyc(100);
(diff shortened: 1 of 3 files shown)


gparmer:
Somehow missed this PR.

You cannot call the backend free function while executing in the thread.  The thd_delete function will likely deallocate the thread (i.e. its kernel memory), so execution would *not* proceed after that.  the backend free function should free the thread's resources including its structure and stack, so we have to ensure that is done.  However, we can't do it in the thread that is *executing using that stack*.

Thus we likely need a linked list of threads pending shutdown that are not in the runqueue, and a single reaper thread that is unblocked when a thread enters the linked list, and it is in charge of doing all of these deallocations.

phanikishoreg:
Thank you for clarifying that, I now understand `/* TODO: add logic for the graveyard to delay this deallocation if t == current */` comment very clearly and it makes sense that `sl_thd_free` might actually set thread state to `SL_THD_DYING` and the reaper thread will eventually make it `SL_THD_FREE` so the thread slot in `sl` is actually free and available for reuse. 
I will keep that comment until we have that reaper functionality.

phanikishoreg:
If there are no additional comments in this PR review, I'll close this PR and make a new one with the same changes (+ that comment) and with the `Intent of the PR` to integrate changes.

PR #264, opened 2017-07-25
Critical-section in `sl_thd_alloc` and `sl_thd_free`

phanikishoreg:

* cFE unit-tests have thread deallocation tests and found out that thread was being woken up from timeout-queue when it is being deleted and also more subtle issues on deleted threads. To avoid races, and to cleanup the thread (except the actual system resource) use critical-section and delete it from timeout queue and scheduling policy module. This is what @Others and I've debugged recently.
* It would be good if `sl_thd_alloc` is also atomic w.r.t sl library, so that there will not be races in backend allocations when users dynamically allocate more threads. Added that.
* removed assert when there is nothing to schedule from `sl_mod_schedule`.
* also added basic unit-tests for `sl_thd_free`. (idle thread runs after all test threads are deleted in the system.)

**Reference: PR #262**

### Intent for your PR

Choose one (Mandatory):

### Reviewers (Mandatory):
@gparmer @Others 

### Code Quality

### Testing

- unit_schedtest with tests added for `sl_thd_free`.

### DIFF
--- a/src/components/implementation/tests/unit_schedtests/unit_schedlib.c
+++ b/src/components/implementation/tests/unit_schedtests/unit_schedlib.c
@@ -84,11 +84,14 @@ test_blocking_directed_yield(void)
 	sl_thd_param_set(high, sph.v);
 }
 
+#define TEST_ITERS 1000
+
 void
 test_high_wakeup(void *data)
 {
-	unsigned toggle  = 0;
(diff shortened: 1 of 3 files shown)


gparmer:
To be clear, the `TODO` still applies, and we need to fix that.

PR #265, opened 2017-07-26
Added C++ support

vnitu02:

- Added a posix abstraction layer over COS
- Enabled the `malloc`/`free` real implementation in musl
- Fixed the `vas_frontier` bug in `cos_pgtbl_intern_expandwith()`
- Added the C++ library

### Intent for your PR

Choose one (Mandatory):

### Reviewers (Mandatory):
@gparmer @ryuxin @phanikishoreg 

### Code Quality

### Testing

-unit_cxx with tests for C++ library


src/components/Makefile.comp
```diff
@@ -35,6 +39,7 @@ SHARED_FLAGS=-fno-merge-constants -nostdinc -nostdlib -fno-pic
 OPT= -g
 #OPT= -O3
 CFLAGS=-m32 -D__x86__ -Wall -Wextra -Wno-unused-parameter -Wno-unused-function -fno-stack-protector -fno-omit-frame-pointer -Wno-unused-variable -fvar-tracking $(INC_PATH) $(MUSLINC) $(LWIPINC) $(LUAINC) $(OPT) $(SHARED_FLAGS)
+CXXFLAGS=-fno-exceptions -fno-threadsafe-statics -Wno-write-strings -fpermissive $(CFLAGS)
```
Others: Why do we need `-fpermissive` in this Makefile? Isn't this preventing any component from getting the full range of errors?

src/components/implementation/Makefile.subsubdir
```diff
@@ -80,7 +82,11 @@ $(COMPONENT): $(OBJS)
 %.o:%.c
 #	$(info |     [CC]   $<: Compiling)
 	@$(CC) $(INCLUDE) $(CFLAGS) -o $@ -c $<
-
+       
+%.o:%.cc
+#	$(info |     [CC]   $<: Compiling)
```
Others: It might be nice to note in the message that it's C++ compiling here. I'm not sure that's really necessary though.

src/components/implementation/tests/unit_cxx/Makefile
```diff
@@ -0,0 +1,18 @@
+C_OBJS=cxx.o
+COMPONENT=cxx_test.o
+INTERFACES=
+DEPENDENCIES=cbuf_mgr printc sched valloc mem_mgr_large torrent evt timed_blk lock
```
Others: Are these dependencies really necessary?

src/components/implementation/tests/unit_cxx/Makefile
```diff
@@ -0,0 +1,18 @@
...
+IF_LIB=./libstdc++.a ./libsupc++.a 
+OBJLIBS += $(POSIX_LIB)
+ADDITIONAL_LIBS=-lcobj_format -lcos_kernel_api
+
+pre: first all
+first:
+	$(info |     [CP]   copying libsupc++.a from libcxx)
+	cp $(LIBSUPCXXOBJ) .
+	$(info |     [CP]   copying libstdc++.a from libcxx)
+	cp $(LIBSTDCXXOBJ) .
```
Others: Is this really the best way to do this? Shouldn't this leverage the existing library system and be in ADDITIONAL_LIBS?

src/components/implementation/tests/unit_cxx/cxx.cc
```diff
@@ -0,0 +1,154 @@
...
+	new_delete_test();
+	polymorphism_test();
+	rtti_test();
+	vector_test();
+	string_test();
+	cout<<"==========CXX TEST DONE========="<<endl;
+}
+
+
+
```
Others: I think there is some style cleanup that could be done here. There some extra newlines at the end of the file, and some operators are missing spacing in the code.

src/platform/linker/cl_macros.h
```diff
@@ -35,7 +35,7 @@
 
 #define UNDEF_SYMB_TYPE 0x1
 #define EXPORTED_SYMB_TYPE 0x2
-#define MAX_SYMBOLS 1024
+#define MAX_SYMBOLS 1024*64
```
Others: I think it's probably good style to put parens around this like:
`#define MAX_SYMBOLS (1024 * 64)`
I don't think it really matters here, but it's better to be obviously hygienic.

src/components/lib/posix/posix.c
```diff
@@ -1,26 +1,31 @@
...
 #define SYSCALLS_NUM 378
 
-extern void *do_mmap(size_t);
-extern int do_munmap(void*, size_t);
+extern struct cos_compinfo parent_cinfo;
 
 typedef long (*cos_syscall_t)(long a, long b, long c, long d, long e, long f);
 cos_syscall_t cos_syscalls[SYSCALLS_NUM];
 
+typedef enum { false, true } bool;
```
Others: If you're going to do this, why not just use `stdbool.h`?

src/components/lib/posix/posix.c
```diff
@@ -34,46 +39,29 @@ libc_syscall_override(cos_syscall_t fn, int syscall_num)
...
-	td = tsplit(cos_spd_id(), td_root, (char *)pathname, strlen(pathname), TOR_ALL, evt);
-
-	if (td <= 0) {
-		printc("open() failed!\n");
-		assert(0);
-	}
-
-	return td + 3; /* fd number adjust for reserved fd 0, 1 and 2 */
+	printc("open not implemented\n");
+	return 0;
```
Others: Should these "not implemented" methods fault? It feels like you want to fail fast if these syscalls aren't implemented.

src/components/lib/posix/posix.c
```diff
@@ -134,9 +126,14 @@ cos_mmap(void *addr, size_t length, int prot, int flags, int fd, off_t offset)
...
-	void *ret = do_mmap(length);
+	addr = (void *)cos_page_bump_allocn(&parent_cinfo, length);
+	if (!addr){
+		ret = (void *) -1;
+	} else {
+		ret = addr;
+	}
 
-	if (ret == (void *)-1) { /* return value comes from man page */
+	if (ret == (void *)-1) {  //return value comes from man page
```
Others: cos style requires `/* */` comments I think

src/components/lib/posix/posix.c
```diff
@@ -228,8 +182,9 @@ CCTOR static void
 posix_init(void)
 {
 	int i;
+	printc("posix init\n");
```
Others: Is this a debugging printc, or do we really want this?

gparmer:
@ryuxin since some of this code was initially his.

src/components/implementation/Makefile.subsubdir
gparmer: I agree.  `[CXX]` would be better (might help debug in the future).

src/components/implementation/tests/unit_cxx/Makefile
gparmer: These are a relic of the previous implementation.  We likely don't want them anymore.

src/components/implementation/tests/unit_cxx/Makefile
```diff
@@ -0,0 +1,18 @@
+C_OBJS=cxx.o
+COMPONENT=cxx_test.o
+INTERFACES=
+DEPENDENCIES=cbuf_mgr printc sched valloc mem_mgr_large torrent evt timed_blk lock
+IF_LIB=./libstdc++.a ./libsupc++.a
```
gparmer: I wonder if we should just auto-include these when building C++?  this will be easy to forget, and will give opaque error messages if you don't include them.

src/components/implementation/tests/unit_cxx/cxx.cc
gparmer: I know you got htis code from @ryuxin, but we do some a little consistency.  The main thing (as @Others pointed out) is that binary operators (`<<`, `+`, etc) should be surrounded by spaces.

src/components/implementation/tests/unit_cxx/cxx.h
```diff
@@ -0,0 +1,175 @@
...
+#endif
+
+class Global_foo {
+public:
+	Global_foo();
+	~Global_foo();
+};
+Global_foo::Global_foo()
+{
+//	cout<<"Global foo constructor"<<endl;
```
gparmer: remove transient comments.

src/components/implementation/tests/unit_cxx/cxx.h
```diff
@@ -0,0 +1,175 @@
...
+}
+Global_foo::~Global_foo()
+{
+	cout<<"Global foo destructor"<<endl;
+}
+
+class Global_bar {
+public:
+	Global_bar();
+//	~Global_bar();
```
gparmer: ditto.  Below as well.

src/components/implementation/tests/unit_cxx/cxx.h
```diff
@@ -0,0 +1,175 @@
...
+cl::cl(int a)
+{
+	i = a;
+	cout<<"create object "<<i<<endl;
+}
+cl::~cl()
+{
+	cout<<"destroy object "<<i<<endl;
+}
+int cl::get_i()
```
gparmer: function return type should be on the previous line.  We likely want blank lines between methods.

src/components/implementation/tests/unit_cxx/init.c
```diff
@@ -0,0 +1,12 @@
+#include <cos_component.h>
+#include <cobj_format.h>
+#include <cos_kernel_api.h>
+#include "init.h"
+
+struct cos_compinfo parent_cinfo;
+
+void init_cinfo(void){
```
gparmer: This is interesting.  I suppose we really can have a default initialization for all of this.  This should likely be in the `cos_defkernel_api` so that we can use it elsewhere.  We *might* already have it there!  Please check.  If not, we'll want two functions.  One for the "current" component's compinfo, and one for a "child" components.  Either feel free to add this, or add an issue in the issue tracker to "Create default cos_compinfo initialization functions".

src/components/lib/cos_kernel_api.c
```diff
@@ -359,6 +359,7 @@ cos_pgtbl_intern_expandwith(struct cos_compinfo *ci, pgtblcap_t intern, vaddr_t
 	if (ci->vasrange_frontier != round_to_pgd_page(mem)) return -1;
 
 	if ((unsigned long)ps_faa(&ci->vasrange_frontier, PGD_RANGE) > round_to_pgd_page(mem)) return -1;
+	if ((unsigned long)ps_faa(&ci->vas_frontier, PGD_RANGE) > round_to_pgd_page(mem)) return -1;
```
gparmer: :+1:

src/platform/linker/cl_macros.h
gparmer: yes, this is a defense programming way to prevent variable capture by the surrounding environment.

src/components/lib/posix/posix.c
gparmer: `return -ENOTSUP;`.

src/components/lib/posix/posix.c
```diff
@@ -125,6 +114,9 @@ cos_brk(void *addr)
 void *
 cos_mmap(void *addr, size_t length, int prot, int flags, int fd, off_t offset)
 {
+	void *ret=0;
+
+	printc("mmap\n");
```
gparmer: Remove debugging printouts from commits.

src/components/lib/posix/posix.c
```diff
@@ -125,6 +114,9 @@ cos_brk(void *addr)
 void *
 cos_mmap(void *addr, size_t length, int prot, int flags, int fd, off_t offset)
 {
+	void *ret=0;
```
gparmer: spaces around binary operators (` = `).

src/components/lib/posix/posix.c
gparmer: It seems odd to have a condition previously that sets the ret to `-1`, then have another here to check it.  why not just include this code above in the `if (!addr){` block?

src/components/lib/posix/posix.c
```diff
@@ -134,9 +126,14 @@ cos_mmap(void *addr, size_t length, int prot, int flags, int fd, off_t offset)
 		assert(0);
 	}
 
-	void *ret = do_mmap(length);
+	addr = (void *)cos_page_bump_allocn(&parent_cinfo, length);
+	if (!addr){
```
gparmer: space required between `){`.

src/components/lib/posix/posix.c
```diff
@@ -134,9 +126,14 @@ cos_mmap(void *addr, size_t length, int prot, int flags, int fd, off_t offset)
...
+		ret = (void *) -1;
```
gparmer: No space after unary operator (should be `(void *)-1`, as you have later...)

src/components/lib/posix/posix.c
```diff
@@ -228,8 +182,9 @@ CCTOR static void
 posix_init(void)
 {
 	int i;
+	printc("posix init\n");
 	for (i = 0; i < SYSCALLS_NUM; i++) {
-		cos_syscalls[i] = (cos_syscall_t)default_syscall;
+		cos_syscalls[i] = 0/*(cos_syscall_t)default_syscall*/;
```
gparmer: remove stale comments...

gparmer:
I can't see what the modification to the `ps` library is, but it says that there is a one-line change.  It is unlikely that is desirable.

gparmer:
I'm closing this for now so that your next PR will be a clean submit.  Thanks @Others for the thorough comments here.

@RobertGiff you're going to be looking into `musl` integration with rumpkernels, so it might be useful to see how the posix library works here, and how to define new "system calls".

phanikishoreg:
We'll probably need a section in CSG for C++ coding style if we see ourselves writing C++ code in Composite in the future. ;)

PR #268, opened 2017-08-09
SL with new thread types, vkernel with sl scheduling

phanikishoreg:

kernel:
* Avoid redundant scheduler notifications on `asnd`s.
* mechanism to timeouts to parent through a `cos_sched_rcv/cos_rcv` from a child thread.

`sl` library:
* Extended thread types in SL: THD, AEP, AEP w TCAP, COMP, COMP w TCAP
* Simple deferrable server in `sl` core
* Fixes to handle possible redundant BLOCKED events. 

vkernel:
* Changed to use `cos_defkernel_api` where necessary and sl library for scheduling.
* use `sinv` to parent to exit a child component or to get child component VMID

This seems like a useful push for many people to work with different types of threads and hierarchical scheduling, and vkernel as a simple example of using parent scheduling through `sl` library.

**Major TODO:** `sched_tok_t` support in `cos_asnd` and `cos_tcap_delegate`. Right now, I've made it simple to use `cos_tcap_transfer` + `cos_asnd` also to avoid races in excess budget transfers, so it must be easy to start with supporting `sched_tok_t` in `cos_asnd`. 

**UPDATE Aug12**:
- Scrapped THREAD TYPES, instead using just flags for `OWNTC` or `SEND`. 
- TODO: If OWNTC is out of budget and `asnd` or `switch` returns `-EPERM`, how do we handle it? Perhaps block the thread until it's next replenishment period! 

### Intent for your PR

Choose one (Mandatory):

### Reviewers (Mandatory):
@gparmer @Others 
@hungry-foolish - Gabe mentioned you'll need hierarchical scheduling for your paper, so I wanted to push this to mainline as soon as I've it to make it easier for you to merge and work with.
@vnitu02 - I'm thinking you'll also need extended thread types for scheduling between NFs and for interrupts etc.

### Code Quality

### Testing

- vkernel, micro booter with extended tests for cos_rcv with timeout
- unit_schedtests, unit_defci, unit_cxx, llbooter_test - Mainly to see if they complete.


src/kernel/capinv.c
```diff
@@ -478,14 +478,25 @@ cap_thd_switch(struct pt_regs *regs, struct thread *curr, struct thread *next,
...
 	curr_notif = arcv_notif = arcv_thd_notif(prev_notif);
+
+	/*
+	 * Enqueue send sched event only if it's the first one after the thread
+	 * state changed to THD_STATE_RCVING.
+	 * This is to avoid duplicate wakeup events at user-level scheduling.
+	 * Basically changing to Edge-triggered wakeup notification from 
+	 * some form of level-triggered.
+	 */ 
+	if (send && (!(rcv_thd->state & THD_STATE_RCVING) ||
```
gparmer: :+1:  great job with this.

src/components/implementation/no_interface/vkernel/Makefile
```diff
@@ -1,10 +1,9 @@
-C_OBJS=mb_tests.o micro_booter.o vkernel.o vk_api.o
-ASM_OBJS=cos_asm_scheduler.o inv.o
+ASM_OBJS=cos_asm_scheduler.o inv.o invvk.o
```
phanikishoreg: Kk, I'll get rid of it.

src/components/implementation/no_interface/vkernel/invvk.S
```diff
@@ -0,0 +1,24 @@
...
+#define RET_CAP (1 << 16)
+#endif
+
+.text
+.globl __inv_vkernel_serverfn 
+.type  __inv_vkernel_serverfn, @function
+__inv_vkernel_serverfn:
+	COS_ASM_GET_STACK
+
+	pushl %ebp /* arg 4 */
```
phanikishoreg: Thank you, I'll fix it.

src/components/implementation/no_interface/vkernel/micro_booter.c
```diff
@@ -3,7 +3,7 @@
 #include "vk_api.h"
 
 struct cos_compinfo booter_info;
-thdcap_t termthd = VM_CAPTBL_SELF_EXITTHD_BASE;	/* switch to this to shutdown */
+thdcap_t termthd = BOOT_CAPTBL_SELF_INITTHD_BASE;	/* no exit thread for now */
```
phanikishoreg: I'll change the comment and rename the file to perhaps `vm_booter`.

src/components/implementation/no_interface/vkernel/micro_booter.c
```diff
@@ -17,25 +17,31 @@ int vmid;
 void
 vm_init(void *d)
 {
-	vmid = (int)d;
+	int         rcvd, blocked;
+	cycles_t    cycles;
+	thdid_t     tid;
+	tcap_time_t timeout = 0, thd_timeout;
+
+	vmid = cos_sinv(VM_CAPTBL_SELF_SINV_BASE, VK_SERV_VM_ID << 16 | cos_thdid(), 0, 0, 0);
```
phanikishoreg: I wanted to set a protocol for invocation, where the second parameter is "option" + "initthd id of vm requesting it", nothing to do with the unused arguments, perhaps if I extend this SINV, will need those for something or maybe not.

src/components/implementation/no_interface/vkernel/micro_booter.c
```diff
@@ -17,25 +17,31 @@ int vmid;
...
+
 	cos_meminfo_init(&booter_info.mi, BOOT_MEM_KM_BASE, VM_UNTYPED_SIZE, BOOT_CAPTBL_SELF_UNTYPED_PT);
 	cos_compinfo_init(&booter_info, BOOT_CAPTBL_SELF_PT, BOOT_CAPTBL_SELF_CT, BOOT_CAPTBL_SELF_COMP,
-			  (vaddr_t)cos_get_heap_ptr(), vmid == 0 ? DOM0_CAPTBL_FREE : VM_CAPTBL_FREE, &booter_info);
+			(vaddr_t)cos_get_heap_ptr(), vmid == 0 ? DOM0_CAPTBL_FREE : VM_CAPTBL_FREE, &booter_info);
```
phanikishoreg: Yeah, they're not identical because we have DOM0 and VMs, where DOM0 has I/O capabilities to all VMs and VMs only have I/O caps to DOM0.

src/components/implementation/no_interface/vkernel/vk_api.c
```diff
@@ -1,71 +1,105 @@
...
+static struct cos_aep_info *
+vm_schedaep_get(struct vms_info *vminfo)
+{ return cos_sched_aep_get(&(vminfo->dci)); }
+
 void
 vk_initcaps_init(struct vms_info *vminfo, struct vkernel_info *vkinfo)
 {
-	struct cos_compinfo *vmcinfo = &vminfo->cinfo;
-	struct cos_compinfo *vkcinfo = &vkinfo->cinfo;
+	struct cos_compinfo *vk_cinfo = cos_compinfo_get(cos_defcompinfo_curr_get());
```
phanikishoreg: Kk, I'll look into renaming the function.

src/components/implementation/no_interface/vkernel/vk_api.c
```diff
@@ -1,71 +1,105 @@
...
+	ret = cos_cap_cpy_at(vmcinfo, BOOT_CAPTBL_SELF_UNTYPED_PT, vk_cinfo, vmutpt);
+	assert(ret == 0);
+	ret = cos_cap_cpy_at(vmcinfo, BOOT_CAPTBL_SELF_COMP, vk_cinfo, vmcinfo->comp_cap);
+	assert(ret == 0);
 
-	vminfo->initthd = cos_thd_alloc(vkcinfo, vmcinfo->comp_cap, vm_init, (void *)vminfo->id);
-	assert(vminfo->initthd);
-	vminfo->inittid = (thdid_t)cos_introspect(vkcinfo, vminfo->initthd, THD_GET_TID);
-	printc("\tInit thread= cap:%x tid:%x\n", (unsigned int)vminfo->initthd, (unsigned int)vminfo->inittid);
+	printc("\tInit thread= cap:%x\n", (unsigned int)initaep->thd);
```
phanikishoreg: Gotcha. I'll remove them, hadn't considered because it is mainly within vkernel component.

src/components/implementation/no_interface/vkernel/vk_api.c
```diff
@@ -1,71 +1,105 @@
...
-	ret = cos_cap_cpy_at(vmcinfo, VM_CAPTBL_SELF_EXITTHD_BASE, vkcinfo, vminfo->exitthd);
+
+	ret = cos_cap_cpy_at(vmcinfo, VM_CAPTBL_SELF_SINV_BASE, vk_cinfo, vkinfo->sinv); 
 	assert(ret == 0);
+}
 
-	vminfo->inittcap = cos_tcap_alloc(vkcinfo);
-	assert(vminfo->inittcap);
+void
+vk_sl_thd_init(struct vms_info *vminfo)
```
phanikishoreg: kk, will change the name.

src/components/implementation/no_interface/vkernel/vk_api.c
```diff
@@ -1,71 +1,105 @@
...
+	sl_thd_param_set(vminfo->inithd, spsameprio.v);
+	sl_thd_param_set(vminfo->inithd, spsameC.v);
+	sl_thd_param_set(vminfo->inithd, spsameT.v);
 
-	/*
-	 * Create send end-point in VKernel to each VM's INITRCV end-point
-	 */
-	vkinfo->vminitasnd[vminfo->id] = cos_asnd_alloc(vkcinfo, vminfo->initrcv, vkcinfo->captbl_cap);
-	assert(vkinfo->vminitasnd[vminfo->id]);
+	printc("\tsl_thd 0x%x created for thread = cap:%x, id=%u\n", (unsigned int)(vminfo->inithd),
```
phanikishoreg: got it.

src/components/implementation/no_interface/vkernel/vk_api.c
```diff
@@ -1,71 +1,105 @@
...
+	       (unsigned int)sl_thd_thdcap(vminfo->inithd), (vminfo->inithd)->thdid);
 }
 
 void
 vk_iocaps_init(struct vms_info *vminfo, struct vms_info *dom0info, struct vkernel_info *vkinfo)
 {
-	struct cos_compinfo *vmcinfo = &vminfo->cinfo;
-	struct cos_compinfo *d0cinfo = &dom0info->cinfo;
-	struct cos_compinfo *vkcinfo = &vkinfo->cinfo;
+	struct cos_compinfo *vmcinfo = cos_compinfo_get(&vminfo->dci);
```
phanikishoreg: I think it's fine because the library is mainly using `struct vms_info` as it's abstraction and acting on that data structure.

src/components/implementation/no_interface/vkernel/vk_api.c
```diff
@@ -97,17 +131,19 @@ void
...
 {
+	struct cos_compinfo *vmcinfo = cos_compinfo_get(&(vminfo->dci));
+	struct cos_compinfo *vk_cinfo = cos_compinfo_get(cos_defcompinfo_curr_get());
 	vaddr_t addr;
 
 	assert(vminfo && vkinfo);	
 
 	for (addr = 0 ; addr < range ; addr += PAGE_SIZE) {
-		vaddr_t src_pg = (vaddr_t)cos_page_bump_alloc(&vkinfo->cinfo), dst_pg;
+		vaddr_t src_pg = (vaddr_t)cos_page_bump_alloc(vk_cinfo), dst_pg;
```
phanikishoreg: kk, will try to change it.

src/components/implementation/no_interface/vkernel/vk_api.h
```diff
@@ -4,10 +4,10 @@
 #include "vk_types.h"
 
 /* extern functions */
-extern void vm_exit(void *);
 extern void vm_init(void *);
 extern void dom0_io_fn(void *);
 extern void vm_io_fn(void *);
+extern void *__inv_vkernel_serverfn(int a, int b, int c);
```
phanikishoreg: you're right, it is supposed to be a VM management library. I'll move these externs to other appropriate place and take care of `__inv..` function too.

src/components/implementation/no_interface/vkernel/vk_server.c
```diff
@@ -0,0 +1,47 @@
+
+#include "vk_types.h"
```
phanikishoreg: These are headers in the vkernel directory and only applicable to current dir, so I think `"<local_header>"` is the way to do it.

src/components/implementation/no_interface/vkernel/vk_server.c
```diff
@@ -0,0 +1,47 @@
...
+	for (i = 0 ; i < VM_COUNT ; i ++) {
+		if ((vmx_info[i].inithd)->thdid == tid) break;
+	}
+	assert (i < VM_COUNT);
+
+	return i;
+}
+
+int
+vkernel_server(int a, int b, int c)
```
phanikishoreg: This is the INV function provided by Vkernel to VMs, so named it vk_server, that sounded appropriate to me.

src/components/implementation/no_interface/vkernel/vk_server.c
```diff
@@ -0,0 +1,47 @@
...
+	int ret = 0;
+	int i;
+
+	switch(option) {
+	case VK_SERV_VM_EXIT:
+	{
+		/* free sl threads */
+		i = vkernel_find_vm(thdid);
+		printc("VM%d EXIT\n", i);
+		sl_thd_free(vmx_info[i].inithd);
```
phanikishoreg: There is indeed a lot to free like really undoing what `vk_vm_create` does, but right now my focus is to free the thread so the `sl` doesn't schedule it any more. I'll add a `/* TODO: */` here to take care of that in the future.

src/components/implementation/no_interface/vkernel/vk_types.h
```diff
@@ -3,6 +3,9 @@
 
 #include <cos_types.h>
 #include <cos_kernel_api.h>
+#include <cos_defkernel_api.h>
+#include <sl.h>
+#include <sl_thd.h>
```
phanikishoreg: No, it doesn't.

src/components/implementation/no_interface/vkernel/vk_api.h
phanikishoreg: I'm sorry, this is not really a generic library, it is very specific to vkernel component and the main focus for `vk_api.c/h` is modularizing the code.

src/components/implementation/no_interface/vkernel/vk_types.h
```diff
@@ -11,14 +14,14 @@
 #define VM_SHM_SZ       (1<<20)	        /* Shared memory mapping for each vm = 4MB */
 #define VM_SHM_ALL_SZ   ((VM_COUNT>0)?(VM_COUNT*VM_SHM_SZ):VM_SHM_SZ)
 
-#define VM_BUDGET_FIXED 400000
-#define VM_PRIO_FIXED   TCAP_PRIO_MAX
+#define VM_FIXED_PERIOD_MS 10
+#define VM_FIXED_BUDGET_MS 5
```
phanikishoreg: I didn't focus here at all on making parts of vkernel a generic library. Maybe, I'll add that as an issue to move such parts especially "forking" as library.

src/components/implementation/no_interface/vkernel/vk_types.h
```diff
@@ -11,14 +14,14 @@
...
 
 enum vm_captbl_layout {
-	VM_CAPTBL_SELF_EXITTHD_BASE    = BOOT_CAPTBL_FREE,
+	VM_CAPTBL_SELF_SINV_BASE       = BOOT_CAPTBL_FREE,
 
 	/* VM1~ I/O Capabilities layout */
-        VM_CAPTBL_SELF_IOTHD_BASE      = round_up_to_pow2(VM_CAPTBL_SELF_EXITTHD_BASE + NUM_CPU_COS*CAP16B_IDSZ, CAPMAX_ENTRY_SZ),
+        VM_CAPTBL_SELF_IOTHD_BASE      = round_up_to_pow2(VM_CAPTBL_SELF_SINV_BASE + NUM_CPU_COS*CAP32B_IDSZ, CAPMAX_ENTRY_SZ),
```
phanikishoreg: You're right, I just looked into `cos_types.h`, we have `__captbl_cap2sz` that takes in CAP TYPE and returns size.

phanikishoreg: I'll change to use that.

src/components/implementation/tests/micro_booter/mb_tests.c
```diff
@@ -401,7 +346,7 @@ test_timer(void)
 		if (i > 0) t += c-p;
 
 		/* FIXME: we should avoid calling this two times in the common case, return "more evts" */
```
phanikishoreg: I've the `RCV_ALL_PENDING` support but this `cos_sched_rcv` isn't fixed.

src/components/implementation/tests/micro_booter/mb_tests.c
```diff
@@ -468,10 +413,11 @@ test_budgets_single(void)
 
 	PRINTC("Budget switch latencies: ");
 	for (i = 1 ; i < 10 ; i++) {
```
phanikishoreg: This has been there, I didn't work on fixing hard-coded things in `micro_booter`. Changes here are specific to `cos_rcv/cos_sched_rcv` api changes.

src/components/implementation/tests/micro_booter/mb_tests.c
```diff
@@ -925,15 +876,35 @@ test_run_mb(void)
  * TODO: Fix those eventually.
```
phanikishoreg: I'll look into it.

src/components/implementation/tests/micro_booter/mb_tests.c
```diff
@@ -512,12 +459,13 @@ test_budgets_multi(void)
 		if (cos_tcap_transfer(mbt.c.rc, mbt.p.tcc, res/2, TCAP_PRIO_MAX + 2)) assert(0);
 		if (cos_tcap_transfer(mbt.g.rc, mbt.c.tcc, res/4, TCAP_PRIO_MAX + 2)) assert(0);
 
+		mbt.p.cyc = mbt.c.cyc = mbt.g.cyc = 0;
 		rdtscll(s);
 		if (cos_switch(mbt.g.tc, mbt.g.tcc, TCAP_PRIO_MAX + 2, TCAP_TIME_NIL, BOOT_CAPTBL_SELF_INITRCV_BASE, cos_sched_sync())) assert(0);
 		rdtscll(e);
 		PRINTC("g:%llu c:%llu p:%llu => %llu,\t", mbt.g.cyc - s, mbt.c.cyc - s, mbt.p.cyc - s, e - s);
```
phanikishoreg: kk, got it. will keep that in mind.

src/components/implementation/tests/micro_booter/mb_tests.c
```diff
@@ -925,15 +876,35 @@ test_run_mb(void)
...
  */
 void
+__block_vm(void)
+{
+	int blocked, rcvd;
+	cycles_t cycles, now;
+	tcap_time_t timeout, thd_timeout;
+	thdid_t tid;
+
+	while (cos_sched_rcv(BOOT_CAPTBL_SELF_INITRCV_BASE, RCV_ALL_PENDING, 0, &rcvd, &tid, &blocked, &cycles, &thd_timeout) > 0);
```
phanikishoreg: This is to receive any pending events, thanks for pointing out. I should also use `RCV_NON_BLOCKING` flag. The `cos_sched_rcv` after this line is guaranteed to block with a timeout.

src/components/implementation/tests/unit_defcompinfo/unit_defcompinfo.c
```diff
@@ -199,7 +201,7 @@ cos_init(void)
 
 		/* TEST BLOCKING */
 		/* TODO: Challenge - how does a component know at runtime if can call cos_rcv or not? - It does not at runtime. */
-		cos_rcv(BOOT_CAPTBL_SELF_INITRCV_BASE, 0, NULL);
+		cos_rcv(BOOT_CAPTBL_SELF_INITRCV_BASE, 0, 0, NULL);
```
phanikishoreg: Makes sense. Will need to look into. For now though, i'll add this as an issue.

src/components/include/cos_kernel_api.h
```diff
@@ -137,9 +137,9 @@ int cos_thd_mod(struct cos_compinfo *ci, thdcap_t c, void *tls_addr); /* set tls
 
 int cos_asnd(asndcap_t snd, int yield);
 /* returns non-zero if there are still pending events (i.e. there have been pending snds) */
-int cos_rcv(arcvcap_t rcv, rcv_flags_t flags, int *rcvd);
+int cos_rcv(arcvcap_t rcv, rcv_flags_t flags, tcap_time_t timeout, int *rcvd);
```
phanikishoreg: I thought it's an useful utility for some threads that they either wait for an event or for a timeout.

src/components/include/sl.h
```diff
@@ -321,6 +328,31 @@ sl_timeout_wakeup_expired(cycles_t now)
...
+static inline int
+sl_thd_activate(struct sl_thd *t, sched_tok_t tok)
+{
+	struct cos_defcompinfo *dci = cos_defcompinfo_curr_get();
+	struct cos_compinfo    *ci  = &dci->ci;
+
+	switch(t->type) {
+	case SL_THD_THD:
+	case SL_THD_AEP:
+	{
```
phanikishoreg: kk, will consider it. Thank you.

src/components/include/sl.h
```diff
@@ -321,6 +328,31 @@ sl_timeout_wakeup_expired(cycles_t now)
...
+		return cos_defswitch(sl_thd_thdcap(t), t->prio, sl__globals()->timeout_next, tok);	
+	}
+	case SL_THD_AEP_TCAP:
+	{
+		return cos_switch(sl_thd_thdcap(t), sl_thd_tcap(t), t->prio, sl__globals()->timeout_next, sl__globals()->sched_rcv, tok);
```
phanikishoreg: AEPs can either be using scheduler TCaps or their own. So this switch. 
I've a TODO to abstract this is `cos_defkernel_api` to have `cos_aep_defswitch()` which will use the tcap associated with that AEP. 
The second type of AEP is COMP, which is child component, it could be having it's own tcap/ will probably use parent's.

src/components/include/sl.h
```diff
@@ -321,6 +328,31 @@ sl_timeout_wakeup_expired(cycles_t now)
...
+		return cos_defswitch(sl_thd_thdcap(t), t->prio, sl__globals()->timeout_next, tok);	
+	}
+	case SL_THD_AEP_TCAP:
+	{
+		return cos_switch(sl_thd_thdcap(t), sl_thd_tcap(t), t->prio, sl__globals()->timeout_next, sl__globals()->sched_rcv, tok);	
+	}
+	case SL_THD_COMP:
+	case SL_THD_COMP_TCAP:
+	{
+		return cos_asnd(t->sndcap, 1);
```
phanikishoreg: Not yet, I'm working on returning perhaps `-EPERM` from `cos_asnd` and handle that here to like you said suspend it.

src/components/include/sl.h
```diff
@@ -382,13 +412,22 @@ sl_cs_exit_schedule_nospin_arg(struct sl_thd *to)
...
-	prio   = t->prio;
 
+	if (t->type == SL_THD_AEP_TCAP || t->type == SL_THD_COMP_TCAP) {
+		assert(t->budget && t->period);
+
+		if (t->last_replenish == 0 || t->last_replenish + t->period <= now) {
+			t->last_replenish = now;
+
+			/* TODO: need to change logic for SNDCAP with tcap_delegate, and error handling */
+			if (cos_tcap_transfer(sl_thd_rcvcap(t), sl__globals()->sched_tcap, t->budget, t->prio)) assert(0);
```
phanikishoreg: Transfers are only if the AEP or COMP thread has it's own TCAP. For the ones that use the current scheduler's tcap, we don't do it. So that's why I've 4 different types of AEPs.

phanikishoreg: Thanks for catching it, it should just be the difference. Will fix it.

src/components/include/sl_thd.h
```diff
@@ -17,20 +17,52 @@ typedef enum {
 	SL_THD_DYING,
 } sl_thd_state;
 
+typedef enum {
+	SL_THD_THD = 0,
+	SL_THD_AEP,
+	SL_THD_AEP_TCAP,
```
phanikishoreg: gotcha, will add comments.

src/components/lib/sl/sl.c
```diff
@@ -133,10 +133,26 @@ sl_thd_block_no_cs(struct sl_thd *t, sl_thd_state block_type)
 		return 1;
 	}
 
+	/*
+	 * If an AEP/a child COMP was blocked and an interrupt caused it to wakeup and run
+	 * but blocks itself before the scheduler could see the wakeup event.. Scheduler
```
phanikishoreg: Kk. I'll try to explain it here: If an AEP thread (not considering child schedulers for now) is in BLOCKED state and an Interrupt caused it to unblock, kernel adds an event to the current scheduler (AEP's parent) and switches to the AEP thread if it's of higher prio than the currently running thread. If that AEP interrupt thread finishes it's execution and ends up (suspends) in `cos_rcv` and only after this if the scheduler wakes up and sees an event for this AEP thread but because the AEP is now in `cos_rcv`, it sees it as a blocked event. Basically that is a blocked event on an already blocked thread for `sl` library.

src/components/lib/sl/sl.c
```diff
@@ -291,48 +306,119 @@ sl_thd_yield(thdid_t tid)
 	sl_thd_yield_cs_exit(tid);
 }
 
+static void
+sl_thd_aepinfo_caps_init(struct sl_thd *t, thdcap_t thd, arcvcap_t rcv, tcap_t tc)
```
phanikishoreg: kk, I'll get rid of `caps` from it.

src/kernel/capinv.c
phanikishoreg: thank you.

phanikishoreg:
Yes, I'll soon be addressing this major concern of being able to detect out of budget case and handling it appropriately in `sl`.

src/components/implementation/no_interface/vkernel/micro_booter.c
gparmer: OK, but why not set a protocol of "first argument is option, second is thread id"?  Just want to save registers for actual use?

src/components/implementation/no_interface/vkernel/micro_booter.c
gparmer: this doesn't seem like something we want to hard code in the library.  It seems like something that would be hardcoded by the booter that is *using* the library.  Different booters might want to set up the VMs differently, right?

src/components/implementation/no_interface/vkernel/vk_api.c
gparmer: If it sin't a library, I'm with you.  If it is a library (and I'd like it to be), then it isn't desirable.

src/components/implementation/no_interface/vkernel/vk_api.c
gparmer: Fair reply.  I meant (and didn't convey) something different.

Do we really want to be hardcoding dom0 into the library?  Shouldn't that be a construct of the code *using* the library.  See previous comment.

src/components/implementation/no_interface/vkernel/micro_booter.c
phanikishoreg: Yes, somewhat that. 
I can make it so first is option, second is thread id. But if I have to extend in the future to accommodate more arguments, then I'll have to change the protocol. This protocol I've right now is designed so I don't have to change the basic parts of that protocol.

src/components/implementation/tests/micro_booter/mb_tests.c
```diff
@@ -925,15 +876,35 @@ test_run_mb(void)
  * TODO: Fix those eventually.
  */
 void
+__block_vm(void)
```
phanikishoreg: ;) I don't know why I did that. :( I'll change it, also I'll need to make it a `static` function.

src/components/include/sl.h
```diff
@@ -382,13 +412,22 @@ sl_cs_exit_schedule_nospin_arg(struct sl_thd *to)
 		if (unlikely(!pt)) t = sl__globals()->idle_thd;
 		else               t = sl_mod_thd_get(pt);
 	}
-	thdcap = t->thdcap;
-	prio   = t->prio;
 
+	if (t->type == SL_THD_AEP_TCAP || t->type == SL_THD_COMP_TCAP) {
```
phanikishoreg: I agree. But I don't know of a better way to handle this and have no impact on the fast-path.

src/components/implementation/tests/micro_booter/mb_tests.c
phanikishoreg: Yeah, I think it's my habit to prefix `__` for a function that's private (like `static`) but keep forgetting to make it `static` at the very least.

src/components/implementation/no_interface/vkernel/vk_api.h
gparmer: OK, so how do you want to go forward with this?  Will other people be able to use any of this code?  Will you be able to use this code for your research (this paper)?  It is obvious we need to make this general at some point (i.e. the next level of libraries above cos_kernel_api + cos_defkernel_api + sl).  When do we do that?

src/components/implementation/no_interface/vkernel/vk_server.c
gparmer: Using <> puts the power in the build system, which is generally what you want.  I believe that `-I.` is in the include path so it should find local files.  What do you think?

src/components/implementation/no_interface/vkernel/vk_server.c
gparmer: `server` is a noun which is strange for an execution entry point.

`vkernel_hypercall` might better convey what is going on here.

src/components/implementation/no_interface/vkernel/vk_types.h
gparmer: going to 64 bit will mess up all of this if we don't use the helper macros.

src/components/include/cos_kernel_api.h
gparmer: schedulers should handle timeouts, not general threads.  Don't implement everywhere something that should be handled in a very specific place.  Lets keep timeouts as a scheduler operation unless we have a concrete reason to add complexity.

src/components/include/sl.h
gparmer: Having a single flag as suggested above instead of so many thread types would be a better way, I tihnk.

src/components/include/sl.h
gparmer: another reason I don't like the `COMP` in all of this is because it is possible that we want to schedule a thread using that style (it has its own tcap and aep) even if it isn't a separate component...  "COMP" seems orthogonal to the decisions being made here.

src/components/implementation/no_interface/vkernel/micro_booter.c
phanikishoreg: I'm confused when you say `library` and you've said that so many times in this review. :( 
I don't really have a library, it's only parts of `vkernel` component that are modularised into `vk_api` and parts of it is specific to how the vkernel, vms, dom0 interract with each other in the current vkernel component implementation.

I may be able to make parts of it generic, like, really not call it `vm`, make it a `fork_comp` kind of API. None of the current `vkernel` has that. I totally understand that it is good to make parts of it a generic library but for the current `vkernel`, we don't have a library, that's my confusion when you say `library`. I'll add an issue to possible make a generic library out of it and work on it later. 

Coming to this `micro_booter.c` and `vm_init`, it is the main function of a `vm` and is doing `cos_compinfo_init` for it's component so it starts its `cap frontier` where it knows it should which is why when DOM0 boots up it's frontier is set to a different value than when another VM boots up.

src/components/implementation/no_interface/vkernel/vk_api.c
phanikishoreg: Yes, but I'm sorry I agreed it's a library here. It is not really but it is some kind of abstraction within vkernel and is very specific to how it is designed and implemented right now.

src/components/implementation/no_interface/vkernel/vk_api.h
phanikishoreg: I'm able to use it for my research and others are mainly using this as reference. I know and I understand that it has the potential to be a library but I don't know if it is going to take a lot of time to make it one. :( 
I'd work on that after the deadline.

src/components/implementation/no_interface/vkernel/vk_server.c
phanikishoreg: Is this a `personal preference` or something like a CSG standard? 
To me it is good to have `""` for local headers because when I look at it, I know that it is specific to this directory and changing anything there will not affect other parts of the system. I'd still go with `""` if it were a `personal choice`.

src/components/implementation/no_interface/vkernel/vk_types.h
gparmer: OK, sounds good.

src/components/implementation/no_interface/vkernel/vk_server.c
phanikishoreg: kk, thank you. I'm really bad at naming functions! :( I'll change it.

src/components/implementation/no_interface/vkernel/vk_types.h
phanikishoreg: I've used mainly how `BOOT_CAPTBL_SELF_XXX_BASE` in `cos_types.h` were setup as reference, this has been there from the first implementation of `vkernel`. 

I totally agree that it's not portable and I'll change. Thank you!

src/components/include/cos_kernel_api.h
phanikishoreg: Kk, I'll undo this change for `cos_rcv`.

src/components/include/sl.h
phanikishoreg: I'll explain each thread type: 

1. `SL_THD_THD`: This is a thread that is not associated with RCVCAP or TCAP. So scheduler will basically use it's tcap to schedule this thread using `cos_switch()`. This is how we'd implement normal worker threads. 
2. `SL_THD_AEP`: This is a thread that is associated with it's RCVCAP and associates to scheduler TCAP for it's time. This is not a child component thread, you can think of it as a Interrupt thread. So scheduler would switch to it using `cos_defswitch` by using scheduler tcap's time. This is how we'd implement interrupts that don't need temporal isolation from other threads in that scheduling domain and tcap quality properties entirely. This is for implementing `cosXen` and it's variations that we had in our previous paper where all parts of that VM uses a single TCap. 
3. `SL_THD_AEP_TCAP`: This is very much like the one above, but instead of using scheduler's tcap for time, it uses it's own TCap. This is again an interrupt thread so scheduler would switch to it using `cos_switch` and provides time to that AEP through `cos_tcap_transfer`. 
This is how we'd implement AEP that get both its own time from its scheduler and perhaps delegations from other parts of the system. 
4. `SL_THD_COMP`: This is a child component, and uses current scheduler's tcap for it's time. So the current scheduler doesn't have any additional `cos_tcap_transfer` for transfering time to it. Scheduler dispatches to it using `cos_asnd`. I'm not sure if there is a good reason for me to have this nor do I know specific experiments that we'd use this for. I can only say, we'd use this if everything in the system uses one timeslice source including root, child schedulers. 
5. `SL_THD_COMP_TCAP`: This is a child component, that we'd have in most cases with it's own TCAP. Scheduler either uses `cos_tcap_delegate` or a pair of `cos_tcap_transfer` + `cos_asnd` to provide time and dispatch to it. This is how we'd have most implementations of components, also enables distributed time management through Tcaps. 

Here, AEP doesn't mean use `asnd` to switch to it. Scheduler uses `cos_switch` or `cos_defswitch` for the first 3 types and `cos_asnd` for the last 2. 
My idea here is to be able to provide all possible types of threads that can be required by different implementations and different experiments that we'd do. Especially because we'd use `sl` for most scheduling the system. If you think I have types that don't need to be there ex: `SL_THD_COMP` without it's own TCap, I can remove them.

src/components/implementation/tests/micro_booter/mb_tests.c
phanikishoreg: Actually, I think there is nothing to fix here, its just that we cannot use all micro-benchmark test cases for virtual-machine environments, so taking out only the FIXME.

src/components/implementation/tests/unit_defcompinfo/unit_defcompinfo.c
phanikishoreg: I've taken out the `timeout` parameter from `cos_rcv`, do you still think this is something we should look into even without that?

src/components/include/sl.h
phanikishoreg: Removed thread types, using flags like you suggested instead!

src/components/include/sl.h
phanikishoreg: Changed types to flags: `FLAG_SEND` to asnd instead of switch and `FLAG_OWNTC` to use a dedicated tcap for this thread.

src/components/include/sl.h
phanikishoreg: Still a TODO. I've added `-EPERM` to `cos_asnd` but need to handle it.

src/components/include/sl.h
phanikishoreg: Got it! Done!

phanikishoreg:
@gparmer I've updated the commits to this PR with fixes for your comments. [only pending item: need to address detecting budget expiry case].

src/kernel/include/inv.h
```diff
@@ -309,8 +309,6 @@ sret_ret(struct thread *thd, struct pt_regs *regs, struct cos_cpu_local_info *co
 	}
 
 	pgtbl_update(ci->pgtbl);
-	/* Set 2/3 return values into esi and edi */
-	__userregs_setretvals(regs, 0, thd->tid, 0);
```
phanikishoreg: @gparmer I wasn't sure why this exists in `SRET` path, so I took it out when I was extending `__userregs_setretvals` to return 3 values. I may be missing something, even `cos_sinv` API doesn't have additional output parameters. 
Is there some gap between the intention behind the `cos_sinv` API and actual implementation?

src/components/include/sl.h
```diff
@@ -382,13 +404,24 @@ sl_cs_exit_schedule_nospin_arg(struct sl_thd *to)
 		if (unlikely(!pt)) t = sl__globals()->idle_thd;
 		else               t = sl_mod_thd_get(pt);
 	}
-	thdcap = t->thdcap;
-	prio   = t->prio;
 
+	if (t->flags & SL_FLAG_OWNTC) {
+		assert(t->budget && t->period);
```
gparmer: Can we move this assertion down, and just have this conditional be 

```
if (owned tcap && (replenish == 0 || replenishment period elapsed)) {...}
```

???

src/components/lib/sl/sl.c
```diff
@@ -291,48 +306,101 @@ sl_thd_yield(thdid_t tid)
...
+	int                     ret;
+
+	if (flags & SL_FLAG_SEND) {
+		struct cos_aep_info *sa;
+
+		assert(comp);
+		sa = cos_sched_aep_get(comp);
+
+		/* FIXME: inefficient.. find a zerocopy way! */
+		aep.tc  = sa->tc;
```
gparmer: I honestly think that `aep = *sa;` would work here (the compiler turns it into memcpy).

This whole approach is not pretty, but I think you did a good job working with the APIs we have.

src/components/lib/sl/sl.c
```diff
@@ -343,20 +411,36 @@ sl_thd_alloc(cos_thd_fn_t fn, void *data)
...
+	t = sl_thd_alloc_intern(fn, data);
 	sl_cs_exit();
 
 	return t;
 }
 
-/* Allocate a thread that executes in the specified component */
 struct sl_thd *
-sl_thd_comp_alloc(struct cos_defcompinfo *comp)
+sl_thd_aep_alloc(cos_aepthd_fn_t fn, void *data, int owntc)
```
gparmer: I like changing the `tc` to `tcap` and perhaps two words `own_tcap`, or better `separate_tcap`.  I don't think that everyone will know what a `tc` is as it is an edge of the api.

src/components/lib/sl/sl.c
```diff
@@ -343,20 +411,36 @@ sl_thd_alloc(cos_thd_fn_t fn, void *data)
...
-	t = sl_thd_alloc_intern(NULL, NULL, comp, 1);
+	t = sl_thd_aep_alloc_intern(fn, data, NULL, owntc ? SL_FLAG_OWNTC: 0);
+	sl_cs_exit();
+
+	return t;
+}
+
+/* sl object for inithd in the child comp */
+struct sl_thd *
+sl_thd_comp_init(struct cos_defcompinfo *comp, int owntc)
```
gparmer: As you commented, do we care about the case where the child uses the same tcap?  Perhaps so.  Maybe worth a 5 minute discussion on Wed.  Can keep it as it is (perhaps with a new name) for now.

src/components/lib/sl/sl.c
```diff
@@ -343,20 +411,36 @@ sl_thd_alloc(cos_thd_fn_t fn, void *data)
...
+{
+	struct sl_thd *t;
+
+	assert(comp);
+	if (owntc) assert(cos_sched_aep_get(comp)->tc != sl_thd_tcap(sl__globals()->sched_thd));
```
gparmer: You cannot put code that must run in `assert`s.  They can be compiled away at some future point in time.

src/components/include/sl.h
```diff
@@ -321,6 +328,21 @@ sl_timeout_wakeup_expired(cycles_t now)
...
 
+static inline int
+sl_thd_activate(struct sl_thd *t, sched_tok_t tok)
+{
+	struct cos_defcompinfo *dci = cos_defcompinfo_curr_get();
+	struct cos_compinfo    *ci  = &dci->ci;
+
+	if (t->flags & SL_FLAG_SEND) {
+		return cos_asnd(t->sndcap, 1);
+	} else {
```
gparmer: This all feels a lot cleaner to me.  You too?

src/components/include/sl.h
```diff
@@ -321,6 +328,21 @@ sl_timeout_wakeup_expired(cycles_t now)
...
 }
 
+static inline int
+sl_thd_activate(struct sl_thd *t, sched_tok_t tok)
+{
+	struct cos_defcompinfo *dci = cos_defcompinfo_curr_get();
+	struct cos_compinfo    *ci  = &dci->ci;
+
+	if (t->flags & SL_FLAG_SEND) {
+		return cos_asnd(t->sndcap, 1);
```
gparmer: We have an issue to add the token as optional to `asnd` right?

phanikishoreg: No. It's literally the next thing I'll be pushing perhaps in the same PR.

src/components/include/sl.h
phanikishoreg: Yes, we can. I'll fix it.

src/components/include/sl.h
```diff
@@ -382,13 +404,24 @@ sl_cs_exit_schedule_nospin_arg(struct sl_thd *to)
...
-	prio   = t->prio;
 
+	if (t->flags & SL_FLAG_OWNTC) {
+		assert(t->budget && t->period);
+
+		if (t->last_replenish == 0 || t->last_replenish + t->period <= now) {
+			tcap_res_t currbudget;
+
+			t->last_replenish = now;
+			currbudget        = (tcap_res_t)cos_introspect(ci, sl_thd_tcap(t), TCAP_GET_BUDGET);
```
phanikishoreg: I'll add an issue and will remember to discuss with you on this. 

I don't think we have a good way to track the expenditure of a tcap. we need expenditure of a tcap especially in case of child component, but we literally cannot do anything about this right now. It gets complicated when we have the child component scheduler transfering time to its child AEPs. Even `introspect` is not a complete solution to this.

src/components/include/sl.h
```diff
@@ -382,13 +404,24 @@ sl_cs_exit_schedule_nospin_arg(struct sl_thd *to)
...
+			/* TODO: need to change logic for SNDCAP with tcap_delegate, and error handling */
+			if (currbudget < t->budget && cos_tcap_transfer(sl_thd_rcvcap(t), sl__globals()->sched_tcap, (t->budget - currbudget), t->prio)) assert(0);
```
phanikishoreg: My bad, it should not be an assertion. Especially if we try to transfer from within child scheduler having finite budget.

src/components/include/sl.h
phanikishoreg: Kk, I'll fix it.

src/components/include/sl_thd.h
```diff
@@ -17,20 +17,49 @@ typedef enum {
 	SL_THD_DYING,
 } sl_thd_state;
 
+typedef enum {
+	SL_FLAG_OWNTC = 1,      /* Thread owns a tcap */
```
phanikishoreg: Kk, I don't really have better ideas for naming. I'll go with `SL_THD_PROPERTY_XXXX`, it sounds good. Will fix the enum typedef naming too.

src/components/lib/sl/sl.c
phanikishoreg: Kk, I'll do that.

src/components/lib/sl/sl.c
phanikishoreg: kk, i'll take care of that.

src/components/lib/sl/sl.c
phanikishoreg: We probably don't. 
I think there is still a missing piece here, we have `cos_defkernel_api` child alloc api where we use parent's `tc` and also child component doesn't have a scheduler. 
So in that case, we need to make this flag `int is_sched` and turn the flag into `XXX_SEND` or into `0` as it's just an AEP with parent's tcap. 
In that case, I think we'll need invocation API in `sl` library for the child to alloc more threads to be handled by parent scheduler?

src/components/lib/sl/sl.c
phanikishoreg: There is no useful code that's used outside of that `assert`, It is specifically to do some consistency checks with assertions and if asserts are compiled away, it is absolutely fine here.

gparmer:
sorry, but @Others PR killed this one's ability to merge correctly.  Feel free to either close this PR and reissue, or update it after doing a `pull`.  Sorry!!!  Shouldn't be *any* logic changes, just formatting.

PR #289, opened 2017-09-23
Bugfixes in the kernel, changed sl to allow NON_BLOCKING sched_rcv 

phanikishoreg:

Major changes:
KERNEL:
* Fixed a problem where timeout was set when we change the tcap and switch to parent or another scheduler. This is a TODO for child scheds abiding by parent timeout rules, but for now, we don't have it.
* A problem with restoring registers which I honestly complicated with confusing conditions before. Fixed that.

SL LIB:
* Allow `sl_sched_loop` to use `NON_BLOCKING` in `cos_sched_rcv`. I feel like the naming in `SL_SCHEDRCV_DEFAULT` is not great. 
* Setting timeouts for `cos_defswitch()` when switching to the scheduler is not necessary. In fact, it was causing RK scheduler to switch to chronos even though RK had INF budget. Fixed that!

### Intent for your PR

Choose one (Mandatory):

### Reviewers (Mandatory):
@gparmer 

### Code Quality

### Testing

- microbooter, unitschedtest, unit fprr
- rump2cos env where most of these bugs are found.

### DIFF
--- a/src/components/include/sl.h
+++ b/src/components/include/sl.h
@@ -528,6 +528,6 @@ sl_cs_exit_switchto(struct sl_thd *to)
  * sl_sched_loop(); <- loop here
  */
 void sl_init(microsec_t period);
-void sl_sched_loop(void);
+void sl_sched_loop(int non_block);
 
 #endif /* SL_H */
--- a/src/components/include/sl.h
+++ b/src/components/include/sl.h
@@ -363,14 +363,16 @@ sl_thd_activate(struct sl_thd *t, sched_tok_t tok)
 {
 	struct cos_defcompinfo *dci = cos_defcompinfo_curr_get();
 	struct cos_compinfo    *ci  = &dci->ci;
+	struct sl_global       *g   = sl__globals();
 
 	if (t->properties & SL_THD_PROPERTY_SEND) {
-		return cos_sched_asnd(t->sndcap, sl__globals()->timeout_next, sl__globals()->sched_rcv, tok);
+		return cos_sched_asnd(t->sndcap, g->timeout_next, g->sched_rcv, tok);
 	} else if (t->properties & SL_THD_PROPERTY_OWN_TCAP) {
 		return cos_switch(sl_thd_thdcap(t), sl_thd_tcap(t), t->prio,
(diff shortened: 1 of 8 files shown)


src/components/include/sl.h
```diff
@@ -526,6 +528,6 @@ sl_cs_exit_switchto(struct sl_thd *to)
  * sl_sched_loop(); <- loop here
  */
 void sl_init(microsec_t period);
-void sl_sched_loop(void);
+void sl_sched_loop(int non_block);
```
gparmer: Shouldn't this has an `enum` type or something?  Or is it flags?  The use of this is very unclear to me.

src/components/lib/sl/sl.c
```diff
@@ -595,9 +596,10 @@ sl_init(microsec_t period)
 }
 
 void
-sl_sched_loop(void)
+sl_sched_loop(int non_block)
 {
-	struct sl_global *g = sl__globals();
+	struct sl_global *g   = sl__globals();
+	rcv_flags_t       rfl = (non_block ? RCV_NON_BLOCKING : 0) | RCV_ALL_PENDING;
```
gparmer: OK, so you're just using `non_block` as binary, but then are assigning a name to it.  This seems strange.

I can't imagine this should be an argument at the API level.  `sl_sched_loop()` should remain for default use, and you can create a `sl_sched_loop_nonblock()` for whatever this is all for.

You keep wanting to make the API more complex for common usage.  You need to push back against this impulse.

All of that said, I have no idea why want this to be nonblocking.

gparmer:
I don't see a comment for the TODO you're referencing.

src/components/lib/sl/sl.c
phanikishoreg: Yes, I had the same feeling that I made the API a bit confusing to use. 
I have mentioned the reasoning for nonblocking API briefly in the commit message, but I will hold this PR for now and discuss about it the next time we meet.

phanikishoreg:
I'm sorry, I don't have a TODO in the code. I'll add it.

phanikishoreg:
Closing this for now.
Will do fresh PR after fixing for your comments and with some updates for Baremetal documentation soon.

PR #291, opened 2017-10-18
Add futex implementation to libposix

### LINKED ISSUE #287: Data Race in Rust - println!()
rusts println!() does not appear to take a lock when printing and is able to be preempted causing an "already borrowed:borrowMutError"

Others:

Add futex implementation to libposix. Also improve our faking of thread local data for musl. Fixes #287

### Intent for your PR

Choose one (Mandatory):

### Reviewers (Mandatory):
@phanikishoreg @maloneya @gparmer 

### Code Quality

### Testing

- events_prototype

### DIFF
--- a/src/components/lib/posix/posix.c
+++ b/src/components/lib/posix/posix.c
@@ -100,7 +100,6 @@ cos_mmap(void *addr, size_t length, int prot, int flags, int fd, off_t offset)
 {
 	void *ret=0;
 
-	printc("mmap\n");
 	if (addr != NULL) {
 		printc("parameter void *addr is not supported!\n");
 		errno = ENOTSUP;
@@ -198,12 +197,21 @@ cos_tkill(int tid, int sig)
 	return 0;
 }
 
+static inline microsec_t
+time_to_microsec(const struct timespec *t)
+{
+	time_t seconds    		= t->tv_sec;
+	long nano_seconds 		= t->tv_nsec;
+	microsec_t microseconds = seconds * 1000000 + nano_seconds / 1000;
+	return microseconds;
+}
+
 long
 cos_nanosleep(const struct timespec *req, struct timespec *rem)
 {
-	time_t seconds, remaining_seconds;
-	long nano_seconds, remaining_nano_seconds;
-	microsec_t microseconds, remaining_microseconds;
+	time_t remaining_seconds;
+	long remaining_nano_seconds;
+	microsec_t remaining_microseconds;
 	cycles_t wakeup_deadline, wakeup_time;
 	int completed_successfully;
 
@@ -211,11 +219,8 @@ cos_nanosleep(const struct timespec *req, struct timespec *rem)
 		errno = EFAULT;
 		return -1;
 	}
-	seconds      = req->tv_sec;
-	nano_seconds = req->tv_nsec;
-	microseconds = seconds * 1000000 + nano_seconds / 1000;
 
-	wakeup_deadline        = sl_now() + sl_usec2cyc(microseconds);
+	wakeup_deadline        = sl_now() + sl_usec2cyc(time_to_microsec(req));
 	completed_successfully = sl_thd_block_timeout(0, wakeup_deadline);
 	wakeup_time   = sl_now();
 
@@ -273,7 +278,6 @@ setup_thread_area(struct sl_thd *thread, void* data)
 int
 cos_set_thread_area(void* data)
 {
-	printc("cos_set_thread_area %p\n", data);
 	setup_thread_area(sl_thd_curr(), data);
 	return 0;
 }
@@ -291,6 +295,149 @@ cos_clone(int (*func)(void *), void *stack, int flags, void *arg, pid_t *ptid, v
 		setup_thread_area(thd, tls);
 	}
 	return thd->thdid;
+}
+
+#define FUTEX_WAIT		0
+#define FUTEX_WAKE		1
+#define FUTEX_FD		2
+#define FUTEX_REQUEUE		3
+#define FUTEX_CMP_REQUEUE	4
+#define FUTEX_WAKE_OP		5
+#define FUTEX_LOCK_PI		6
+#define FUTEX_UNLOCK_PI		7
+#define FUTEX_TRYLOCK_PI	8
+#define FUTEX_WAIT_BITSET	9
+
+#define FUTEX_PRIVATE 128
+
+#define FUTEX_CLOCK_REALTIME 256
+
+struct futex_data
+{
+	int *uaddr;
+	struct ps_list_head waiters;
+};
+
+struct futex_waiter
+{
+	thdid_t thdid;
+	struct ps_list list;
+};
+
+#define FUTEX_COUNT 20
+struct futex_data futexes[FUTEX_COUNT];
+
+struct futex_data *
+lookup_futex(int* uaddr)
+{
+	int last_free = -1;
+	int i;
+	for (i = 0; i < FUTEX_COUNT; i++) {
+		if (futexes[i].uaddr == uaddr) {
+			return &futexes[i];
+		} else if (futexes[i].uaddr == 0) {
+			last_free = i;
+		}
+	}
+	if (last_free >= 0) {
+		futexes[last_free] = (struct futex_data) {
+			.uaddr = uaddr
+		};
+		ps_list_head_init(&futexes[last_free].waiters);
+		return &futexes[last_free];
+	}
+	printc("Out of futex ids!");
+	assert(0);
+}
+
+// TODO: Cleanup empty futexes
+
+struct sl_lock futex_lock = SL_LOCK_STATIC_INIT();
+
+int
+cos_futex_wait(struct futex_data *futex, int *uaddr, int val, const struct timespec *timeout)
+{
+	int not_awoken;
+	cycles_t   deadline;
+	microsec_t wait_time       = 0;
+	struct futex_waiter waiter = (struct futex_waiter) {
+		.thdid = sl_thdid()
+	};
+
+	if (*uaddr != val) {
+		return EAGAIN;
+	}
+
+	ps_list_init_d(&waiter);
+	ps_list_head_append_d(&futex->waiters, &waiter);
+
+	if (timeout != NULL) {
+		wait_time = time_to_microsec(timeout);
+	}
+
+	// No race here, we'll enter the awoken state if things go wrong
+	sl_lock_release(&futex_lock);
+	if (wait_time == 0) {
+		sl_thd_block(0);
+	} else {
+		deadline = sl_now() + sl_usec2cyc(wait_time);
+		sl_thd_block_timeout(0, deadline);
+	}
+	sl_lock_take(&futex_lock);
+	ps_list_rem_d(&waiter);
+	return 0;
+}
+
+int cos_futex_wake(struct futex_data *futex, int wakeup_count)
+{
+	struct futex_waiter *waiter;
+	int awoken = 0;
+	ps_list_foreach_d(&futex->waiters, waiter) {
+		if (awoken >= wakeup_count) {
+			return 0;
+		} else {
+			sl_thd_wakeup(waiter->thdid);
+			awoken += 1;
+		}
+	}
+	return awoken;
+}
+
+int
+cos_futex(int *uaddr, int op, int val,
+          const struct timespec *timeout, /* or: uint32_t val2 */
+		  int *uaddr2, int val3)
+{
+	int result = 0;
+	struct futex_data *futex;
+
+	sl_lock_take(&futex_lock);
+
+	if (op & FUTEX_PRIVATE) {
+		op ^= FUTEX_PRIVATE;
+	}
+	assert(!(op & FUTEX_CLOCK_REALTIME));
+
+	futex = lookup_futex(uaddr);
+	switch (op)
+	{
+		case FUTEX_WAIT:
+			result = cos_futex_wait(futex, uaddr, val, timeout);
+			if (result != 0) {
+				errno = result;
+				result = -1;
+			}
+			break;
+		case FUTEX_WAKE:
+			result = cos_futex_wake(futex, val);
+			break;
+		default:
+			printc("Unsupported futex operation");
+			assert(0);
+	}
+
+	sl_lock_release(&futex_lock);
+	return result;
 }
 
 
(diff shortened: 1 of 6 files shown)


src/components/lib/posix/posix.c
```diff
@@ -198,24 +197,30 @@ cos_tkill(int tid, int sig)
 	return 0;
 }
 
+static inline microsec_t
+time_to_microsec(const struct timespec *t)
+{
+	time_t seconds    		= t->tv_sec;
+	long nano_seconds 		= t->tv_nsec;
+	microsec_t microseconds = seconds * 1000000 + nano_seconds / 1000;
+	return microseconds;
```
phanikishoreg: separate declarations from `return`. Also, spaces in declaration needs to be fixed.

src/components/lib/posix/posix.c
```diff
@@ -293,6 +297,149 @@ cos_clone(int (*func)(void *), void *stack, int flags, void *arg, pid_t *ptid, v
 	return thd->thdid;
 }
 
+#define FUTEX_WAIT		0
```
phanikishoreg: Is it possible to move all the futex implementation to a separate file?

src/components/lib/posix/posix.c
```diff
@@ -293,6 +297,149 @@ cos_clone(int (*func)(void *), void *stack, int flags, void *arg, pid_t *ptid, v
...
+
+#define FUTEX_COUNT 20
+struct futex_data futexes[FUTEX_COUNT];
+
+struct futex_data *
+lookup_futex(int* uaddr)
+{
+	int last_free = -1;
+	int i;
+	for (i = 0; i < FUTEX_COUNT; i++) {
```
phanikishoreg: blank like after declarations!

src/components/lib/posix/posix.c
```diff
@@ -293,6 +297,149 @@ cos_clone(int (*func)(void *), void *stack, int flags, void *arg, pid_t *ptid, v
...
+			.uaddr = uaddr
+		};
+		ps_list_head_init(&futexes[last_free].waiters);
+		return &futexes[last_free];
+	}
+	printc("Out of futex ids!");
+	assert(0);
+}
+
+// TODO: Cleanup empty futexes
```
phanikishoreg: comment style.

src/components/lib/posix/posix.c
```diff
@@ -293,6 +297,149 @@ cos_clone(int (*func)(void *), void *stack, int flags, void *arg, pid_t *ptid, v
...
+
+int
+cos_futex_wait(struct futex_data *futex, int *uaddr, int val, const struct timespec *timeout)
+{
+	int not_awoken;
+	cycles_t   deadline;
+	microsec_t wait_time       = 0;
+	struct futex_waiter waiter = (struct futex_waiter) {
+		.thdid = sl_thdid()
+	};
```
phanikishoreg: should probably indent lines 364,365 further to line up after `=`.

src/components/lib/posix/posix.c
```diff
@@ -293,6 +297,149 @@ cos_clone(int (*func)(void *), void *stack, int flags, void *arg, pid_t *ptid, v
...
+
+	if (*uaddr != val) {
+		return EAGAIN;
+	}
```
phanikishoreg: `if (*uaddr != val) return EAGAIN;` in one line would look better.

src/components/lib/posix/posix.c
```diff
@@ -293,6 +297,149 @@ cos_clone(int (*func)(void *), void *stack, int flags, void *arg, pid_t *ptid, v
...
+
+	ps_list_init_d(&waiter);
+	ps_list_head_append_d(&futex->waiters, &waiter);
+
+	if (timeout != NULL) {
+		wait_time = time_to_microsec(timeout);
+	}
```
phanikishoreg: one line conditional may look better.

src/components/lib/posix/posix.c
```diff
@@ -293,6 +297,149 @@ cos_clone(int (*func)(void *), void *stack, int flags, void *arg, pid_t *ptid, v
...
+
+	// No race here, we'll enter the awoken state if things go wrong
```
phanikishoreg: comment style. use `/* ... */` for single line comments.

src/components/lib/posix/posix.c
```diff
@@ -293,6 +297,149 @@ cos_clone(int (*func)(void *), void *stack, int flags, void *arg, pid_t *ptid, v
...
+	sl_lock_release(&futex_lock);
```
phanikishoreg: Why `release` and then `lock`? where is the lock taken before? could use some comments.

src/components/lib/posix/posix.c
```diff
@@ -293,6 +297,149 @@ cos_clone(int (*func)(void *), void *stack, int flags, void *arg, pid_t *ptid, v
...
+	if (wait_time == 0) {
+		sl_thd_block(0);
+	} else {
+		deadline = sl_now() + sl_usec2cyc(wait_time);
+		sl_thd_block_timeout(0, deadline);
+	}
+	sl_lock_take(&futex_lock);
+	ps_list_rem_d(&waiter);
+	return 0;
```
phanikishoreg: blank line before return

src/components/lib/posix/posix.c
```diff
@@ -293,6 +297,149 @@ cos_clone(int (*func)(void *), void *stack, int flags, void *arg, pid_t *ptid, v
...
+}
+
+int cos_futex_wake(struct futex_data *futex, int wakeup_count)
+{
+	struct futex_waiter *waiter;
+	int awoken = 0;
```
phanikishoreg: blank line after decl.

src/components/lib/posix/posix.c
```diff
@@ -293,6 +297,149 @@ cos_clone(int (*func)(void *), void *stack, int flags, void *arg, pid_t *ptid, v
...
+	sl_lock_take(&futex_lock);
+
+	if (op & FUTEX_PRIVATE) {
+		op ^= FUTEX_PRIVATE;
+	}
+	assert(!(op & FUTEX_CLOCK_REALTIME));
+
+	futex = lookup_futex(uaddr);
+	switch (op)
+	{
```
phanikishoreg: should be `switch (op) {`.

src/components/lib/posix/posix.c
```diff
@@ -293,6 +297,149 @@ cos_clone(int (*func)(void *), void *stack, int flags, void *arg, pid_t *ptid, v
...
+			break;
+		case FUTEX_WAKE:
+			result = cos_futex_wake(futex, val);
+			break;
+		default:
+			printc("Unsupported futex operation");
+			assert(0);
+	}
+
+	sl_lock_release(&futex_lock);
```
phanikishoreg: blank line before return.

src/components/lib/posix/posix.c
```diff
@@ -337,6 +484,7 @@ syscall_emulation_setup(void)
 	libc_syscall_override((cos_syscall_t)cos_set_thread_area, __NR_set_thread_area);
 	libc_syscall_override((cos_syscall_t)cos_set_tid_address, __NR_set_tid_address);
 	libc_syscall_override((cos_syscall_t)cos_clone, __NR_clone);
+	libc_syscall_override((cos_syscall_t)cos_futex, __NR_futex);
```
phanikishoreg: I feel like all the system call names should be `cos_syscall_xxxx`. what do you think?

Others: Maybe `syscall_XXX` would be better? But I honestly think the current names are fine.

src/components/lib/posix/posix.c
Others: The lock is taken in `cos_futex`, will comment

src/components/lib/posix/posix.c
Others: It's difficult with the current way libposix is setup as an object library, maybe doable though?

src/components/lib/posix/posix.c
Others: I tried this and it looks really bad... I'm going to leave it as is, unless there is a really compelling argument for it.

src/components/lib/posix/posix.c
Others: Tried the other way, prefer the way it is. I'm open to changing if necessary though.

src/components/lib/posix/posix.c
phanikishoreg: ah, I see. the build system auto-generates .a files, so we'd have to generate special rules for having everything in a separate .c file. 

Another easy and better way is perhaps to have a "futex.h" header within "posix/" with all the implementation in it and include "futex.h" in posix.c (with some warning if already included). 

Or, at the very least, we could have the "#define", "struct"s in a separate header.

src/components/lib/posix/posix.c
phanikishoreg: kk. consider: `sl_thd_block_no_cs()`.

src/components/lib/posix/posix.c
Others: Resolved

src/components/lib/posix/posix.c
Others: It's actually more complex than that, because `libposix` is an object library... It isn't an easy fix, so I'm just going to leave it for now -- perhaps a future commit could fix this.

src/components/lib/posix/posix.c
phanikishoreg: yeah, i'm not very sure either.
But my feeling is that having a namespace for `posix` api and it's internals would probably be better. `cos_` is a generic namespace for everything in composite.

src/components/lib/posix/posix.c
```diff
@@ -293,6 +298,152 @@ cos_clone(int (*func)(void *), void *stack, int flags, void *arg, pid_t *ptid, v
...
+	}
+	printc("Out of futex ids!");
+	assert(0);
+}
+
+/* TODO: Cleanup empty futexes */
+
+struct sl_lock futex_lock = SL_LOCK_STATIC_INIT();
+
+int
```
gparmer: Comment required:

```
/* 
 * precondition: futex_lock is taken
 * Note that this function releases and re-takes that lock, thus critical section state must be re-checked after this returns.
 */
```

If you can add an assertion to check that we have the lock in this function, that is better than the `precondition` above.

I'm aware that none of this matters right now.  It might in the future if the futex implementation evolves, thus this comment.

src/components/lib/posix/posix.c
```diff
@@ -293,6 +298,152 @@ cos_clone(int (*func)(void *), void *stack, int flags, void *arg, pid_t *ptid, v
...
+          const struct timespec *timeout, /* or: uint32_t val2 */
+		  int *uaddr2, int val3)
+{
+	int result = 0;
+	struct futex_data *futex;
+
+	sl_lock_take(&futex_lock);
+
+	if (op & FUTEX_PRIVATE) {
+		op ^= FUTEX_PRIVATE;
```
gparmer: Isn't this condition and logic the same as `op &= ~FUTEX_PRIVATE`?  If we can avoid a branch, we should.

src/components/lib/posix/posix.c
```diff
@@ -293,6 +298,152 @@ cos_clone(int (*func)(void *), void *stack, int flags, void *arg, pid_t *ptid, v
...
+	/* No race here, we'll enter the awoken state if things go wrong */
+	sl_lock_release(&futex_lock);
+	if (wait_time == 0) {
+		sl_thd_block(0);
+	} else {
+		deadline = sl_now() + sl_usec2cyc(wait_time);
+		sl_thd_block_timeout(0, deadline);
+	}
+	sl_lock_take(&futex_lock);
+	ps_list_rem_d(&waiter);
```
gparmer: `waiter` should likely have been removed in `wake` rather than here.  If it is not removed in `wake`, we can have multiple wakes for the same thread which is not really what we want.

src/components/lib/posix/posix.c
```diff
@@ -293,6 +298,152 @@ cos_clone(int (*func)(void *), void *stack, int flags, void *arg, pid_t *ptid, v
...
+int cos_futex_wake(struct futex_data *futex, int wakeup_count)
+{
+	struct futex_waiter *waiter;
+	int awoken = 0;
+
+	ps_list_foreach_d(&futex->waiters, waiter) {
+		if (awoken >= wakeup_count) {
+			return awoken;
+		} else {
+			sl_thd_wakeup(waiter->thdid);
```
gparmer: can remove from the list here.

Others: @gparmer Can I? Won't that invalidate the iterator I'm using?

src/components/lib/posix/posix.c
Others: @gparmer It already will assert, although that is subtle. (When we try and release the lock, there is an assert that we hold it.) Will add a comment.

src/components/lib/posix/posix.c
```diff
@@ -293,6 +298,152 @@ cos_clone(int (*func)(void *), void *stack, int flags, void *arg, pid_t *ptid, v
...
+{
+	thdid_t thdid;
+	struct ps_list list;
+};
+
+#define FUTEX_COUNT 20
+struct futex_data futexes[FUTEX_COUNT];
+
+struct futex_data *
+lookup_futex(int* uaddr)
```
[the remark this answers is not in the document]
Others: @gparmer yes, it can be any address

src/components/lib/posix/posix.c
```diff
@@ -198,24 +197,31 @@ cos_tkill(int tid, int sig)
 	return 0;
 }
 
+static inline microsec_t
+time_to_microsec(const struct timespec *t)
+{
+	time_t seconds          = t->tv_sec;
+	long nano_seconds       = t->tv_nsec;
+	microsec_t microseconds = seconds * 1000000 + nano_seconds / 1000;
```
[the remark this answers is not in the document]
Others: @gparmer It is 64 bits
```
src/kernel/include/shared/cos_types.h
27:typedef u64_t         microsec_t;
```

src/components/lib/posix/posix.c
Others: @gparmer Yes, but it is in a very private header. (We probably could access it with some hacky include path, but I didn't think it was worth it.)

src/components/lib/posix/posix.c
gparmer: `ps_for_each_del`

PR #298, opened 2017-11-29
Update llbooter boot deps with sl caps

maloneya:
Added additional capabilities to boot_newcomp_create in boot_deps to allow for the creation of components that can make SL calls.
Fixed bug where newly created components cap_frontier was being set to reserved values

**Add description here.**

### Intent for your PR

Choose one (Mandatory):

### Reviewers (Mandatory):

### Code Quality

### Testing

-

### DIFF
--- a/src/components/implementation/no_interface/llbooter/boot_deps.h
+++ b/src/components/implementation/no_interface/llbooter/boot_deps.h
@@ -5,6 +5,7 @@
 #include <llprint.h>
 
 #define UNDEF_SYMBS 64
+#define CMP_UTYPMEM_SZ (1 << 26)
 
 /* Assembly function for sinv from new component */
 extern void *__inv_test_entry(int a, int b, int c);
@@ -68,7 +69,7 @@ boot_compinfo_init(spdid_t spdid, captblcap_t *ct, pgtblcap_t *pt, u32_t vaddr)
 	assert(*pt);
 
 	new_comp_cap_info[spdid].compinfo = &new_compinfo[spdid];
-	cos_compinfo_init(new_comp_cap_info[spdid].compinfo, *pt, *ct, 0, (vaddr_t)vaddr, 4, &boot_info);
+	cos_compinfo_init(new_comp_cap_info[spdid].compinfo, *pt, *ct, 0, (vaddr_t)vaddr, BOOT_CAPTBL_FREE, &boot_info);
 }
 
 static void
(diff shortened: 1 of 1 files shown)


src/components/implementation/no_interface/llbooter/boot_deps.h
```diff
@@ -68,7 +69,7 @@ boot_compinfo_init(spdid_t spdid, captblcap_t *ct, pgtblcap_t *pt, u32_t vaddr)
 	assert(*pt);
 
 	new_comp_cap_info[spdid].compinfo = &new_compinfo[spdid];
-	cos_compinfo_init(new_comp_cap_info[spdid].compinfo, *pt, *ct, 0, (vaddr_t)vaddr, 4, &boot_info);
+	cos_compinfo_init(new_comp_cap_info[spdid].compinfo, *pt, *ct, 0, (vaddr_t)vaddr, BOOT_CAPTBL_FREE, &boot_info);
```
gparmer: This is slightly inefficient in the case where we have a component with only sinv capabilities.  Now we're starting to allocate them at a higher place in the captable than is necessary.  I'm not sure how bad this is, but to figure out, we'll just need to do the math on how many capabilities are in the first captbl node from BOOT_CAPTBL_FREE to the end of the first node in the capability table.

gparmer: I would assume that we'd want two ways to create components (i.e. parameterize this function by which starting capability image it uses.

src/components/implementation/no_interface/llbooter/boot_deps.h
```diff
@@ -125,6 +127,20 @@ boot_newcomp_create(spdid_t spdid, struct cos_compinfo *comp_info)
 
 	boot_newcomp_sinv_alloc(spdid);
 
+	vmutpt = cos_pgtbl_alloc(&boot_info);
```
gparmer: 130-140 looks like a limited range of the capabilities that you're adding in here.  Any reason you chose these, and not others?  (I assume because this was "sufficient to get it working", but I'm curious if there was more thought put into it.)

maloneya:
1. From my tests it works fine, i am able to boot a skeleton component along with a resource manager, and that skeleton component is able to invoke the resource manager. 

2. This makes sense to me, i just assumed the work would be done from this commit.

src/components/implementation/no_interface/llbooter/boot_deps.h
maloneya: This was purely based on what was sufficient, literally added caps till it booted... 

would be interested to discuss (forgot to bring this up on friday) what the full capability range actually is

src/components/implementation/no_interface/llbooter/boot_deps.h
maloneya: once we determine where the set of non resource managers ends in the cap table should probably add another value to (COMP_CAPTBL_FREE?) the enum in cos_types so were not passing magic numbers like 4.

PR #303, opened 2018-01-22
Add some missing flags to build on x64

pingerino:
This does not actually build all the way on x64 - but it gets a lot
further.

I attempted to get composite to build on my x64 Linux. I didn't get it completely - the linker app doesn't build on 64bit, and apparently I can't install a copy of 32 bit binutils along side 64 bit, so that's where I got stuck.

I ended up building a trusty chroot following these instructions: https://jblevins.org/log/ubuntu-chroot, which might be useful to link in your docs on building composite, as it worked out of the box.

Not sure if you are keen to merge these fixes, but I though I'll let the team decide rather than throw them away.

### Intent for your PR

Choose one (Mandatory):

### Reviewers (Mandatory):

Well I have met @gparmer once at RTSS. 

### Code Quality

N/A it's just some build flags

### Testing

- make

### DIFF
--- a/src/components/implementation/Makefile.subsubdir
+++ b/src/components/implementation/Makefile.subsubdir
@@ -74,7 +74,7 @@ $(COMPONENT): $(OBJS)
 #	$(info |     [LD]   $@: Loading stubs and cos libraries $(if $(strip $(LIB_LIBRARIES)),including $(LIB_LIBRARIES)) into $<)
 	$(info |     [LD]   $(COMPNAME) ($@): Linking $(if $(strip $(LIB_LIBRARIES)),$(LIB_LIBRARIES)) $^)
 	@$(LD) $(LDFLAGS) -r -o $@.$(TMP_STR) $^ $(IF_SERV_STUBS) $(IF_CLIENT_STUBS) $(LIB_ASM_MANDITORY) $(LIB_FLAGS)
-	@$(MUSLCC) $@.$(TMP_STR) $(MUSLCFLAGS) -Xlinker -r -o $@.$(TMP_STR2)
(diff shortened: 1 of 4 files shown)


RobertGiff:
Hi @pingerino, I ran into the exact same issue this previous week. I also found it exceptionally frustrating that 32bit binutils doesn't work along side 64bit. More than that, I also wasn't able to get *only* 32bit binutils installed as a standalone. I was running into an issue with installing further dependencies that simply didn't exist according to the package manager and I really didn't want to hunt them down in source. This pull request I believe is great regardless (I have a very similar PR that has been pending on my TODO these last view days), so I'm happy to see it someone else submitted it :)
You mentioned you worked on trust 14.04. These flags that you added are useful in helping Composite compile and execute on 16.04 32bit, which is nice for the updated gcc.

pingerino:
@RobertGiff I tried building on 32bit xenial but also ran into package issues there so set up a trusty chroot instead.

RobertGiff:
@pingerino do you remember the package issues that you were experiencing on 32bit Xenial? That is the current set up that I am using.

gparmer:
@pingerino Great to see the PR Anna!  I've missed seeing you at conferences since we met (was it in Rome?).  First, a high-level comment:  If you are having issue with anything, feel free to be active here, and we'll help out.  We have a very narrow way to get everything up and running, and it is not very approachable yet.

Someone did have everything running on a 64 bit system before, so I know it is possible, but indeed it does not work by default.  The long-term solution is to port the system, but that is currently off the fast-path.  @RobertGiff ran into the same problem with 32-bit binutils.  I propose that we might want to just make the place where this is broken with 64 bit binutils (cos_info structures) have explicit sizing for all its fields (using the u32 etc sizes).  We are currently working to replace the current linking infrastructure, but that won't be ready for a bit, unfortunately.

Regardless, this is a great patch, thanks!

PR #309, opened 2018-02-06
Thread initalization and Ubuntu 32bit 16.04 compilation updates

RobertGiff:

This PR is 2 fold:
1. Defer initial thread creation from the booter to higher level schedulers
a. This involves the addition of a cos_kernel_api functionality to invoke the booter for various management resources
2. Update the build system to enable and remove warnings for compiling on Ubuntu 16.04 LTS 32bit.
a. NOTE: There are just a couple of compilation warnings left when running `make init`. These stem from the C++ library as some functions have become deprecated. As I don't know C++, I didn't attempt to resolve those warnings.

NOTE: The previous runscripts llbooter_poing.sh and llbooter_test.sh have been slightly updated within this PR to enable thread initialization within the booter. We simply take advantage of @maloneya `sl_` object prefix to enable the threads "special privledge" to have their init threads created by the booter.

### Intent for your PR

Choose one (Mandatory):

- [X] This PR is for a code-review and is intended to get feedback, but not to be pulled yet.

### Reviewers (Mandatory):

@gparmer, @lab176, @maloneya, @Others, @phanikishoreg 

### Code Quality

### Testing

- llbooter_test.sh
- rumpkernboot.sh (*rumpkernel branch only*)


src/components/lib/cos_kernel_api.c
```diff
@@ -640,6 +644,79 @@ cos_sinv(sinvcap_t sinv, word_t arg1, word_t arg2, word_t arg3, word_t arg4)
 	return call_cap_op(sinv, 0, arg1, arg2, arg3, arg4);
 }
 
+/*
+ * TODO:
+ * talk about this naming choice
+ * can't remember what Gabe thought was best choice here
+ */
+void *
```
phanikishoreg: I'd prefer it to be part of a llbooter_lib. 
Just breaks the abstraction in `cos_kernel_api`!! 
IMO `hypervisor` is a misnomer.
Just `llbooter_res_get()` would do it if we make a small library for that with the llbootlib.h defining those enumerations and just this API.

Even better:
Because all these are making sync invs to the booter, why not make llbooter interface with different functions and let the loader make them for us?

src/components/lib/musl-1.1.11/src/internal/cos_syscalls.c
```diff
@@ -1,7 +1,7 @@
 __attribute__((weak, regparm(1))) long
 __cos_syscall(int syscall_num, long a, long b, long c, long d, long e, long f, long g)
 {
-        long *x = *(long*)0;
+        long *x = (long*)0;
```
phanikishoreg: why not just `long *x = NULL`?

src/kernel/include/shared/cos_types.h
```diff
@@ -40,7 +50,16 @@ typedef u32_t         sched_tok_t;
  */
 #define TCAP_TIME_QUANTUM_ORD 12
 #define TCAP_TIME_MAX_ORD (TCAP_TIME_QUANTUM_ORD + (sizeof(tcap_time_t) * 8))
-#define TCAP_TIME_MAX_BITS(c) ((c >> TCAP_TIME_MAX_ORD) << TCAP_TIME_MAX_ORD)
+/*
+ * FIXME
+ * TCAP_TIME_MAX_ORD comes out as 44, thus shifting by this much on a 32 bit
+ * system leads to warnings that shift is greater than the width of the type.
```
phanikishoreg: Explicitly typecasting to u64_t or even cycles_t should fix the warning right?

src/components/Makefile.comp
```diff
@@ -40,7 +40,7 @@ OPT= -g
 #OPT= -O3
 CFLAGS=-m32 -D__x86__ -Wall -Wextra -Wno-unused-parameter -Wno-unused-function -fno-stack-protector -fno-omit-frame-pointer -Wno-unused-variable -fvar-tracking $(INC_PATH) $(MUSLINC) $(LWIPINC) $(LUAINC) $(OPT) $(SHARED_FLAGS)
 CXXFLAGS=-fno-exceptions -fno-threadsafe-statics -Wno-write-strings $(CFLAGS)
-LDFLAGS=-m elf_i386
+LDFLAGS=-melf_i386
```
Others: Is this a useful/meaningful diff?

src/components/implementation/Makefile.subsubdir
```diff
@@ -74,7 +74,7 @@ $(COMPONENT): $(OBJS)
 #	$(info |     [LD]   $@: Loading stubs and cos libraries $(if $(strip $(LIB_LIBRARIES)),including $(LIB_LIBRARIES)) into $<)
 	$(info |     [LD]   $(COMPNAME) ($@): Linking $(if $(strip $(LIB_LIBRARIES)),$(LIB_LIBRARIES)) $^)
 	@$(LD) $(LDFLAGS) -r -o $@.$(TMP_STR) $^ $(IF_SERV_STUBS) $(IF_CLIENT_STUBS) $(LIB_ASM_MANDITORY) $(LIB_FLAGS)
-	@$(MUSLCC) $@.$(TMP_STR) $(MUSLCFLAGS) -Wl,-melf_i386 -Xlinker -r -o $@.$(TMP_STR2)
+	@$(MUSLCC) $@.$(TMP_STR) $(MUSLCFLAGS) -Xlinker -melf_i386 -o $@.$(TMP_STR2)
```
Others: Why do we remove -r here?

src/components/implementation/no_interface/events_prototype/Makefile
```diff
@@ -6,10 +6,15 @@ DEPENDENCIES=
...
-export XARGO_RUST_SRC=$(shell rustc --print sysroot)/lib/rustlib/src/rust/src/
-export RUSTFLAGS=-L ../../../../lib/musl-1.1.11/lib/ -L lib
 CRATE_DIR=events
 
+XARGO_RUST_SRC=$(shell rustc --print sysroot)/lib/rustlib/src/rust/src/
+RUSTFLAGS=-L $(shell pwd)/../../../lib/musl-1.1.11/lib/ -L $(shell pwd)/$(CRATE_DIR)/lib
+RUST_TARGET_PATH=$(shell pwd)/$(CRATE_DIR)
+export XARGO_RUST_SRC
+export RUSTFLAGS
+export RUST_TARGET_PATH
```
Others: Are you sure these modifications are right? Have you tested these changes on ZSH Ubuntu 16, bash Ubuntu 16, and bash Ubuntu 14? Sorry to be nitpicky, but I don't want Rust to be broken on any platform. 

Also, why split up the exports onto their own lines?

src/components/implementation/no_interface/llbooter/boot_deps.h
```diff
@@ -94,73 +106,242 @@ boot_newcomp_sinv_alloc(spdid_t spdid)
...
 {
-	printc("Booter: done creating system.\n\n");
-	cos_thd_switch(schedule[sched_cur]);
+	printc("Booter: done creating system.\n");
+	printc("Booter: Starting SL event loop\n\n");
+	sl_sched_loop();
+}
+
+static int
+boot_check_scheduler(char *comp_name)
```
Others: This hack is gross, but probably fine for now. We should fix this/put this information in the bootscript in the future.

src/components/implementation/no_interface/llbooter/boot_deps.h
```diff
@@ -94,73 +106,242 @@ boot_newcomp_sinv_alloc(spdid_t spdid)
...
+{
+	int i;
+	char *prefix = "sl_";
+
+	for (i = 0; i < 3; i++) {
+		if (comp_name[i] == '\0' || comp_name[i] != prefix[i]) 
+			return 0;
+	}
```
Others: Can't you just do `strncmp`?

src/components/implementation/no_interface/llbooter/boot_deps.h
```diff
@@ -94,73 +106,242 @@ boot_newcomp_sinv_alloc(spdid_t spdid)
...
+static int
+boot_check_shdmem(char *comp_name)
+{
+	int i;
+	char *prefix = "shmem";
+
+	for (i = 0; i < 5; i++) {
+		if (comp_name[i] == '\0' || comp_name[i] != prefix[i])
+			return 0;
+	}
```
Others: `strncmp`?

src/components/implementation/no_interface/llbooter/llbooter.c
```diff
@@ -211,8 +207,8 @@ boot_comp_map_populate(struct cobj_header *h, spdid_t spdid, vaddr_t comp_info)
 		}
 
 		if (sect->flags & COBJ_SECT_CINFO) {
-			assert(left == PAGE_SIZE);
-			assert(comp_info == dest_daddr);
+			assert((left % PAGE_SIZE) == 0);
+			assert(comp_info == (dest_daddr + (((left/PAGE_SIZE)-1)*PAGE_SIZE)));
```
Others: Oooh, this might fix an off by 1 error I'm having w/ Alex's code...

src/components/implementation/no_interface/llbooter/llbooter.c
```diff
@@ -306,8 +299,9 @@ boot_init_ndeps(int num_cobj)
 	int i = 0;
 
 	printc("MAX DEPS: %d\n", MAX_DEPS);
-	for (i = 0; i < deps_list[i].server; i++) {
-//		if (deps_list[i].client != 0) printc("client: %d, server: %d \n", deps_list[i].client, deps_list[i].server);
+	for (i = 0; (short int)i < deps_list[i].server; i++) {
+		/* TODO, why is this commented? Do we need this? */
+		//if (deps_list[i].client != 0) printc("client: %d, server: %d \n", deps_list[i].client, deps_list[i].server);
```
Others: Does someone tagged on this PR have the context to fix this? @gparmer maybe?

src/components/include/cos_kernel_api.h
```diff
@@ -156,6 +156,12 @@ int cos_introspect(struct cos_compinfo *ci, capid_t cap, unsigned long op);
 
 int cos_sinv(sinvcap_t sinv, word_t arg1, word_t arg2, word_t arg3, word_t arg4);
 
+/*
+ * Calls down to the booter to access necessary resources that are needed for
+ * higher level resource managers
+ */
+void * cos_hypervisor_get_resource(int op, void *arg1, void *arg2, void *arg3);
```
Others: Any docs on how to use this function? (also the font is weird here with the `_` and the `g` looking really close together. Funky...)

src/components/lib/cos_kernel_api.c
```diff
@@ -640,6 +644,79 @@ cos_sinv(sinvcap_t sinv, word_t arg1, word_t arg2, word_t arg3, word_t arg4)
...
+					(int)cap_index);
+			ret = (void *)cap_index;
+			break;
+		}
+		default:
+			assert(0);
+	}
+
+	return ret;
+}
```
Others: Why is this one big function? Why can't it be broken down into different ones, for ease of use?

src/kernel/capinv.c
```diff
@@ -930,6 +930,8 @@ cap_introspect(struct captbl *ct, capid_t capid, u32_t op, unsigned long *retval
 		return thd_introspect(((struct cap_thd *)ch)->t, op, retval);
 	case CAP_TCAP:
 		return tcap_introspect(((struct cap_tcap *)ch)->tcap, op, retval);
+	default:
+		return -EINVAL;
 	}
 	return -EINVAL;
```
Others: What's the point having `default: return -EINVAL;` directly before `return -EINVAL`?

src/kernel/include/shared/cos_types.h
```diff
@@ -40,7 +50,16 @@ typedef u32_t         sched_tok_t;
...
+ * As we shift right by this amount, it zeros out the number. To avoid
+ * warnings I have temporarily replaced this initial shift by 32 bits.
+ * Fix when I know the original purpose of this macro.
+ */
+//#define TCAP_TIME_MAX_BITS(c) ((c >> TCAP_TIME_MAX_ORD) << TCAP_TIME_MAX_ORD)
+#define TCAP_TIME_MAX_BITS(c) ((c >> 32) << 32)
```
Others: @gparmer Do you have the context for this? I feel like merging an unresolved todo kinda sucks. (Also there's a // comment here Robbie.)

src/platform/i386/runscripts/llbooter_pong.sh
```diff
@@ -1,4 +1,8 @@
 #!/bin/sh
 
 cp llboot_test.o llboot.o
-./cos_linker "llboot.o, ;ppong.o, ;ppong_two.o, ;pingp.o, ;pingp_two.o, :pingp.o-ppong.o;pingp_two.o-ppong_two.o" ./gen_client_stub
+cp ppong.o       sl_ppong.o
+cp ppong_two.o   sl_ppong_two.o
+cp pingp.o       sl_pingp.o
+cp pingp_two.o   sl_pingp_two.o
+./cos_linker "llboot.o, ;sl_ppong.o, ;sl_ppong_two.o, ;sl_pingp.o, ;sl_pingp_two.o, :sl_pingp.o-sl_ppong.o;sl_pingp_two.o-sl_ppong_two.o" ./gen_client_stub
```
Others: Elegant boot script here...

src/components/implementation/no_interface/llbooter/boot_deps.h
```diff
@@ -94,73 +106,242 @@ boot_newcomp_sinv_alloc(spdid_t spdid)
...
+			 */
+			boot_thd_done();
+			break;
+		case REQ_PGTBL_CAP:
+			/* arg1: dst comp, arg2: src comp, arg3: cap_slot */
+			boot_pgtbl_cap_transfer((int)arg1, (int)arg2, (int)arg3);
+			ret = (void *)0;
+			break;
+		case REQ_NUM_COMPS:
+			ret = (void *)num_cobj;
```
gparmer: Do we want to return the count *including* the scheduler?  Isn't the scheduler asking for how many *other* components there are?

src/components/implementation/no_interface/llbooter/boot_deps.h
```diff
@@ -94,73 +106,242 @@ boot_newcomp_sinv_alloc(spdid_t spdid)
...
+		case REQ_SINV_CAP:
+			printc("Sinv cap request not implemented\n");
+			break;
+		case REQ_CAP_FRONTIER:
+			/* arg1: spdid */
+			ret = (void *)new_comp_cap_info[(int)arg1].compinfo->cap_frontier;
+			break;
+		case REQ_COMP_CAP:
+			/* arg1: dst comp spdid, arg2: src comp spdid, arg3: cap_slot */
+			boot_comp_cap_transfer((int)arg1, (int)arg2, (int)arg3);
```
gparmer: Again with propagating return values.

src/components/implementation/no_interface/llbooter/boot_deps.h
```diff
@@ -94,73 +106,242 @@ boot_newcomp_sinv_alloc(spdid_t spdid)
...
+			ret = (void *)0;
+			break;
+		default:
+			printc("op: %d not supported!\n", op);
+			assert(0);
```
gparmer: nonono

src/components/implementation/no_interface/llbooter/llbooter.c
```diff
@@ -292,8 +280,13 @@ boot_create_cap_system(void)
 		if (boot_spd_inv_cap_alloc(h, spdid)) BUG();
 		if (boot_comp_map(h, spdid, ci, pt)) BUG();
 
-		boot_newcomp_create(spdid, new_comp_cap_info[spdid].compinfo);
-		printc("\nComp %d (%s) created @ %x!\n\n", h->id, h->name, sect->vaddr);
+		/* check for hardcoded "sl_" prefix in c obj to determine which cap image we create */
+		is_sched  = boot_check_scheduler(h->name);
```
gparmer: Really feels like these should just be flags (bitmap) associated with a scheduler.
```
typedef enum {
    BOOT_COMP_FLAG_SCHED = 1,
    BOOT_COMP_FLAG_MM = 1<<1,
} boot_comp_flag_t;
```

src/components/implementation/no_interface/llbooter/llbooter.c
```diff
@@ -306,8 +299,9 @@ boot_init_ndeps(int num_cobj)
 	int i = 0;
 
 	printc("MAX DEPS: %d\n", MAX_DEPS);
-	for (i = 0; i < deps_list[i].server; i++) {
-//		if (deps_list[i].client != 0) printc("client: %d, server: %d \n", deps_list[i].client, deps_list[i].server);
+	for (i = 0; (short int)i < deps_list[i].server; i++) {
```
gparmer: Yeah, remove this.  Missed it in a previous PR.

src/components/include/cos_kernel_api.h
gparmer: `cos_hypervisor_get_resource` -> `cos_hypervisor_hypercall`???

src/components/include/cos_kernel_api.h
```diff
@@ -178,6 +184,8 @@ int cos_tcap_transfer(tcap_t src, arcvcap_t dst, tcap_res_t res, tcap_prio_t pri
 int cos_tcap_delegate(asndcap_t dst, tcap_t src, tcap_res_t res, tcap_prio_t prio, tcap_deleg_flags_t flags);
 int cos_tcap_merge(tcap_t dst, tcap_t rm);
 
+capid_t cos_capid_bumkp_alloc(struct cos_compinfo *ci, cap_t cap);
```
gparmer: What is this?  Don't see it anywhere else.

src/kernel/include/shared/cos_types.h
phanikishoreg: I left reviewing this PR in the middle. 
I think explicit typecasting it to 64bit should fix the compilation warning.
Because if you look at it, we are really using that to make a 64bit unsigned int.

src/platform/linker/vdc.c
```diff
@@ -142,6 +144,9 @@ find_symbol_exporter_mark_resolved(struct symb *s, struct dependency *exporters,
 	return NULL;
 }
 
+struct service_symbs *find_symbol_exporter_mark_resolved(struct symb *s, struct dependency *exporters, int num_exporters,
+                                   struct symb **exported);
```
phanikishoreg: nitpick: I think it should line up with the first parameter in the first line.

src/components/lib/cos_kernel_api.c
gparmer: I agree that this seems out of place in `cos_kernel_api`.  Unfortunately our proxy for "requires low-level access to direct capabilities to resources" is "uses cos_kernel_api".  Perhaps the "right way" to handle this is the work that you're doing (@phanikishoreg) with the resource manager so that the component has access to *no* capabilities other than those given to it by the resource manager.

In the end, this call is a hack.  It is likely a hack that solves a problem with multiple booters are hierarchically arranged (i.e. *not* when we use the resource manager).  Given this, I think that @phanikishoreg is right:  This code should be in some part of the booter's library.

It cannot use an interface as the linker does not look at the booter's API to satisfy sinvs right now, and I don't want to hack that in.

src/components/lib/cos_kernel_api.c
```diff
@@ -640,6 +644,79 @@ cos_sinv(sinvcap_t sinv, word_t arg1, word_t arg2, word_t arg3, word_t arg4)
...
+ */
+void *
+cos_hypervisor_get_resource(int op, void *arg1, void *arg2, void *arg3)
+{
+	void *ret;
+	assert(op >= 0);
+
+	/* Dpending on the operation we will need to allocate our own a cap space */
+	switch(op) {
+		case INIT_DONE: {
```
gparmer: Again with wrong formatting.

src/components/lib/cos_kernel_api.c
```diff
@@ -640,6 +644,79 @@ cos_sinv(sinvcap_t sinv, word_t arg1, word_t arg2, word_t arg3, word_t arg4)
...
+ * talk about this naming choice
+ * can't remember what Gabe thought was best choice here
+ */
+void *
+cos_hypervisor_get_resource(int op, void *arg1, void *arg2, void *arg3)
+{
+	void *ret;
+	assert(op >= 0);
+
+	/* Dpending on the operation we will need to allocate our own a cap space */
```
gparmer: Don't you have spell check on in your editor for comments?  Or does `vim` not have that...

src/components/lib/cos_kernel_api.c
```diff
@@ -640,6 +644,79 @@ cos_sinv(sinvcap_t sinv, word_t arg1, word_t arg2, word_t arg3, word_t arg4)
 	return call_cap_op(sinv, 0, arg1, arg2, arg3, arg4);
 }
 
+/*
+ * TODO:
+ * talk about this naming choice
+ * can't remember what Gabe thought was best choice here
+ */
```
gparmer: Again, `void *` not so much.

src/components/lib/cos_kernel_api.c
```diff
@@ -640,6 +644,79 @@ cos_sinv(sinvcap_t sinv, word_t arg1, word_t arg2, word_t arg3, word_t arg4)
...
+		}
+		case REQ_PGTBL_CAP: {
+			/*
+			 * FIXME: There is no mechanism for checking to see if a comp
+			 * should be permitted to do this
+			 */
+			capid_t cap_index = cos_capid_bump_alloc((struct cos_compinfo *)arg3,
+					CAP_PGTBL);
+			assert(cap_index > 0);
+			cos_sinv(BOOT_CAPTBL_SINV_CAP, op, (int)arg1, (int)arg2,
```
gparmer: When you add the return value to this, you'll have to check it.

src/components/lib/cos_kernel_api.c
```diff
@@ -640,6 +644,79 @@ cos_sinv(sinvcap_t sinv, word_t arg1, word_t arg2, word_t arg3, word_t arg4)
...
+			 * FIXME: Is this sensitive information? Can someone use this
+			 * maliciously?
+			 */
+			ret = (void *)cos_sinv(BOOT_CAPTBL_SINV_CAP, op, (int)arg1,
+					(int)arg2, (int)arg3);
+			break;
+		}
+		case REQ_COMP_CAP: {
+			/*
+			 * FIXME: There is no mechanism for checking to see if a comp
```
gparmer: This should be a `FIXME` for the entire function, right?

src/components/lib/cos_kernel_api.c
```diff
@@ -640,6 +644,79 @@ cos_sinv(sinvcap_t sinv, word_t arg1, word_t arg2, word_t arg3, word_t arg4)
...
+			 * should be permitted to do this
+			 */
+			/* TODO: set each arg to a variable with a real name */
+			capid_t cap_index = cos_capid_bump_alloc((struct cos_compinfo *)arg3,
+					CAP_COMP);
+			assert(cap_index > 0);
+			cos_sinv(BOOT_CAPTBL_SINV_CAP, op, (int)arg1, (int)arg2,
+					(int)cap_index);
```
gparmer: Odd spacing for this.

src/components/lib/cos_kernel_api.c
```diff
@@ -640,6 +644,79 @@ cos_sinv(sinvcap_t sinv, word_t arg1, word_t arg2, word_t arg3, word_t arg4)
...
+			break;
+		}
+		case REQ_COMP_CAP: {
+			/*
+			 * FIXME: There is no mechanism for checking to see if a comp
+			 * should be permitted to do this
+			 */
+			/* TODO: set each arg to a variable with a real name */
+			capid_t cap_index = cos_capid_bump_alloc((struct cos_compinfo *)arg3,
+					CAP_COMP);
```
gparmer: odd spacing for tihs.

src/components/lib/musl-1.1.11/src/internal/cos_syscalls.c
gparmer: Do we not use `NULL` because we need to avoid a dependency on `stdlib.h`?  Try it out and see.

src/kernel/capinv.c
```diff
@@ -930,6 +930,8 @@ cap_introspect(struct captbl *ct, capid_t capid, u32_t op, unsigned long *retval
 		return thd_introspect(((struct cap_thd *)ch)->t, op, retval);
 	case CAP_TCAP:
 		return tcap_introspect(((struct cap_tcap *)ch)->tcap, op, retval);
+	default:
```
gparmer: Redundant with line `936`?

src/kernel/include/shared/cos_types.h
```diff
@@ -32,6 +32,16 @@ typedef u64_t         tcap_uid_t;
 typedef u32_t         sched_tok_t;
 #define PRINT_CAP_TEMP (1 << 14)
 
+/* Parameters for specifying resource requests from booter */
+typedef enum {
```
gparmer: Namespace should not be `REQ_`.  Should match the function name.  Perhaps `BOOT_HYP_`.

This should not be in `cos_types.h` which is for types that need to be shared between kernel and user-level (with some stupid exceptions).  If the functions above are in the booter library, then this should be there too.

The type name should match the namespace, e.g. `boot_hyp_` and should end with `_t`.  These mismatches are a little messy.

src/kernel/include/shared/cos_types.h
gparmer: @phanikishoreg you need to figure out the solution to this.  I don't remember the context.

src/components/implementation/no_interface/events_prototype/Makefile
gparmer: I don't think that @RobertGiff has any context for Rust.  I honestly (wrongly) assumed that you had provided these changes for him as they seem oddly specific.

src/components/implementation/Makefile.subsubdir
gparmer: I'm also not convinced that `-melf_i386` is the same as `W1,-melf_i386`.  Are we losing fidelity here?

src/components/Makefile.comp
RobertGiff: I've tested on my systems and for some reason I've found `-m elf_i386` to not work

RobertGiff: Also, why not stay consistent with `-m32`? `-melf_i386` certainly works so I see no reason to introduce a space when we aren't using one anywhere else in the above code

src/platform/i386/runscripts/llbooter_pong.sh
RobertGiff: The beauty of it is blinding, I know.

src/components/Makefile.comp
gparmer: Linker, assember, and compiler might take different arguments.  Just make sure it works on all systems.

src/components/lib/musl-1.1.11/src/internal/cos_syscalls.c
RobertGiff: It doesn't cause any warnings or errors. But it is strange to think that stdlib.h can be included in musl src code.

src/components/implementation/Makefile.subsubdir
RobertGiff: @Others `-r` is already in `MUSLCFLAGS`, no need to repeat it.

@gparmer it is not just `-melf_i386`, it is `-Xlinker -melf_i386`. GCC docs defines `-Xlinker` and `-Wl,` as both "Pass option as an option to the linker". The difference that I can discern between the two is that `-Xlinker` can be used to provide system-specific linker options that GCC does not recognize while `-Wl,` is able to pass multiple options to the linker by separating them with commas. We don't need that extra functionality here so I can switch it back to using `-Wl,` however we use `-Xlinker` in Makefile.comp, so I chose to go that route for consistency.

src/components/implementation/no_interface/events_prototype/Makefile
RobertGiff: I made these changes so that it would compile on 16.04 ZSH. @Others it does also work on 16.04 Bash. I unfortunately don't have a set up for Ubuntu 14 currently, if you pull and it doesn't work for some reason on Ubuntu 14, revert back to old form until a compatible fix is put in.

src/components/lib/cos_kernel_api.c
RobertGiff: We would have to allocate different `sinvs` for each function. By just having one, we are able to use a known specific offset in the captable. Adding more hardcoded sinvs and we would have to have more offsets instead of a single known one and this adds more setup complexity for a hack that will be changed anyway. I don't want to set up that dance for all components that need this interface.

src/kernel/capinv.c
RobertGiff: Added this due to new gcc compilation warnings about enum cases not being handled. Removed the second `return -EINVAL` outside of the switch statement.

src/components/implementation/no_interface/llbooter/Makefile
```diff
@@ -4,7 +4,7 @@ COMPONENT=llboot_test.o
 INTERFACES=
 DEPENDENCIES=
 IF_LIB=
-ADDITIONAL_LIBS=-lcobj_format -lcos_kernel_api
+ADDITIONAL_LIBS=-lcobj_format -lcos_kernel_api -lcos_defkernel_api -lsl -lheap -lsl_thd_static_backend -lsl_lock -lsl_mod_fprr
```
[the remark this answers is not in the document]
RobertGiff: This includes @maloneya PR which was never merged because of this reason. I can fix this in a commit that comes after the other changes I'm making right now.

src/components/implementation/no_interface/llbooter/boot_deps.h
```diff
@@ -3,37 +3,45 @@
...
+	vaddr_t                 vaddr_user_caps; // vaddr of user caps table in comp
+	vaddr_t                 addr_start;
+	vaddr_t                 vaddr_mapped_in_booter;
+	vaddr_t                 upcall_entry;
 } new_comp_cap_info[MAX_NUM_SPDS + 1];
 
-struct cos_compinfo boot_info;
-struct cos_compinfo new_compinfo[MAX_NUM_SPDS + 1];
-
-int                      schedule[MAX_NUM_SPDS + 1];
```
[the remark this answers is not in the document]
RobertGiff: That would make a lot of sense but assumes that this hierarchy is defined properly in the dependency order to pass to the linker correct? Would it make more sense to define this information as an explicit boot config that depend on the ordering of the linking components to be right? Then we would pass this information to the scheduler through the component configuration scheme we talked about previously. (Using the end of the heap to pass key-value pairs).

src/components/implementation/no_interface/llbooter/boot_deps.h
RobertGiff: yes good point. This is also used by the memmgr when needing to allocate caps for pgtbls

src/components/lib/cos_kernel_api.c
RobertGiff: kek

src/kernel/capinv.c
RobertGiff: Added this due to new gcc compilation warnings about enum cases not being handled. Removed the second `return -EINVAL` outside of the switch statement.

RobertGiff:
TODO: Remove SL from llbooter

src/components/implementation/no_interface/llbooter/boot_deps.h
```diff
@@ -159,66 +165,66 @@ boot_newschedcomp_cap_init(spdid_t spdid, captblcap_t ct, pgtblcap_t pt, compcap
...
 		boot_newschedcomp_cap_init(spdid, ct, pt, cc);
 
-		thd = sl_thd_comp_init(new_comp_cap_info[spdid].defcompinfo, is_sched);
+		thd = sl_thd_comp_init(new_comp_cap_info[spdid].defcompinfo, comp_flags);
 		assert(thd);
 		sl_thd_param_set(thd, sched_param_pack(SCHEDP_PRIO, THD_PRIO));
 	}
 
-	if (is_sched) {
+	if (comp_flags == BOOT_COMP_FLAG_SCHED) {
```
gparmer: Don't use flags this way.

`if (comp_flags & BOOT_COMP_FLAG_SCHED)`

src/components/implementation/no_interface/llbooter/boot_deps.h
```diff
@@ -247,29 +251,23 @@ boot_done(void)
...
-	int i;
-	char *prefix = "shmem";
+	int ret;
+	const char prefix[] = "shmem";
 
-	for (i = 0; i < 5; i++) {
-		if (comp_name[i] == '\0' || comp_name[i] != prefix[i])
-			return 0;
-	}
+	ret = strncmp(comp_name, prefix, sizeof(prefix)-1);
```
gparmer: This will cause friction for the convention is that `ret` is what we return.  The `!ret` will cause a double-take.

Likely should be:  `ret = !strncmp(comp_name, prefix, sizeof(prefix)-1);`

src/components/implementation/no_interface/llbooter/boot_deps.h
```diff
@@ -140,10 +145,13 @@ boot_newcomp_definfo_init(spdid_t spdid, compcap_t cc, boot_comp_flag_t comp_fla
 	struct cos_compinfo    *child_ci  = cos_compinfo_get(new_comp_cap_info[spdid].defcompinfo);
 
 	child_ci->comp_cap = cc;
-	child_aep->thd = cos_initthd_alloc(boot_info(), child_ci->comp_cap);
 
-	if (comp_flags == BOOT_COMP_FLAG_SCHED) {
+	if (comp_flags) {
```
gparmer: Now you're removing the scheduler-specific stuff here?  I'm not following.

src/components/implementation/no_interface/llbooter/boot_deps.h
```diff
@@ -216,36 +223,40 @@ boot_newcomp_create(spdid_t spdid, struct cos_compinfo *comp_info, boot_comp_fla
...
-		thd = sl_thd_comp_init(new_comp_cap_info[spdid].defcompinfo, comp_flags);
-		assert(thd);
-		sl_thd_param_set(thd, sched_param_pack(SCHEDP_PRIO, THD_PRIO));
-	}
-
-	if (comp_flags == BOOT_COMP_FLAG_SCHED) {
-		sl_thd_param_set(thd, sched_param_pack(SCHEDP_BUDGET, THD_BUDGET * 1000));
-		sl_thd_param_set(thd, sched_param_pack(SCHEDP_WINDOW, THD_PERIOD * 1000));
-	}
+	if (comp_flags) boot_newschedcomp_cap_init(spdid, ct, pt, cc);
```
gparmer: We want to do this for memory managers as well?  I'd naively expect `if (comp_flags & ...SCHED...)`

src/components/implementation/no_interface/llbooter/boot_deps.h
```diff
@@ -94,73 +117,238 @@ boot_newcomp_sinv_alloc(spdid_t spdid)
...
+		assert(child_aep->tc);
+
+		child_aep->rcv = cos_arcv_alloc(boot_info(), child_aep->thd, child_aep->tc,
+				boot_info()->comp_cap, sched_aep->rcv);
+		assert(child_aep->rcv);
+	} else {
+		child_aep->tc  = sched_aep->tc;
+		child_aep->rcv = sched_aep->rcv;
+	}
+
```
RobertGiff: If we are creating the root scheduler, `comp_flags & BOOT_COMP_FLAG_SCHED`, then we must provide that component with infinite tcap so we don't switch back to the llbooter component.

phanikishoreg:
This PR I understand took a lot of effort to get it to work. And I'm really sorry ask to not merge it.

My major concern is that, I've a different implementation for major changes in this PR:
- detection of schedulers from component name-mangling (also works for hierarchical schedulers, non-schedulers)
- llboot library and API that is used by resmgr and or other components.
However I do not have the 64bit PC compilation changes which I can cherrypick and merge if you allow me to.

The point I'm trying to make is, most of llbooter changes here are required by RESMGR (which is why I didn't want to be blocked) and my implementation of RESMGR already uses a diff implementation of the library, so merging these changes will not help (I'll have to redo my impl to use this or undo this and I'm most likely going to undo this as I've llbooter library that returns multiple return values for SINV, so I don't need to make multiple SINVs to llbooter from RESMGR or other components) except for bug-fixes and 64bit compilation support. 

I understand that this is used in shared-mem-mgr in rumpkernel env, but that also will need to be replaced with resource manager in my PR. 

So my request is to not merge this PR and to allow me to cherry-pick changes /bugfixes (which I'm sure are minimal and won't break my PR and would be easy to merge - I'll do that) that I do not have and merge them to my PR. 
We should talk about this in tomorrow's RK meeting if we have time!!

Others:
Hey, not to be controversial, but I am strongly opposed to @phanikishoreg's proposal here. The cFE is already dependent on this code, and not merging this PR's changes wholesale will break the cFE composite merge. 

I think this PR should be merged, and then Phanis PR modified to work on top of this one. (In a backwards compatible way.)

Yes this is inconvenient for Phani, but throwing away this PR with a ton of quality code and reviews would be a total waste. (Not to mention it completely breaking the cFE work I've done over the past weeks.)

This PR is much more mature than the other, and basically ready to merge. The cFE team urgently needs one of these merged. I acknowledge that this sucks for Phani, but denying this merge will waste everyone else's time. (Especially Robbie, who has put a lot of time into this code.)

phanikishoreg:
@Others I know it was a big ask. I hate to make weeks worth of work go waste. 
But if all of work (in the project that we're all working together for) is going to use RESMGR, I don't know why your merge of this PR will be needed(other than specifics that if I were to merge to my PR) when my PR is merged. 

I take back my request for now to allow all of your work to progress. 
I'll merge once this gets approved and then we'll take it from there in my PR.

phanikishoreg:
Alright, I've robbie's code merged and fixed all the conflicts.

I've also modified the `hypercall.h` (llbooter_inv.h in your PR) to use wrappers and to provide the functionality actually used by resource manager (removed the existing ones.) This is one of the two things that I partially/fully scrapped from this PR. Other thing is sched flags in root scheduler. Everything else is kept. I had to undo a few changes in my PR, like booter returning the capfrontier for resmgr after copying required capabilities using `cos_cap_cpy`. This was one of the things that Gabe & I talked about in our meeting. Debugging was a bit painful but fixed all the bugs after merge! 

I'm going to pull from mainline ppos after this is merged and update my PR. 
I'll not update my PR until this is merged though!

src/components/implementation/no_interface/llbooter/boot_deps.h
```diff
@@ -94,73 +117,247 @@ boot_newcomp_sinv_alloc(spdid_t spdid)
...
+	/* Populate untyped memory */
+	cos_meminfo_alloc(new_comp_cap_info[spdid].compinfo, BOOT_MEM_KM_BASE, CMP_UTYPMEM_SZ);
+
+	cos_cap_cpy_at(new_comp_cap_info[spdid].compinfo, BOOT_CAPTBL_SELF_INITTHD_BASE,
+			boot_info(), child_aep->thd);
+	cos_cap_cpy_at(new_comp_cap_info[spdid].compinfo, BOOT_CAPTBL_SELF_INITTCAP_BASE,
+			boot_info(), child_aep->tc);
+	cos_cap_cpy_at(new_comp_cap_info[spdid].compinfo, BOOT_CAPTBL_SELF_INITRCV_BASE,
+			boot_info(), child_aep->rcv);
+}
```
phanikishoreg: This whole copying thing is more tricky and I've the updates in my PR anyway. 
Just FYI. Like not every component should be given CT/PT/COMP/UNTYPED_PT etc.

src/components/include/cos_defkernel_api.h
```diff
@@ -56,6 +56,10 @@ struct cos_compinfo *cos_compinfo_get(struct cos_defcompinfo *defci);
  * cos_sched_aep_get: returns the sched aep info from the defcompinfo.
  */
 struct cos_aep_info *cos_sched_aep_get(struct cos_defcompinfo *defci);
+/*
+ * boot_info: leverages the above calls to conveniently fetch the current components cos_compinfo struct
+ */
+struct cos_compinfo *boot_info(void);
```
phanikishoreg: Got rid of this too in my merge of your code! Not just from cos_defkernel_api, but really not using this function.

gparmer:
Man, I look away from the comments for a day, and I miss all of the heated, interesting discussion!!!

Phani and I talked about this yesterday, and we still have a few design issues to work out with the resource manager, so I don't want to slow this down.  Some of the changes this would impose on the resource manager are, as I see, required changes anyway (global namespace for memory, passing capability slots to use to populate initial info, etc...).  Very sorry @phanikishoreg that this will cause issues, but on balance, I think this is the right way to go.

In the future, the way to solve this is through collaboration and discussion before we get to this point.

PR #317, opened 2018-02-22
Baremetal memory scan fix

phanikishoreg:

Issue: If a region size is greater than the maximum possible in current composite, it breaks in the sense that it doesn't detect that region as usable.
   => `0xc0100000 + <region_size>` must be less than 1. what is addressable by composite (<1GB currently), 2. should also consider MMIO mappings (for ex. LAPIC is usually mapped at/after 0xFE000000). Fixed that!

VGA printks before paging init is also brought back as I'm not sure what the problem is or I'm not able to replicate it on the HW I've with me.

### Intent for your PR

Choose one (Mandatory):

### Reviewers (Mandatory):
@gparmer @ryuxin 

### Code Quality

### Testing

- microbooter on Qemu, Optiplex 990 (i7 & i5)=4GB DIMM and Inspiron (i5 - newer PC) =8GB DIMM.

### DIFF
--- a/src/kernel/include/shared/cos_config.h
+++ b/src/kernel/include/shared/cos_config.h
@@ -32,6 +32,11 @@
 
 #define COS_MEM_KERN_VA_SZ (1 << 24) /* 16 MB from KERN_START_VA + end of kernel image onward */
 
+#define COS_HW_MMIO_START_VA 0xf0000000 /* Assuming all hardware virtual addresses are beyond this */
+#define COS_PHYMEM_MAX_SZ (COS_HW_MMIO_START_VA-(COS_MEM_KERN_START_VA+COS_MEM_KERN_PA)) /* Maximum addressable physical memory */
+
+#define BOOT_KERN_MEMSCAN_MAX (COS_HW_MMIO_START_VA)
+
 /* To get more memory, we need many PTE caps in the captbl. So give
  * multiple pages to it. 5 is enough for 512 MBs.*/
 #define BOOT_CAPTBL_NPAGES 1
--- a/src/platform/i386/kernel.c
+++ b/src/platform/i386/kernel.c
@@ -140,6 +140,9 @@ kmain(struct multiboot *mboot, u32_t mboot_magic, u32_t esp)
 #ifdef ENABLE_CONSOLE
 	console_init();
 #endif
+#ifdef ENABLE_VGA
+	vga_init();
+#endif
 	max = MAX((unsigned long)mboot->mods_addr,
 	          MAX((unsigned long)mboot->mmap_addr, (unsigned long)(chal_va2pa(&end))));
 	kern_paging_map_init((void *)(max + PGD_SIZE));
@@ -154,7 +157,7 @@ kmain(struct multiboot *mboot, u32_t mboot_magic, u32_t esp)
 	paging_init();
 #ifdef ENABLE_VGA
 	/* uses virtual address for VGA. should be after paging_init() */
-	vga_init();
+	vga_high_init();
 #endif
 	kern_boot_comp();
 	smp_init();
--- a/src/platform/i386/kernel.c
+++ b/src/platform/i386/kernel.c
@@ -91,15 +91,20 @@ kern_memory_setup(struct multiboot *mb, u32_t mboot_magic)
 		struct multiboot_mem_list *mem      = &mems[i];
 		u8_t *                     mod_end  = glb_memlayout.mod_end;
 		u8_t *                     mem_addr = chal_pa2va((paddr_t)mem->addr);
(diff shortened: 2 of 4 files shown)


src/kernel/include/shared/cos_config.h
```diff
@@ -32,6 +32,11 @@
 
 #define COS_MEM_KERN_VA_SZ (1 << 24) /* 16 MB from KERN_START_VA + end of kernel image onward */
 
+#define COS_HW_MMIO_START_VA 0xf0000000 /* Assuming all hardware virtual addresses are beyond this */
```
gparmer: I'll continue looking onwards, but this doesn't make much sense to me initially.  I'm likely missing the utility that I'll understand as I go on.  I'm guessing this is a reserved region to map in dynamically created physical apertures for MMIO?

src/kernel/include/shared/cos_config.h
```diff
@@ -32,6 +32,11 @@
...
+#define COS_PHYMEM_MAX_SZ (COS_HW_MMIO_START_VA-(COS_MEM_KERN_START_VA+COS_MEM_KERN_PA)) /* Maximum addressable physical memory */
+
+#define BOOT_KERN_MEMSCAN_MAX (COS_HW_MMIO_START_VA)
```
gparmer: It isn't clear if this type of stuff to should appear in the memory map or not.

src/kernel/include/shared/cos_config.h
phanikishoreg: I feel like this should really be MAX SZ of HW MMIO/DMA MAPs and not some fixed VA!. I'll change it to that soon and update.

src/kernel/include/shared/cos_config.h
phanikishoreg: This should be 4Gig, so we scan up to that to find a usable region. 
Though, what we usually find usable starts `0x00100000` and definitely ends before LAPIC MMIO `0xfe000000` if not before.

src/platform/i386/kernel.c
```diff
@@ -91,15 +91,20 @@ kern_memory_setup(struct multiboot *mb, u32_t mboot_magic)
 		struct multiboot_mem_list *mem      = &mems[i];
 		u8_t *                     mod_end  = glb_memlayout.mod_end;
 		u8_t *                     mem_addr = chal_pa2va((paddr_t)mem->addr);
+		unsigned long              mem_len  = (mem->len > COS_PHYMEM_MAX_SZ ? COS_PHYMEM_MAX_SZ : mem->len); /* maximum allowed */ 
 
-		printk("\t- %d (%s): [%08llx, %08llx)\n", i, mem->type == 1 ? "Available" : "Reserved ", mem->addr,
-		       mem->addr + mem->len);
+		if (mem->addr > BOOT_KERN_MEMSCAN_MAX || mem->addr + mem_len > BOOT_KERN_MEMSCAN_MAX) {
+			printk("\tScanned maximum allowed regions\n");
+			break;
```
gparmer: I forget:  does multiboot guarantee that the memory maps are ordered from lower to higher addresses?  If not, we don't want to break out, right?  Just `continue`?

Do you know?

src/platform/i386/vga.c
```diff
@@ -161,15 +161,19 @@ cls(void)
...
-	int i = 0;
-
 	video = chal_pa2va(VIDEO);
+}
+
+/* Clear the screen and initialize VIDEO, XPOS and YPOS. */
+void
+vga_init(void)
+{
+	video = (unsigned char *) VIDEO;
```
gparmer: This is great.  We can now also map this into the booter, and it can pass it on to another component.  Would be nice to move all of the print infrastructure out of the kernel.  Thanks for enabling this!

gparmer:
I don't see any terminal issues with this.  @ryuxin if you end up having any comments, @phanikishoreg can go back and fix them in a future PR.  At this point, if it fixes our current hardware boot issues, this is great!!!

Really great job debugging this @phanikishoreg .

src/kernel/include/shared/cos_config.h
phanikishoreg: I commented on this. This is not a great way to fix the current issue. 
I think a better way is to have a limit on HW MMIO maps, so we can correctly set the addressable VA (kernel/user memory) memory.

It should really be `COS_HW_MMIO_MAP_MAX` and I'd set it to 64MB to start with.

src/kernel/include/shared/cos_config.h
phanikishoreg: This is really used to scan up to a max memory address range and not beyond.
I think it should really be up to what's addressable == ULONG_MAX.

src/platform/i386/kernel.c
phanikishoreg: No I don't know the specification for that, will look into it and fix it to use `continue` if there is no specification on the order.

PR #321, opened 2018-03-09
Add multicore support for composite

yzcode:

These changes allow us to use multicore in i386 arch. This PR use universal boot algorithm to boot up APs. The data structure of composite will be setup after AP's booting. Finally, all cores include the BSP and APs will be booted up into the user level(upcall).

The APs can be booted into the user level but we still need some time to modify the kernel's API to support the multicore logic. I add the while(1) before the upcall for APs to avoid the crash. 

This pull request mainly includes:
- AP assembly boot code in loader.S
- lapic code for ipi and universal SMP boot algorithm in lapic.c
- smp_kmain for AP in kernal.c
- all other related changes in the environment setup code

### Intent for your PR

Choose one (Mandatory):

- [X] This PR is for a code-review and is intended to get feedback, but not to be pulled yet.

### Reviewers (Mandatory):
@gparmer @ryuxin @phanikishoreg 

### Code Quality

### Testing

- tested with unit_schedtests (sh qemu-mp.sh unit_schedtests.sh) for 2 cores, 3 cores, and 4 cores. I manually checked all the registers value of all CPUs, and APs are able to execute the calculations and printc function. The problem is that we have some kernel's APIs which do not support the multicore logic, so I am not able to test if the scheduler can run correctly on APs. Anyway, the BSP can always execute the scheduler correctly while APs might crash or block on something.


src/kernel/include/shared/cos_types.h
```diff
@@ -252,10 +252,10 @@ enum
 	BOOT_CAPTBL_COMP0_COMP         = 24,
 	BOOT_CAPTBL_SINV_CAP           = 28,
 	BOOT_CAPTBL_SELF_INITTHD_BASE  = 32,
-	BOOT_CAPTBL_SELF_INITTCAP_BASE = BOOT_CAPTBL_SELF_INITTHD_BASE + NUM_CPU_COS * CAP16B_IDSZ,
-	BOOT_CAPTBL_SELF_INITRCV_BASE  = round_up_to_pow2(BOOT_CAPTBL_SELF_INITTCAP_BASE + NUM_CPU_COS * CAP16B_IDSZ,
+	BOOT_CAPTBL_SELF_INITTCAP_BASE = BOOT_CAPTBL_SELF_INITTHD_BASE + (NUM_CPU_COS + 1) * CAP16B_IDSZ,
```
phanikishoreg: You should use `NUM_CPU` here. Because `NUM_CPU_COS` wastes a capability slot when we work on single core, that is when NUM_CPU is 1.

src/platform/i386/boot_comp.c
```diff
@@ -12,6 +12,9 @@
 
 extern u8_t *boot_comp_pgd;
 
+void *thd_mem[NUM_CPU], *tcap_mem[NUM_CPU];
+struct captbl *ct;
```
phanikishoreg: Perhaps use `glb_boot_ct` as the variable name here? It's confusing with the rest of the functions which use `ct` though from it's local scope.

src/platform/i386/boot_comp.c
```diff
@@ -92,18 +95,19 @@ kern_boot_thd(struct captbl *ct, void *thd_mem, void *tcap_mem)
...
+	assert(cos_info->cpuid == cpu_id);
 	assert(sizeof(struct cos_cpu_local_info) == STK_INFO_SZ);
 	memset(cos_info, 0, sizeof(struct cos_cpu_local_info));
-	cos_info->cpuid          = 0;
+	cos_info->cpuid          = cpu_id;
 	cos_info->invstk_top     = 0;
 	cos_info->overflow_check = 0xDEADBEEF;
-
-	ret = thd_activate(ct, BOOT_CAPTBL_SELF_CT, BOOT_CAPTBL_SELF_INITTHD_BASE, thd_mem, BOOT_CAPTBL_SELF_COMP, 0);
+	ret = thd_activate(ct, BOOT_CAPTBL_SELF_CT, BOOT_CAPTBL_SELF_INITTHD_BASE + cpu_id * captbl_idsize(CAP_THD), thd_mem, BOOT_CAPTBL_SELF_COMP, 0);
```
phanikishoreg: I think we need wrappers to get the capability offset of a core id given the offset of the current core id and the capability type.

phanikishoreg: Because we're going to access cap offsets of a core in many places including user-level components, like `llbooter`, `capmgr`, etc.

src/platform/i386/boot_comp.c
```diff
@@ -134,16 +139,21 @@ kern_boot_thd(struct captbl *ct, void *thd_mem, void *tcap_mem)
...
 {
 	int            ret = 0, nkmemptes;
-	struct captbl *ct;
 	unsigned int   i;
 	u8_t *         boot_comp_captbl;
-	void *         thd_mem, *tcap_mem;
 	pgtbl_t        pgtbl     = (pgtbl_t)chal_va2pa(&boot_comp_pgd), boot_vm_pgd;
 	u32_t          hw_bitmap = 0xFFFFFFFF;
 
+	if (cpu_id > 0 ){
```
phanikishoreg: Must be `if (cpu_id > 0) {`.. Spacing correction only!

src/platform/i386/boot_comp.c
```diff
@@ -161,9 +171,11 @@ kern_boot_comp(void)
 		assert(!ret);
 	}
 
-	thd_mem  = mem_boot_alloc(1);
-	tcap_mem = mem_boot_alloc(1);
-	assert(thd_mem && tcap_mem);
+	for (i = 0; i < NUM_CPU; ++i) {
```
phanikishoreg: nit: I don't remember what CSG says for use of `pre-increment vs post-increment` in `for` loops (though the effect is same here).
In most places we use `post-increment`.

src/platform/i386/boot_comp.c
```diff
@@ -220,9 +232,11 @@ kern_boot_upcall(void)
 	u8_t *entry = mem_bootc_entry();
 	u32_t flags = 0;
 	void *p;
+	if (get_cpuid() >= 0) {
```
phanikishoreg: blank line after declarations and before body.

Can `get_cpuid()` return -ve values? Isn't that an error condition? 
`chal_user_upcall` should not be called with it right? 
Just thinking `assert(get_cpuid() >= 0);` and no `if` condition would probably be the right way to go here.

src/platform/i386/boot_comp.h
```diff
@@ -1,2 +1,2 @@
-void kern_boot_comp(void);
+void kern_boot_comp(const u32_t cpu_id);
```
phanikishoreg: We should probably `typedef cpuid_t unsigned short` and use that everywhere.

src/platform/i386/chal.c
```diff
@@ -89,14 +89,23 @@ chal_init(void)
 	if (d & (1 << 22)) printk("acpi ");
 	if (d & (1 << 9)) printk("apic ");
 	if (c & (1 << 21)) printk("x2apic ");
+	if (c & (1 << 24)) printk("tsc-deadline ");
```
phanikishoreg: nice addition.

src/platform/i386/chal.c
```diff
@@ -89,14 +89,23 @@ chal_init(void)
...
+	a = (a << 16) >> 16;
+	if (a) {
+		printk("\tCPUID base frequency: %d (* 1Mhz)\n", a);
+		printk("\tCPUID max  frequency: %d (* 1Mhz)\n", (b << 16) >> 16);
+	}
+
 	readmsr(MSR_PLATFORM_INFO, &a, &b);
-	printk("\tFrequency: %d (* 100Mz)\n", (a >> 8) & ((1 << 7) - 1));
+	a = (a >> 8) & ((1<<7)-1);
+	if (a) printk("\tMSR Frequency: %d (* 100Mhz)\n", a);
```
phanikishoreg: Nice check for `a`. Perhaps `printk("\nMSR Frequency information not available!\n");` in the else?

src/platform/i386/chal_cpu.h
```diff
@@ -91,10 +92,11 @@ static void
 chal_cpu_init(void)
 {
 	u32_t cr4 = chal_cpu_cr4_get();
+	int cpu_id = get_cpuid();
```
phanikishoreg: Use of a `cpuid_t` (not present yet) would be great imo!

src/platform/i386/gdt.c
```diff
@@ -16,7 +16,8 @@ static u64_t make_gdtr_operand(u16_t limit, void *base);
 void
 chal_tls_update(vaddr_t addr)
 {
-	gdt[SEL_UGSEG / sizeof *gdt] = make_data_desc_at(3, (u32_t)addr);
+	int cpu_id = get_cpuid();
```
phanikishoreg: blank line after declarations.

src/platform/i386/gdt.c
```diff
@@ -26,25 +27,25 @@ chal_tls_update(vaddr_t addr)
...
 
 	/* Initialize GDT. */
-	gdt[SEL_NULL / sizeof *gdt]  = 0;
-	gdt[SEL_KCSEG / sizeof *gdt] = make_code_desc(0);
-	gdt[SEL_KDSEG / sizeof *gdt] = make_data_desc(0);
-	gdt[SEL_UCSEG / sizeof *gdt] = make_code_desc(3);
-	gdt[SEL_UDSEG / sizeof *gdt] = make_data_desc(3);
-	gdt[SEL_TSS / sizeof *gdt]   = make_tss_desc(&tss);
-	gdt[SEL_UGSEG / sizeof *gdt] = make_data_desc(3);
+	gdt[cpu_id][SEL_NULL / sizeof *(gdt[cpu_id])]  = 0;
```
phanikishoreg: Would it be better to have a `volatile u64_t *core_gdt[]` set to `gdt[cpu_id]` and use that? Or a wrapper that does that? We probably don't touch this code once it's working, so may not matter as much.

src/platform/i386/kernel.c
```diff
@@ -15,6 +15,8 @@
 #define CMDLINE_REQ_LEN (ADDR_STR_LEN * 2 + 1)
 
 struct mem_layout glb_memlayout;
+volatile int cores_ready[NUM_CPU];
+volatile int bsp_ready = 0;
```
phanikishoreg: Perhaps `is_bsp_ready`?

src/platform/i386/kernel.c
```diff
@@ -135,9 +137,9 @@ kmain(struct multiboot *mboot, u32_t mboot_magic, u32_t esp)
 #define MAX(X, Y) ((X) > (Y) ? (X) : (Y))
 	unsigned long max;
 
-	tss_init();
-	gdt_init();
-	idt_init();
+	tss_init(0);
```
phanikishoreg: There is a `#define INIT_CORE 0` in cos_config.h, might want to use that.

src/platform/i386/kernel.c
```diff
@@ -164,21 +166,47 @@ kmain(struct multiboot *mboot, u32_t mboot_magic, u32_t esp)
...
+
 	/* should not get here... */
 	khalt();
 }
 
+void
+smp_kmain(void)
+{
+	volatile int cpu_id = get_cpuid();
+	struct cos_cpu_local_info *cos_info = cos_cpu_local_info();
```
phanikishoreg: nit: blank line after declarations.

src/platform/i386/kernel.c
```diff
@@ -164,21 +166,47 @@ kmain(struct multiboot *mboot, u32_t mboot_magic, u32_t esp)
...
+	idt_init(cpu_id);
+
+	paging_init();
+	kern_boot_comp(cpu_id);
+	lapic_init();
+	lapic_timer_init();
+
+	printk("New CPU %d Booted\n", cpu_id);
+	cores_ready[cpu_id] = 1;
+	// waiting for all cored booted
```
phanikishoreg: comment style for single line as per CSG: `/* comment */`.

src/platform/i386/kernel.c
```diff
@@ -164,21 +166,47 @@ kmain(struct multiboot *mboot, u32_t mboot_magic, u32_t esp)
...
+	while(bsp_ready == 0);
+	while(1);
```
phanikishoreg: Perhaps use a comment here like /* FIXME: reason for this here */

src/platform/i386/lapic.c
```diff
@@ -95,7 +116,7 @@ u32_t               lapic_timer_calib_init   = 0;
 static void
 lapic_write_reg(u32_t off, u32_t val)
 {
-	*(u32_t *)(lapic + off) = val;
+	*(u32_t volatile *)(lapic + off) = val;
```
phanikishoreg: again a nitpick: I don't remember if CSG says anything about the placement of `volatile` keyword. But I think in most places we use like `volatile u32_t *`. 

Thanks for fixing this! I don't know how it was working fine for timer configurations even without. 
Perhaps this will fix a few issues where the timer interrupts were not triggered in `Qemu`. Hope it does!

src/platform/i386/lapic.c
```diff
@@ -220,9 +241,35 @@ lapic_find_localaddr(void *l)
...
 	return addr;
 }
 
+void
+lapic_init(void)
+{
+	u32_t version;
+
+	assert(lapic);
+	lapic_write_reg(LAPIC_SIV_REG, LAPIC_SIV_ENABLE);
```
phanikishoreg: If I remember correctly, we need to set the vector number for SIV here. 
The vector number for SIV should be of `0xpF` where the lower order 4 bits are all 1 and 0<=p<=F(this is obvious I know) and that the SIV handler should not `ack` the interrupts.

src/platform/i386/lapic.c
```diff
@@ -220,9 +241,35 @@ lapic_find_localaddr(void *l)
...
+
+	version = lapic_read_reg(LAPIC_VERSION_REG);
+	/* don't want to deal with booting up using CMOS */
```
phanikishoreg: Is this a FIXME? or a TODO? Would be good to note that here.

src/platform/i386/lapic.c
```diff
@@ -357,10 +404,99 @@ lapic_timer_init(void)
...
+}
+
+static int
+lapic_ipi_send(u32_t dest, u32_t vect_flags)
+{
+	lapic_write_reg(LAPIC_ICR + 0x10, dest << 24);
+	lapic_read_reg(LAPIC_ICR + 0x10);
+
+	lapic_write_reg(LAPIC_ICR, vect_flags);
+	lapic_read_reg(LAPIC_ICR);
```
phanikishoreg: awesome, now we got IPIs!!!

src/platform/i386/lapic.c
```diff
@@ -357,10 +404,99 @@ lapic_timer_init(void)
...
+
+	/*
+	 * Set up the processor boot-up code .  Use SMC to create the
+	 * smp loader code with the stack address inlined into it at
+	 * an address that real-mode 16-bit code can execute
+	 */
+	memcpy((char *)chal_pa2va(SMP_BOOT_PATCH_ADDR), &smppatchstart, &smppatchend - &smppatchstart);
+	stackpatch  = (char **)chal_pa2va(SMP_BOOT_PATCH_ADDR + (&smpstack - &smppatchstart));
+
+	for (i = 1 ; i < ncpus ; i++) {
```
phanikishoreg: Should be `for (i = 1; i < ncpus; i++)`. No space before `;`.

src/platform/i386/lapic.c
```diff
@@ -357,10 +404,99 @@ lapic_timer_init(void)
...
+		cli->cpuid  = i; /* the rest is initialized during the bootup process */
+
+		/* Now the IPI coordination process to boot the AP: first send init ipi... */
+		lapic_ipi_send(apicids[i], LAPIC_ICR_LEVEL | LAPIC_ICR_ASSERT | LAPIC_ICR_INIT);
+		delay_us(200);
+		/* ...deassert it... */
+		lapic_ipi_send(apicids[i], LAPIC_ICR_LEVEL | LAPIC_ICR_INIT);
+		/* ...wait for 10 ms... */
+		delay_us(10000);
+		for (j = 0 ; j < 2 ; j++) {
```
phanikishoreg: no space before `;`.

src/platform/i386/lapic.c
```diff
@@ -357,10 +404,99 @@ lapic_timer_init(void)
...
+		if (tsc >= end) return;
+		asm("pause");
+	}
+}
+
+/* The SMP boot patchcode from loader.S */
+extern char smppatchstart, smppatchend, smpstack, stack;
+
+void
+smp_bootall(volatile int *cores_ready)
```
phanikishoreg: perhaps `smp_boot_all_ap`?

src/platform/i386/loader.S
```diff
@@ -109,7 +111,7 @@ smploader:
 
 	movw    %cs, %ax
 	movw    %ax, %ds
-	lgdtl   RELOCATE_ADDR_CS(smpgdtdesc) /* using cs as ds: idea taken from Quest */
+	lgdt    RELOCATE_ADDR_CS(smpgdtdesc) /* using cs as ds: idea taken from Quest */
```
phanikishoreg: Just out of curiosity, any reason for this change?

src/platform/i386/vm.c
```diff
@@ -17,6 +17,9 @@ struct liveness_entry __liveness_tbl[LTBL_ENTS] CACHE_ALIGNED LARGE_BSS;
 u32_t boot_comp_pgd[PAGE_SIZE / sizeof(u32_t)] PAGE_ALIGNED = {[0] = 0 | PGTBL_PRESENT | PGTBL_WRITABLE | PGTBL_SUPER,
                                                                [KERN_INIT_PGD_IDX] = 0 | PGTBL_PRESENT | PGTBL_WRITABLE
                                                                                      | PGTBL_SUPER};
+u32_t boot_ap_pgd[PAGE_SIZE / sizeof(u32_t)] PAGE_ALIGNED = {[0] = 0 | PGTBL_PRESENT | PGTBL_WRITABLE | PGTBL_SUPER,
```
phanikishoreg: Is there a reason to not make `boot_comp_pgd` a 2D array with NUM_CPU? I feel like that would be better.

phanikishoreg: Wait.. I thought we share the pgtbls across cores. 
so we don't need a 2D array or a `boot_ap_pgd`, right?

src/platform/i386/vm.c
```diff
@@ -57,7 +60,12 @@ kern_setup_image(void)
 {
 	unsigned long i, j;
 	paddr_t       kern_pa_start, kern_pa_end;
+	int cpu_id = get_cpuid();
 
+	if (cpu_id > 0) {
```
phanikishoreg: Perhaps not call this in for APs, instead directly call `chal_cpu_init()`?

src/platform/i386/kernel.c
gparmer: `while(1) ;` (space before `;`)

For the current code, I'd rather see that the `while (1) ;` be in user-space *after* the `kern_boot_upcall`.  That would be an easier space for @phanikishoreg to jump in.

src/platform/i386/kernel.h
```diff
@@ -54,12 +55,13 @@ void *       acpi_find_apic(void);
 u32_t        lapic_find_localaddr(void *l);
 void         lapic_set_page(u32_t page);
 void         lapic_timer_init(void);
+void 				 lapic_init(void);
```
gparmer: spacing here seems odd

src/platform/i386/lapic.c
gparmer: indeed, storage modifiers before data types.

src/platform/i386/lapic.c
```diff
@@ -107,7 +128,7 @@ lapic_ack(void)
 static u32_t
 lapic_read_reg(u32_t off)
 {
-	return *(u32_t *)(lapic + off);
+	return *(u32_t volatile *)(lapic + off);
```
gparmer: ditto

src/platform/i386/lapic.c
gparmer: :+1:

src/platform/i386/lapic.c
```diff
@@ -357,10 +404,99 @@ lapic_timer_init(void)
...
+	lapic_write_reg(LAPIC_ICR + 0x10, dest << 24);
+	lapic_read_reg(LAPIC_ICR + 0x10);
+
+	lapic_write_reg(LAPIC_ICR, vect_flags);
+	lapic_read_reg(LAPIC_ICR);
+
+	return 0;
+}
+
+/* HACK: assume that the HZ of the processor is equivalent to that on the computer used for compilation. */
```
gparmer: Can we use the clock speed returned by the msr?

I don't have a good solution to this.  @phanikishoreg ?

src/platform/i386/lapic.c
```diff
@@ -357,10 +404,99 @@ lapic_timer_init(void)
...
+static void
+delay_us(u32_t us)
+{
+	unsigned long long hz = CPU_GHZ, hz_per_us = hz * 1000;
+	unsigned long long end;
+	volatile unsigned long long tsc;
```
gparmer: empty line after declarations...

src/platform/i386/lapic.c
```diff
@@ -357,10 +404,99 @@ lapic_timer_init(void)
...
+
+void
+smp_bootall(volatile int *cores_ready)
+{
+	int i;
+	u32_t ret;
+	char **stackpatch;
+
+	/*
+	 * Set up the processor boot-up code .  Use SMC to create the
```
gparmer: remove space before `.`.

src/platform/i386/lapic.c
gparmer: my fault ;-)

src/platform/i386/lapic.c
```diff
@@ -357,10 +404,99 @@ lapic_timer_init(void)
...
+		delay_us(200);
+		/* ...deassert it... */
+		lapic_ipi_send(apicids[i], LAPIC_ICR_LEVEL | LAPIC_ICR_INIT);
+		/* ...wait for 10 ms... */
+		delay_us(10000);
+		for (j = 0 ; j < 2 ; j++) {
+			/* ...send startup IPIs... */
+			assert(!(SMP_BOOT_PATCH_ADDR >> 12 & ~0xFF)); /* some address validation */
+			lapic_ipi_send(apicids[i], LAPIC_ICR_SIPI | (SMP_BOOT_PATCH_ADDR >> 12));
+			/* ...wait for 20 us... */
```
gparmer: 200 us?

src/platform/i386/lapic.c
```diff
@@ -357,10 +404,99 @@ lapic_timer_init(void)
...
+			delay_us(200);
+		}
+		// waiting for booting
+		while(*(int volatile *)(cores_ready + i) == 0);
```
gparmer: volatile int

src/platform/i386/tss.c
```diff
@@ -11,11 +11,11 @@ struct tss tss;
 #define PGMASK BITMASK(PGSHIFT, PGBITS) /* Page offset bits (0:12). */
 
 void
-tss_init(void)
+tss_init(const u32_t cpu_inx)
```
gparmer: consistent:  `cpu_id`.  No reason to create a new var name.

src/platform/i386/vm.c
gparmer: I forget the core of this issue.  @yzcode ?

src/kernel/include/shared/cos_types.h
yzcode: I will use NUM_CPU instead.

src/platform/i386/boot_comp.c
yzcode: good point, I will check this out.

src/platform/i386/boot_comp.c
yzcode: indeed.

src/platform/i386/boot_comp.c
yzcode: Sorry for this. I will double check with the CSG.

yzcode: dito

src/platform/i386/boot_comp.c
yzcode: dito

src/platform/i386/boot_comp.c
yzcode: I will address this

src/platform/i386/boot_comp.h
yzcode: I will use cpuid_t instead.

src/platform/i386/chal_cpu.h
yzcode: indeed

src/platform/i386/gdt.c
yzcode: dito

src/platform/i386/gdt.c
yzcode: I will keep this based on your points.

src/platform/i386/kernel.c
yzcode: Yeah, I will remove bsp_ready, and use cires_ready[0] instead.

src/platform/i386/kernel.c
yzcode: good point

src/platform/i386/kernel.c
yzcode: dito

src/platform/i386/kernel.c
yzcode: dito

src/platform/i386/kernel.c
yzcode: I am not sure where is the start of the code after the `kern_boot_upcall`, so I put the `while (1) ;` here. Could you guys give me the link to the code after the `kern_boot_upcall`.

src/platform/i386/lapic.c
yzcode: dito.

src/platform/i386/lapic.c
yzcode: I will have a look at this later.

src/platform/i386/lapic.c
yzcode: I will comment more details here.

src/platform/i386/lapic.c
yzcode: dito.

src/platform/i386/lapic.c
yzcode: ditto

src/platform/i386/lapic.c
```diff
@@ -357,10 +404,99 @@ lapic_timer_init(void)
...
+		/* ...wait for 10 ms... */
+		delay_us(10000);
+		for (j = 0 ; j < 2 ; j++) {
+			/* ...send startup IPIs... */
+			assert(!(SMP_BOOT_PATCH_ADDR >> 12 & ~0xFF)); /* some address validation */
+			lapic_ipi_send(apicids[i], LAPIC_ICR_SIPI | (SMP_BOOT_PATCH_ADDR >> 12));
+			/* ...wait for 20 us... */
+			delay_us(200);
+		}
+		// waiting for booting
```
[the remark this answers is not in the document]
yzcode: ditto

src/platform/i386/lapic.c
yzcode: Nice name

src/platform/i386/loader.S
yzcode: I cannot remember the detail about this. Sorry for this, I might test the difference later.

src/platform/i386/vm.c
yzcode: We cannot reuse the `boot_comp_pgd` for APs' booting because we will unmap the lower address of `boot_comp_pgd` in `kern_setup_image` later. But we will switch the AP's pgd back to `boot_comp_pgd` after the AP is booted into C-level to keep the simplicity.

src/platform/i386/vm.c
yzcode: Yeah, good point.

src/platform/i386/boot_comp.c
```diff
@@ -82,7 +85,7 @@ boot_pgtbl_mappings_add(struct captbl *ct, capid_t pgdcap, capid_t ptecap, const
 
 /* FIXME:  loops to create threads/tcaps/rcv caps per core. */
 static void
-kern_boot_thd(struct captbl *ct, void *thd_mem, void *tcap_mem)
+kern_boot_thd(struct captbl *ct, void *thd_mem, void *tcap_mem, const u32_t cpu_id)
```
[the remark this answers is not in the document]
yzcode: Indeed.

src/platform/i386/gdt.c
```diff
@@ -4,7 +4,7 @@
 #include "tss.h"
 #include "chal_asm_inc.h"
 
-static volatile u64_t gdt[SEL_CNT] __attribute__((aligned(sizeof(u64_t))));
+static volatile u64_t gdt[NUM_CPU][SEL_CNT] __attribute__((aligned(sizeof(u64_t))));
```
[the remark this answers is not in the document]
yzcode: I will check this out.

src/platform/i386/lapic.c
yzcode: yeah, 200 us.

src/kernel/include/shared/cos_types.h
phanikishoreg: I don't know why we have `NUM_CPU_COS`, it's from old code/kernel I think.

src/platform/i386/chal.c
phanikishoreg: Yes, I think we can get rid of that calibration if we have MSR frequency available and the CPU supports TSC-DEADLINE. 
For Oneshot timers, we'll still need a calibration logic if not using HPET as we're currently.
Also, we need to have `ENABLE_HPET` in the code for HPET driver and calibration if we are using HPET.

src/platform/i386/kernel.c
phanikishoreg: @yzcode It is here in assembly: https://github.com/gwsystems/composite/blob/ppos/src/components/lib/cos_asm_upcall.S#L46

Here in C: https://github.com/gwsystems/composite/blob/ppos/src/components/lib/cos_component.c#L141

src/platform/i386/vm.c
phanikishoreg: What is the usual way of multi-core booting? Does the BSP continue in user-land while the other processors are still booting? 

If that is not the case, we probably can delay the call to `kern_setup_image` (and perhaps subsequent ones) in boot processor for until all other APs get to that point? 

What you've now maybe OK but we need to we very sure we don't have a mapping boot_comp_pgd that's used by other APs or we don't have a mapping in AP that's not in boot_comp_pgd.

src/platform/i386/lapic.c
phanikishoreg: That seems like a great idea to me. We should use the clock speed from the MSR if it's available and fall back to a HACKish solution if it's not.

gparmer:
Please ping us when you've updated all of this to pull.

gparmer:
Does this work now with booting into user-level?

yzcode:
@gparmer Yes, it does. I am trying to fix the micro_booter test to see if it works fine. It looks like @phanikishoreg has fixed the problems in some APIs such as cos_sched_aep_get(), and etc.. in
https://github.com/phanikishoreg/composite/commit/0fdb01346b05f35ddd9c6cc661ffc51276a2643b

src/platform/i386/vm.c
yzcode: @phanikishoreg in my understanding, the boot_ap_pgd will only be used in the assembly booting code, all cores will share the boot_comp_pgd after that.

src/platform/i386/lapic.c
yzcode: This comment was made by @gparmer, so I am not sure about the meaning of this comment here.
Maybe Gabe could explain more about this?

src/platform/i386/lapic.c
yzcode: This comment was made by @gparmer, so I am not sure about the meaning of this comment here.
Maybe Gabe could explain more about this?

src/platform/i386/lapic.c
```diff
@@ -320,6 +320,13 @@ lapic_set_page(u32_t page)
 	printk("\tSet LAPIC @ %p\n", lapic);
 }
 
+int
+lapic_spurious_handler(struct pt_regs *regs)
+{
+	printk("\n\n spurious irq happened on cpu%d @ %x\n", get_cpuid(), regs->ip);
```
phanikishoreg: We don't need a printk here.. 

Adding a reference to where I found that spurious vector handlers should not send EOI (last line in the below pasted information) since @yzcode asked in our last meeting:

This is in the Intel® 64 and IA-32 Architectures Software Developer’s Manual => Volume 3 (3A, 3B, 3C & 3D): System Programming Guide =>
(https://software.intel.com/sites/default/files/managed/39/c5/325462-sdm-vol-1-2abcd-3abcd.pdf)
```
10.9 SPURIOUS INTERRUPT
A special situation may occur when a processor raises its task priority to be greater than or equal to the level of the interrupt for which the processor INTR signal is currently being asserted. 
If at the time the INTA cycle is issued, the interrupt that was to be dispensed has become masked (programmed by software), the local APIC will deliver a spurious-interrupt vector. 
Dispensing the spurious-interrupt vector does not affect the ISR, so the handler for this vector should return without an EOI.
```

I've also previously mentioned that the vector number should end with 0xF for Spurious interrupt, but I'm not sure where I read that. It doesn't seem to be in the manual I referenced. Not sure!

yzcode: Thank you for your clarification, I am using 0x3F (63) for the vector number of the Spurious Interrupt now.

yzcode:
Now, we can run micro_booter on 8 cores at the same time.

yzcode:
@gparmer @phanikishoreg @ryuxin I think we need another round of review for this pr.

src/kernel/include/shared/cos_types.h
```diff
@@ -252,16 +256,42 @@ enum
...
-	BOOT_CAPTBL_SELF_INITHW_BASE   = round_up_to_pow2(BOOT_CAPTBL_SELF_INITRCV_BASE + NUM_CPU_COS * CAP64B_IDSZ,
+	BOOT_CAPTBL_SELF_INITHW_BASE   = round_up_to_pow2(BOOT_CAPTBL_SELF_INITRCV_BASE + NUM_CPU * CAP64B_IDSZ,
                                                         CAPMAX_ENTRY_SZ),
 	BOOT_CAPTBL_LAST_CAP           = BOOT_CAPTBL_SELF_INITHW_BASE + CAP32B_IDSZ,
 	/* round up to next entry */
 	BOOT_CAPTBL_FREE = round_up_to_pow2(BOOT_CAPTBL_LAST_CAP, CAPMAX_ENTRY_SZ)
 };
 
+/* To get more memory, we need many PTE caps in the captbl. So give
+ * multiple pages to it. 5 is enough for 512 MBs.*/
```
phanikishoreg: I think the comments are not covering the point for this macro and the math. 
The MATH is not very intuitive, so perhaps an elaborate comment here would be very useful in my opinion. 

Also comment style is not following CSG!

src/components/implementation/tests/micro_booter/micro_booter.c
```diff
@@ -15,26 +15,45 @@ term_fn(void *d)
...
-	cos_meminfo_init(&booter_info.mi, BOOT_MEM_KM_BASE, COS_MEM_KERN_PA_SZ, BOOT_CAPTBL_SELF_UNTYPED_PT);
-	cos_compinfo_init(&booter_info, BOOT_CAPTBL_SELF_PT, BOOT_CAPTBL_SELF_CT, BOOT_CAPTBL_SELF_COMP,
-	                  (vaddr_t)cos_get_heap_ptr(), BOOT_CAPTBL_FREE, &booter_info);
+	int cycs, i;
+	static int first_init = 1, init_done = 0;
+	static int cpu_counter = 0;
 
-	termthd = cos_thd_alloc(&booter_info, booter_info.comp_cap, term_fn, NULL);
-	assert(termthd);
+	cos_faa(&cpu_counter, 1);
```
phanikishoreg: I talked to @gparmer about this. We want to keep the kernel and user logic simple around the NUMBER of cores booted up. In order to do that, we just make sure we have as many cores (LAPICs) as we configure in `NUM_CPU`. I've modified that in my smp branch if you want to merge that.

src/platform/i386/exception.c
```diff
@@ -9,7 +9,8 @@
 void
 print_regs_state(struct pt_regs *regs)
 {
-	printk("\n\nGeneral registers-> EAX: %x, EBX: %x, ECX: %x, EDX: %x\n", regs->ax, regs->bx, regs->cx, regs->dx);
+	printk("\n\nCPU%d registers:\n", get_cpuid());
```
phanikishoreg: This is not very useful! I think we need `get_cpuid()` in every printk below. 
You may have noticed that the interleavings among cores just makes it impossible to make any sense of the output.. To be able to understand anything that's going on (mainly if fewer cores fault), then we need to know which of the below statements (registers) correspond to which core..

phanikishoreg: This is really based on the experience in debugging some of the issues. I totally understand why you've this here. (Qemu just cannot interleave like the real hw..)..

src/platform/i386/lapic.c
phanikishoreg: Now we have a global variable to get the MSR frequency.. So you should use that if it's not zero!

phanikishoreg:
I guess this is in a great state now! Just a few nitpicks from me! 

Once this is merged, I've my SMP branch up-to-date with this PR and have changes for PER-CORE scheduling across llbooter -> capmgr -> sched -> unit test components.. Most importantly fixes in cos_kernel_api library for multi-core.. I'll be able to immediately make a follow up PR. That should enable everyone to start with many working parts for multi-core. Every component that is required for either immediate projects is working for multi-core.. 

Great job @yzcode !!! Thanks for pushing this work!! 

IPIs and IOAPIC (debugging and API for routing..) on multi-core in the kernel (that @yzcode will be looking into) and cross-core scheduler communication should just make core infrastructure complete for the immediate papers (I'll be)..!!

src/kernel/include/shared/cos_types.h
```diff
@@ -252,16 +256,42 @@ enum
...
+#define BOOT_CAPTBL_NPAGES ((BOOT_CAPTBL_FREE + CAPTBL_EXPAND_SZ + CAPTBL_EXPAND_SZ * 2 - 1) / (CAPTBL_EXPAND_SZ * 2))
+
+#define BOOT_CAPTBL_SELF_INITTCAP_CPU_BASE (captbl_tcap_offset(cos_cpuid()))
+#define BOOT_CAPTBL_SELF_INITTHD_CPU_BASE (captbl_thd_offset(cos_cpuid()))
+#define BOOT_CAPTBL_SELF_INITRCV_CPU_BASE (captbl_arcv_offset(cos_cpuid()))
+
+static inline unsigned long
```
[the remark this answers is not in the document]
phanikishoreg: I think it doesn't matter for the micro booter test you've been focusing on. 

But I've modified the BOOT_CAPTBL_SELF_INITTHD and INITTCAP offsets to not share a cache-line between cores. Just FYI.

I'll merge that in my PR so don't worry about it unless you need it for this PR in anyway!

src/platform/i386/boot_comp.c
```diff
@@ -12,6 +12,9 @@
 
 extern u8_t *boot_comp_pgd;
 
+void *thd_mem[NUM_CPU], *tcap_mem[NUM_CPU];
```
gparmer: Looks to me like we don't need these anymore.  They can just be local in `kern_boot_comp`.  Can you add an issue to fix this?

src/platform/i386/boot_comp.c
```diff
@@ -134,42 +139,50 @@ kern_boot_thd(struct captbl *ct, void *thd_mem, void *tcap_mem)
...
 	int            ret = 0, nkmemptes;
-	struct captbl *ct;
 	unsigned int   i;
 	u8_t *         boot_comp_captbl;
-	void *         thd_mem, *tcap_mem;
 	pgtbl_t        pgtbl     = (pgtbl_t)chal_va2pa(&boot_comp_pgd), boot_vm_pgd;
 	u32_t          hw_bitmap = 0xFFFFFFFF;
 
+	assert(cpu_id >= 0);
+	if (NUM_CPU > 1 && cpu_id > 0) {
```
gparmer: is the first check here redundant with the second?

src/platform/i386/chal.c
```diff
@@ -89,14 +90,26 @@ chal_init(void)
...
 	if (d & (1 << 9)) printk("apic ");
 	if (c & (1 << 21)) printk("x2apic ");
+	if (c & (1 << 24)) printk("tsc-deadline ");
 	chal_cpuid(0x80000001, &a, &b, &c, &d);
 	if (d & (1 << 27)) printk("rdtscp ");
 	chal_cpuid(0x80000007, &a, &b, &c, &d);
 	if (d & (1 << 8)) printk("invariant_tsc ");
 	printk("]\n");
 
+	chal_cpuid(0x16, &a, &b, &c, &d);
```
gparmer: Does this work on processors without the frequency support?  Do we not need a `if (cpuid_hz_support_check) {...}`?  If this works, on all processors, great!

src/platform/i386/loader.S
gparmer: `l` is just the size modifier.  Likely defaults to `l` on 32 bit anyway?

gparmer:
@yzcode Great job with all of this!

PR #332, opened 2018-03-18
sl_thd WOKEN race

maloneya:

**Add description here.**
Intended to solve a race that was occurring that would cause threads to deadlock in the case where a thread is woken before it actually calls block.

though this solves my issue i may have missed some necessary changes elsewhere in parts of the scheduler my code isn't using. 

updated state sl_thd state machine:
    - runnable -> blocked: when the thread blocks on an operation
    - runnable -> woke: when a thread is runnable and is `sl_wakeup`ed by another thread
    - woke -> runnable: when a thread blocks `sl_block`s when it has been previously `sl_wakeup`ed

### Intent for your PR

Choose one (Mandatory):

- [x ] This PR is mature, and ready to be integrated into the repo.

### Reviewers (Mandatory):

### Code Quality

- [ x] Comments adhere to the Style Guide (SG)
- [x ] Naming adhere's to the SG
- [x ] All other aspects of the SG are adhered to, or exceptions are justified in this pull request
- [] I have run the auto formatter on my code before submitting this PR (see doc/auto_formatter.md for instructions)

- [ x] I've made an attempt to remove all redundant code
- [x ] I've considered ways in which my changes might impact existing code, and cleaned it up
- [x ] I've formatted the code in an effort to make it easier to read (proper error handling, function use, etc...)
- [ x] I've commented appropriately where code is tricky
- [ x] I agree that there is no "throw-away" code, and that code in this PR is of high quality

### Testing

voter/demo_app.rs
-

### DIFF
--- a/src/components/include/sl.h
+++ b/src/components/include/sl.h
@@ -114,6 +114,13 @@ sl_cs_owner(void)
 	return sl__globals()->lock.u.s.owner == sl_thd_thdcap(sl_thd_curr());
 }
 
+static inline int
+sl_is_runnable(struct sl_thd *t)
+{
+	/* WOKE threads are still runnable */
+	return (t->state == SL_THD_RUNNABLE || t->state == SL_THD_WOKEN);
+}
+
 /* ...not part of the public API */
 /*
  * @csi: current critical section value
@@ -462,8 +469,7 @@ sl_cs_exit_schedule_nospin_arg(struct sl_thd *to)
 		}
 	}
 
-	//WOKE is subset of RUNNABLE so should be safe to dispatch to
-	assert(t->state == SL_THD_RUNNABLE || t->state == SL_THD_WOKEN);
+	assert(sl_is_runnable(t));
 	sl_cs_exit();
 
 	ret = sl_thd_activate(t, tok);
--- a/src/components/include/sl.h
+++ b/src/components/include/sl.h
@@ -371,7 +371,7 @@ sl_thd_activate(struct sl_thd *t, sched_tok_t tok)
 		return cos_switch(sl_thd_thdcap(t), sl_thd_tcap(t), t->prio,
 				  g->timeout_next, g->sched_rcv, tok);
 	} else {
-		return cos_defswitch(sl_thd_thdcap(t), t->prio, t == g->sched_thd ? 
+		return cos_defswitch(sl_thd_thdcap(t), t->prio, t == g->sched_thd ?
 				     TCAP_TIME_NIL : g->timeout_next, tok);
 	}
 }
@@ -447,7 +447,7 @@ sl_cs_exit_schedule_nospin_arg(struct sl_thd *to)
 
 		if (t->last_replenish == 0 || t->last_replenish + t->period <= now) {
 			tcap_res_t currbudget = 0;
-			cycles_t replenish    = now - ((now - t->last_replenish) % t->period);  
+			cycles_t replenish    = now - ((now - t->last_replenish) % t->period);
 
 			ret = 0;
 			currbudget = (tcap_res_t)cos_introspect(ci, sl_thd_tcap(t), TCAP_GET_BUDGET);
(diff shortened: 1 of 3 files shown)


src/components/include/sl.h
```diff
@@ -462,15 +462,16 @@ sl_cs_exit_schedule_nospin_arg(struct sl_thd *to)
 		}
 	}
 
-	assert(t->state == SL_THD_RUNNABLE);
+	//WOKE is subset of RUNNABLE so should be safe to dispatch to
```
gparmer: `/* */` is the prescribed comment style for single-line comments.

gparmer: I don't think that describing the WOKE state as a *subset* is correct.  Comment *could* read:

`/* WOKE threads are still runnable */`

Likely the extensible way to handle this is to have a function `sl_is_runnable(t)` that can encode which states include threads that can be dispatched to.

src/components/lib/sl/sl.c
```diff
@@ -157,17 +157,24 @@ sl_thd_block_no_cs(struct sl_thd *t, sl_thd_state_t block_type, cycles_t timeout
 void
 sl_thd_block(thdid_t tid)
 {
+
```
gparmer: Random empty line?

src/components/lib/sl/sl.c
```diff
@@ -157,17 +157,24 @@ sl_thd_block_no_cs(struct sl_thd *t, sl_thd_state_t block_type, cycles_t timeout
...
 	struct sl_thd *t;
 
 	/* TODO: dependencies not yet supported */
 	assert(!tid);
 
 	sl_cs_enter();
 	t = sl_thd_curr();
-	if (sl_thd_block_no_cs(t, SL_THD_BLOCKED, 0)) {
+
```
gparmer: Not really a new "paragraph", so no blank line?

src/components/lib/sl/sl.c
```diff
@@ -157,17 +157,24 @@ sl_thd_block_no_cs(struct sl_thd *t, sl_thd_state_t block_type, cycles_t timeout
...
+	if (unlikely(t->state == SL_THD_WOKEN)) {
+		t->state = SL_THD_RUNNABLE;
+	}
+
```
gparmer: absolutely no paragraph between different conditions.  No empty line.

maloneya:
Style changes made. Went ahead ad added sl_is_runnable(t) to sl.h

gparmer:
Sorry, but this isn't mergeable as is.  Do a `pull` as there are conflicts with the current tree in `src/components/lib/sl/sl.c`.

phanikishoreg:
Sorry, this is not a complete fix for the reasons I've mentioned on slack: 

*******************
For AEP threads, we get multiple consecutive `blocked/unblocked` events from kernel and are very common (not edge cases). 

Ex: Lets say that a thread is blocked on `cos_rcv`, and an interrupt woke it up and it runs to completion (because it is higher prio than the currently executing thread). When the thread calls `cos_rcv` again, it activates its scheduler, which sees that the event is a `blocked` event (because that thread is blocked on cos_rcv at this point). At this point, scheduler sees that it's a duplicate blocked event (as in the thread is already blocked). If this repeats, that is multiple blocked events. 

Another case is multiple wake up events.. Ex: A thread is woken up by an interrupt, but hasn't finished executing (was preempted by another higher prio thread or because the thread ran out of budget), when the scheduler activates, it sees that this was a wakeup event. Imagine a case where this repeats. There will be multiple wakeup events. I'm trying to think if there were any additional edge cases, I'll post them if I remember of any. 

@maloneya’s PR  brings back the old code for WOKEN state (without special handling for scheduling events from kernel, seem to work for his setup and I assume because there are no kernel/scheduler events). But that wouldn't work well in a complex system. We need special handling for scheduler events from kernel, and have WOKEN state for user-level races as you suggested before.
*******************

I've made changes to allow WOKEN state for user-level races and have separate block/wakeup API for redundant scheduling events. 
I will be able to submit a PR after @maloneya helps me test and confirm that the deadlock doesn't occur. I've tested the sl library/sched-component unit-tests we have in the mainline and it seems to work. (We need a unit test framework for lock API.)

If it's okay, please close this PR because merging this will not work well for many use-cases we have.

gparmer:
OK, @phanikishoreg work with @maloneya to put together a PR that does fix the requisite issues.  Regardless, we need locks working, so if this is going to take a while, let me know.

PR #333, opened 2018-03-20
cfe2cos, the mega-merge

Others:
The cFE is a complex piece of software, consequently our support implementation is quite complex. I will attempt to summarize what is going on with each change.

#### cFE_booter Support Code 
- Summary: To get from `cos_init` to the actual cFE is a bit of code, most of which is inspired by the posix implementation. We need to set up all of our data structures, start our scheduler, and make sure the filesystem is in good order. It's spread across of few files -- but again, not too complicated. We also boost the stack size so that the cFE doesn't blow the small stack. You can trace configuration logic from `cFE_entrypoint.c` The only other interesting stuff is in `cFE_util.c`.
- Component files: `cFE_booter/cFE_entrypoint.c` `cFE_booter/cFE_util.c` `cFE_booter/cFE_util.h` parts of `cFE_booter/osapi.c` and `cFE_booter/ostask.c` `include/shared/consts.h`
- Known Issues: The first code we wrote. It's a bit overly complex and stupid. It's written in a very formal style, which we quickly diverged from.

#### General OSAL Method Implementations
- Summary: The most straightforward of the straightforward. Some of these are a bit hacky, but the code quality is fine. Many of these methods are irrelevant and unimplemented.
- Component files: `cFE_booter/osapi.c` `cFE_booter/psp.c` `cFE_booter/ostimer.c` `cFE_booter/osnetwork.c`
- Known Issues:

#### Copied Files
- Summary: Shared OSAL code we just copied out of the cFE. Unmodified, doesn't need review.
- Component files: `cFE_booter/cfe_psp_*`
- Known Issues:

#### App Loading
- Summary: Code to work as glue between how the cFE thinks of apps, and how composite thinks of components. Method glue code found elsewhere. 
- Component files: `cFE_booter/osloader.c`
- Known Issues: Creating a new app is a bit of a janky process.  

#### cFE Tasks + Scheduling
- Summary: My baby. Glue for making the cFE task API compatible with composite threads. Relies on a lot of sl code under the hood. Probably needs checking for races. 
- Component files: `cFE_booter/ostask.c` `cFE_booter/ostask.h` `cFE_booter/sl_mod_fprr.c` `cFE_booter/sl_mod_policy.h` `cFE_booter/sl_thd_static_backend.c`
- Known Issues: The semaphore implementation is pure garbage. (Luckily they're not really used.) We should talk about zero allocation semaphores.

#### OSAL Filesystem APIS
- Summary: This is all Joe's doing, he'll leave a comment on it.
- Component files:  `cFE_booter/osfiles.c` `cFE_booter/osfilesys.c` `cFE_booter/osfilesys.h` `cFE_booter/tar.c` `cFE_booter/tar.h`
- Known Issues:

#### Queue Implementation
- Summary: Zach's doing. Not much interesting here, and I haven't looked at it in a long time. Appears to work fine.
- Component files: `cFE_booter/osqueue.c`
- Known Issues:

#### Major Build System Changes
- Summary: Integrating the cFE under composite took a bit of `make` wrangling. We kept most of our original build logic--in python--but it now runs totally under composite's `make`. I think most of the custom stuff is well commented.
- Component files: `.gitignore` `.gitmodules` `src/Makefile` `extern/make.py`
- Known Issues: Half of it is in python. My `make`foo is weak, so some of this is a bit of a cludge. I just don't know how to do it better.

#### cFE App Component Wrappers
- Summary: We wrap every cFE app in a component, and each of these components needs a little bit of helper code. Lots of mostly unavoidable code duplication.
- Component files: `no_interface/sch_lab/*` `no_interface/sample_lib/*` `no_interface/sample_app/*`
- Known Issues: 

#### cFE emulation for apps
- Summary: We need to provide a way for apps to call into the cFE, but intercepting the calls and writing parts into shared memory.
- Component files: `src/components/Makefile.comp` `cFE_booter/cFE_emu_support.c` `lib/cFE_emu.c` `lib/Makefile` `include/cFE_emu.h`
- Known Issues: I've just realized that this should've been done through c stubs. I'll fix that, so don't bother commenting about it.

#### cFE_booter Build System
- Summary: The cFE booter needs some some special build logic. It isn't anything too complicated, just integrating a few external object files.
- Component files: `cFE_booter/.gitignore` `cFE_booter/Makefile` 
- Known Issues:

#### Running the cFE
- Summary: These are our runscripts and Vagrantfile, how the cFE actually gets going. Not that interesting or important.
- Component files: `tools/Vagrantfile` `runscripts/cFE_booter.sh` `runscripts/llboot_cFE.sh`
- Known Issues:

#### Other Outstanding Issues  
- The cFE unit tests are currently broken. This probably should be fixed, but I'm unsure how much work it is. Likely can be done after the deadline.
- `llbooter/boot_deps.h` and `llbooter/llbooter.c` are modified to map an extra page per component. This sucks, but without it the cFE breaks.
- A bunch of stub files are changed. (As well as `include/cos_asm_simple_stacks.h`) This is because their composite versions are utterly broken and don't respect `consts.h`. I don't implement the most elegant solution, but w/e.
 
### Intent for your PR

Choose one (Mandatory):

### Reviewers (Mandatory):
@gparmer @base0x10 @phanikishoreg 

### Code Quality

- [x] I have run the auto formatter on my code before submitting this PR (see doc/auto_formatter.md for instructions)

### Testing

- the cFE


Others:
@gparmer Tried my best to summarize everything. Let me know if I missed anything important out.

gparmer:
@RobertGiff You should likely have a look at the build system stuff to see if it is compatible with the RK stuff.

src/extern/make.py
```diff
@@ -0,0 +1,169 @@
+#!/usr/bin/env python2.7
```
gparmer: yeah, this file is pretty....unique.  I'm fine with it for now, but it should not be in the `extern/` dir.  See below.

src/Makefile
```diff
@@ -1,9 +1,27 @@
...
+	$(info )
+	$(shell cd extern && python make.py >&2)
+
+init_extern:
+	$(info )
+	$(info ***********************************************)
+	$(info ********** Setting up External Libs ***********)
+	$(info ***********************************************)
+	$(info )
+	bash -c 'cd extern/cFE/; source setvars.sh; cd ../..; make -C extern/cFE/build/cpu1 config'
```
gparmer: We can hide all the python specific stuff inside a `cFS` subdir.  See comments later.

.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "src/components/lib/ps"]
 	path = src/components/lib/ps
 	url = https://github.com/gwsystems/ps.git
+[submodule "src/extern/cFE"]
+	path = src/extern/cFE
+	url = https://github.com/Others/cFE.git
```
phanikishoreg: Should this be sourced from gwsystems instead?

Others: No. This is not the cFE repository on gwsystems. (Which really should be changed back to `cFE2cos`...) This cFE repo just has cFE code in it.

src/Makefile
```diff
@@ -36,8 +54,12 @@ distclean: clean
 	$(MAKE) $(MAKEFLAGS) -C components distclean
 	@rm -f `pwd`/../transfer/*
 
-init:
+init: | ensure_config init_extern init_composite
```
phanikishoreg: Never used `|` in such a rule before. I'll need to look up what that does.

src/components/implementation/no_interface/cFE_booter/.gitignore
```diff
@@ -0,0 +1,3 @@
+gen
```
phanikishoreg: why another .gitignore in the same git repo?

src/components/implementation/no_interface/cFE_booter/Makefile
```diff
@@ -0,0 +1,12 @@
+# C_OBJS=cFE_entrypoint.o cFE_stub.o osapi.o osfiles.o osloader.o osqueue.o ostask.o ostimer.o osnetwork.o psp.o scheddev/sl.c scheddev/sl_mod_fprr.c
+ASM_OBJS=
+COMPONENT=cFE_booter.o
+INTERFACES=cFE
+DEPENDENCIES=capmgr
+IF_LIB:=./composite_cFE.o ./cFE_fs.o $(wildcard test/*.o)
+ADDITIONAL_LIBS=-lcobj_format -lcos_kernel_api -lcos_defkernel_api -lsl_capmgr -lsl_sched -lheap -lsl_lock
```
phanikishoreg: I've added a macro for `-lcos_kernel_api -lcos_defkernel_api -lsl_capmgr -lsl_sched` = LIBSLCAPMGR, you can probably use that.

src/components/implementation/no_interface/cFE_booter/cFE_emu_support.c
```diff
@@ -0,0 +1,117 @@
...
+#include <cos_component.h>
+#include <cos_debug.h>
+#include <cos_types.h>
+
+#include <memmgr.h>
+
+#include <cfe_error.h>
+
+#include <cFE_emu.h>
+
```
phanikishoreg: Too many blank lines in header inclusion.

src/components/implementation/no_interface/cFE_booter/cFE_emu_support.c
```diff
@@ -0,0 +1,117 @@
...
+union shared_region *shared_regions[16];
+
+int
+emu_backend_request_memory(spdid_t client)
+{
+	vaddr_t our_addr = 0;
+	int     id       = memmgr_shared_page_alloc(&our_addr);
+	assert(our_addr);
```
phanikishoreg: style, separation of declarations | body | return.

src/components/implementation/no_interface/cFE_booter/cFE_emu_support.c
```diff
@@ -0,0 +1,117 @@
...
+#include <memmgr.h>
+
+#include <cfe_error.h>
+
+#include <cFE_emu.h>
+
+union shared_region *shared_regions[16];
+
+int
+emu_backend_request_memory(spdid_t client)
```
phanikishoreg: should it be `cFE_emu`?

src/components/implementation/no_interface/cFE_booter/cfe_psp_eeprom.c
```diff
@@ -0,0 +1,352 @@
+/*
```
phanikishoreg: Did you have to make changes in this file? If not, perhaps just a soft-link from the extern directory would be better.

src/components/implementation/no_interface/cFE_booter/cfe_psp_memrange.c
```diff
@@ -0,0 +1,311 @@
+/*
```
phanikishoreg: ditto

src/components/implementation/no_interface/cFE_booter/ostask.c
```diff
@@ -0,0 +1,794 @@
...
+/*
+** OS-specific background thread implementation - waits forever for events to occur.
+**
+** This should be called from the BSP main routine / initial thread after all other
+** board / application initialization has taken place and all other tasks are running.
+*/
+void
+OS_IdleLoop(void)
+{
+	while (1) sl_thd_block(0);
```
phanikishoreg: should be `while (1) sl_thd_yield();` instead so if there is nothing else to run, this runs instead of blocking.
Anyone wakes this up?

src/components/implementation/no_interface/cFE_booter/sl_mod_fprr.c
```diff
@@ -0,0 +1,120 @@
+#include <llprint.h>
```
phanikishoreg: I forget why this is here and what the difference is between this and `components/lib/sl/sl_mod_fprr.c`.

phanikishoreg: If it's identical, perhaps soft-link.

src/components/implementation/no_interface/cFE_booter/sl_thd_static_backend.c
```diff
@@ -0,0 +1,69 @@
+/**
```
phanikishoreg: yeah, code duplication. use soft-links instead if this is necessary.

tools/Vagrantfile
```diff
@@ -1,19 +1,30 @@
 # -*- mode: ruby -*-
```
phanikishoreg: `tools` in composite tree? or should it be part of `cFE` tree in extern?

src/Makefile
Others: It sequences the dependencies

src/components/implementation/no_interface/cFE_booter/.gitignore
Others: It's more flexible. We don't need super long paths in the top level `.gitignore`

src/components/implementation/no_interface/cFE_booter/cFE_emu_support.c
Others: We follow the std => cos => interface => cFE => internal header convention

src/components/implementation/no_interface/cFE_booter/cFE_emu_support.c
Others: We're already in the `cFE` component. No need to namespace unnecessarily IMHO. I'm open to changing this tho.

src/components/implementation/no_interface/cFE_booter/sl_mod_fprr.c
Others: It's not identical, we need more priorities.

src/components/implementation/no_interface/cFE_booter/ostask.c
Others: This is called from the root thread, which has a super high priority. Your suggestion would imply we'd need to go through a yield every time we schedule a thread. (Even though that thread would never do anything but yield!)

tools/Vagrantfile
Others: It works for non cFE composite development. Open to changing this though.

src/components/implementation/no_interface/cFE_booter/Makefile
Others: Awesome! That's very convenient

Others:
I've addressed most of the above comments, will investigate the others. I also got rid of the cFE_emu library, and used the existing c_stubs mechanism.

@gparmer In terms of review ordering, this order would be great:

- cFE_booter Support Code
- cFE Tasks + Scheduling
- General OSAL Method Implementations
- Queue Implementation

Otherwise the original order of the summaries is a proxy for my desire to have that part reviewed. (Although, I think the filesystem should go last. It's already been reviewed after all.)

Thanks!

src/components/implementation/no_interface/cFE_booter/ostask.c
phanikishoreg: My bad, I missed the comment before that function. Yeah, yield is not the right option here.

src/components/implementation/no_interface/cFE_booter/cFE_emu_support.c
phanikishoreg: It would be better to have that as comments, so it's followed by others who change these files. 
Like,
```
/* standard includes */
xxx
...
/* composite includes */
...
/* interface includes */
...
/* cFE includes */
...
```
These undocumented groupings may not be followed otherwise.

src/components/implementation/no_interface/cFE_booter/.gitignore
phanikishoreg: I see. I'm not sure how this works, does git ignore files from both local .gitignore and the global repository .gitignore? 
Also, this is probably the longest path we can have in our build system (in terms of levels, not names), so it's not super-long if you think of it.

src/components/implementation/no_interface/cFE_booter/cFE_emu_support.c
Others: It's just a convention we follow. Plus, that's like 4 comments on headers per file. That seems pretty extra, since most of composite groups headers similarly without comments.

src/components/interface/cFE/cFE_emu.h
```diff
@@ -0,0 +1,106 @@
...
+	} cfe_es_getGenCount;
+	struct {
+		uint32 CounterId;
+		char CounterName[EMU_BUF_SIZE];
+	} cfe_es_getGenCounterIDByName;
+};
+
+int emu_backend_request_memory(spdid_t client);
+
+int32 emu_CFE_ES_GetGenCount(spdid_t client);
```
phanikishoreg: You should use the `cos_inv_token()` in the server component for the client's spdid instead.

src/components/implementation/no_interface/cFE_booter/ostask.c
```diff
@@ -0,0 +1,793 @@
...
+
+void
+OS_SchedulerStart(cos_thd_fn_t main_delegate)
+{
+	sl_init(SL_MIN_PERIOD_US);
+
+	struct sl_thd *         main_delegate_thread = sl_thd_alloc(main_delegate, NULL);
+	union sched_param_union sp = {.c = {.type = SCHEDP_PRIO, .value = MAIN_DELEGATE_THREAD_PRIORITY}};
+	sl_thd_param_set(main_delegate_thread, sp.v);
+	main_delegate_thread_id = sl_thd_thdid(main_delegate_thread);
```
gparmer: "delegate thread" is a little bit of a strange name.  Not really sure what "delegate" has to do with all of this.

src/components/implementation/no_interface/cFE_booter/ostask.c
```diff
@@ -0,0 +1,793 @@
...
+	start += sl_usec2cyc(HZ_PAUSE);
+
+	while (1) {
+		rdtscll(now);
+		if (now > start) {
+			CFE_TIME_Local1HzISR(); /* input param is signum. but CFE_PSP_TimerHandler doesn't seem to use
+			                           it. */
+		}
+
+		sl_thd_block_periodic(0);
```
gparmer: just to make sure we aren't spinning in this loop for some strange interaction in the system, it might be smart to count how many times we've iterated without triggering the `now > start` condition, and assert if it goes above some level (say 64).

src/components/implementation/no_interface/cFE_booter/cFE_emu_support.c
```diff
@@ -0,0 +1,117 @@
+#include <stdio.h>
+
+#include <cos_component.h>
+#include <cos_debug.h>
+#include <cos_types.h>
+
```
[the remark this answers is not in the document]
Others: We follow the std => cos => interface => cFE => internal header convention. Otherwise it becomes very confusing where each header is coming from. Especially since we have 5 header sources. Is this something the cos style guide takes a strong stance on?

src/components/implementation/no_interface/cFE_booter/cFE_emu_support.c
```diff
@@ -0,0 +1,117 @@
...
+	union shared_region *s = shared_regions[client];
+	return CFE_SB_CreatePipe(&s->cfe_sb_createPipe.PipeId, s->cfe_sb_createPipe.Depth,
+	                         s->cfe_sb_createPipe.PipeName);
+}
+
+void
+emu_CFE_SB_InitMsg(spdid_t client)
+{
+	union shared_region *s = shared_regions[client];
+	CFE_SB_InitMsg(s->cfe_sb_initMsg.MsgBuffer, s->cfe_sb_initMsg.MsgId, s->cfe_sb_initMsg.Length,
```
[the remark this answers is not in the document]
Others: Can you explain what you mean by that? Isn't `interface/cfe/` for the marshaling functions on the client side?

src/components/implementation/no_interface/cFE_booter/osapi.c
```diff
@@ -0,0 +1,457 @@
...
+
+#include "cFE_util.h"
+
+/*
+** Initialization of API
+*/
+int have_initialized = 0;
+
+int32
+OS_API_Init(void)
```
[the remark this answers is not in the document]
Others: It's that way because the OSAL unit tests expect this to be idempotent. We can't duplicate initialization, so I'm not sure what a better solution would look like.

src/components/implementation/no_interface/cFE_booter/osapi.c
```diff
@@ -0,0 +1,457 @@
...
+OS_time_t  local_time;
+microsec_t last_time_check;
+
+OS_time_t
+OS_AdvanceTime(OS_time_t initial_time, microsec_t usec)
+{
+	microsec_t old_seconds         = (microsec_t)initial_time.seconds;
+	microsec_t old_additional_usec = (microsec_t)initial_time.microsecs;
+
+	microsec_t old_usec = old_seconds * (1000 * 1000) + old_additional_usec;
```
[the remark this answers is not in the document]
Others: It's 64 bit.

src/components/implementation/no_interface/cFE_booter/osapi.c
```diff
@@ -0,0 +1,457 @@
...
+	 */
+
+	os_err_name_t local_name;
+	uint32        return_code = OS_SUCCESS;
+
+	if (err_name == NULL) { return (OS_INVALID_POINTER); }
+
+	switch (error_num) {
+	case OS_SUCCESS:
+		strcpy(local_name, "OS_SUCCESS");
```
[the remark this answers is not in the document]
Others: We need to return into a `os_err_name_t`. This is literally copied from the posix implementation, and I don't see a good reason to change it.

src/components/implementation/no_interface/cFE_booter/cFE_util.c
```diff
@@ -0,0 +1,32 @@
...
+panic_impl(const char *function, char *message)
+{
+	printc("cFE panic in %s: %s", function, message);
+	assert(0);
+}
+
+void
+print_with_error_name(char *message, int32 error)
+{
+	os_err_name_t local_name;
```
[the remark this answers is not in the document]
Others: This is `cFE` core stuff. There is no reasonable way for me to fix this without adding a dependency on a cFE implementation detail.

src/components/implementation/no_interface/cFE_booter/ostask.c
```diff
@@ -0,0 +1,793 @@
...
+
+#define HZ_PAUSE (1000 * 1000)
+
+void
+timer_fn_1hz(void *d)
+{
+	cycles_t now, start;
+
+	rdtscll(start);
+	start += sl_usec2cyc(HZ_PAUSE);
```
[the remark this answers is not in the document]
Others: Uniformly 64 bit.

src/components/implementation/no_interface/cFE_booter/ostask.c
Others: It's called `main_delegate_thread`. It's the thread we delegate main execution to. Do you have a better name in mind?

src/components/implementation/no_interface/cFE_booter/cFE_emu_support.c
```diff
@@ -0,0 +1,117 @@
...
+int32
+emu_CFE_SB_CreatePipe(spdid_t client)
+{
+	union shared_region *s = shared_regions[client];
+	return CFE_SB_CreatePipe(&s->cfe_sb_createPipe.PipeId, s->cfe_sb_createPipe.Depth,
+	                         s->cfe_sb_createPipe.PipeName);
+}
+
+void
+emu_CFE_SB_InitMsg(spdid_t client)
```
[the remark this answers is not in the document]
Others: No. I didn't think they should be?

src/components/implementation/no_interface/cFE_booter/ostask.c
```diff
@@ -0,0 +1,815 @@
...
+	if (is_thread_name_taken(task_name)) { return OS_ERR_NAME_TAKEN; }
+
+	if (priority > 255 || priority < 1) { return OS_ERR_INVALID_PRIORITY; }
+
+	struct sl_thd *thd = sl_thd_alloc(osal_task_entry_wrapper, function_pointer);
+	assert(thd);
+	union sched_param_union sp = {.c = {.type = SCHEDP_PRIO, .value = priority}};
+	sl_thd_param_set(thd, sp.v);
+
+	struct cfe_task_info *task_info = &cfe_tasks[sl_thd_thdid(thd)];
```
gparmer: ha, our thdid function has a pretty redundant name.

src/components/implementation/no_interface/cFE_booter/ostask.c
```diff
@@ -0,0 +1,815 @@
...
+	return OS_SUCCESS;
+}
+
+int32
+OS_TaskDelete(uint32 task_id)
+{
+	struct sl_thd *thd = sl_thd_lkup(task_id);
+	if (!thd) { return OS_ERR_INVALID_ID; }
+
+	struct cfe_task_info *task_info = &cfe_tasks[task_id];
```
gparmer: def/blocketc

src/components/implementation/no_interface/cFE_booter/ostask.c
```diff
@@ -0,0 +1,815 @@
...
+	sl_thd_free(thd);
+
+	return OS_SUCCESS;
+}
+
+uint32
+OS_TaskGetId(void)
+{
+	thdid_t real_id           = sl_thdid();
+	thdid_t possible_override = id_overrides[real_id];
```
gparmer: interesting.  not sure where this is set (will grep through), but I'm hoping there is a comment.

gparmer: OK, I understand what it is doing, but it needs a comment when it is set in `launch_other_component`.

src/components/implementation/no_interface/cFE_booter/ostask.c
```diff
@@ -0,0 +1,815 @@
...
+	sl_thd_free(sl_thd_curr());
+	PANIC("Should be unreachable!");
+}
+
+int32
+OS_TaskInstallDeleteHandler(osal_task_entry function_pointer)
+{
+	if (OS_TaskGetId() == main_delegate_thread_id) { return OS_ERR_INVALID_ID; }
+
+	struct cfe_task_info *task_info = &cfe_tasks[sl_thd_thdid(sl_thd_curr())];
```
gparmer: The whole `cfe_tasks` could/should be replaced by our own thread policy structure for cfe, right?  Are there TODOs for this, and will we add issues after merge?  I know that you want to wait to solve this properlly, if I rememer correctly.

src/components/implementation/no_interface/cFE_booter/ostask.c
```diff
@@ -0,0 +1,815 @@
...
+	task_info->delete_handler = function_pointer;
+
+	return OS_SUCCESS;
+}
+
+int32
+OS_TaskDelay(uint32 millisecond)
+{
+	cycles_t wakeup = sl_now() + sl_usec2cyc(millisecond * 1000);
+	sl_thd_block_timeout(0, wakeup);
```
gparmer: empty lines.

gparmer: wow, relative task delay APIs are broken.  Way to go cfe.  (though Posix made the same mistake).

src/components/implementation/no_interface/cFE_booter/ostask.c
```diff
@@ -0,0 +1,815 @@
...
+
+int32
+OS_TaskSetPriority(uint32 task_id, uint32 new_priority)
+{
+	if (new_priority > 255 || new_priority < 1) { return OS_ERR_INVALID_PRIORITY; }
+
+	struct sl_thd *thd = sl_thd_lkup(task_id);
+	if (!thd) { return OS_ERR_INVALID_ID; }
+
+	union sched_param_union sp = {.c = {.type = SCHEDP_PRIO, .value = new_priority}};
```
gparmer: defblockblah

src/components/implementation/no_interface/cFE_booter/ostask.c
```diff
@@ -0,0 +1,815 @@
...
+	return OS_SUCCESS;
+}
+
+
+int32
+OS_TaskGetIdByName(uint32 *task_id, const char *task_name)
+{
+	if (!task_id || !task_name) return OS_INVALID_POINTER;
+
+	thdid_t i;
```
gparmer: defblockblah

src/components/implementation/no_interface/cFE_booter/ostask.c
```diff
@@ -0,0 +1,815 @@
...
+	for (i = 1; i < SL_MAX_NUM_THDS; i++) {
+		struct sl_thd *thd = sl_thd_lkup(i);
+		if (!thd || thd->state == SL_THD_FREE) continue;
```
gparmer: empty line.

src/components/implementation/no_interface/cFE_booter/ostask.c
```diff
@@ -0,0 +1,815 @@
...
+		if (strcmp(cfe_tasks[i].osal_task_prop.name, task_name) == 0) {
+			*task_id = i;
+			return OS_SUCCESS;
```
gparmer: empty line

src/components/implementation/no_interface/cFE_booter/ostask.c
```diff
@@ -0,0 +1,815 @@
...
+
+	return OS_ERR_NAME_NOT_FOUND;
+}
+
+int32
+OS_TaskGetInfo(uint32 task_id, OS_task_prop_t *task_prop)
+{
+	if (!task_prop) { return OS_INVALID_POINTER; }
+
+	struct sl_thd *thd = sl_thd_lkup(task_id);
```
gparmer: def at head of blocks

gparmer: OK, I'm going to stop listing these so as to not be to annoying.  You can fix this throughout easily.

src/components/implementation/no_interface/cFE_booter/ostask.c
```diff
@@ -0,0 +1,815 @@
...
+
+	return OS_SUCCESS;
+}
+
+/*
+** Main thread waiting API
+*/
+
+/*
+** OS-specific background thread implementation - waits forever for events to occur.
```
gparmer: Is this a cFE comment style?  It doesn't match our typical one, or DOxygen.

src/components/implementation/no_interface/cFE_booter/ostask.c
```diff
@@ -0,0 +1,815 @@
...
+
+
+int32
+OS_MutSemCreate(uint32 *sem_id, const char *sem_name, uint32 options)
+{
+	int32 result = OS_SUCCESS;
+
+	sl_lock_take(&mutex_data_lock);
+
+	if (sem_id == NULL || sem_name == NULL) {
```
gparmer: do this and the next error check *before* taking the lock.  Always want to minimize lock hold time, while, of course, maintaining correctness.

src/components/implementation/no_interface/cFE_booter/ostask.c
```diff
@@ -0,0 +1,815 @@
...
+	sl_lock_release(&mutex_data_lock);
+	return result;
+}
+
+int32
+OS_MutSemGive(uint32 sem_id)
+{
+	int32 result = OS_SUCCESS;
+
+	sl_lock_take(&mutex_data_lock);
```
gparmer: Yeah, this hella deadlocks for the most trivial of interleavings.  Bummer.

src/components/implementation/no_interface/cFE_booter/ostask.c
```diff
@@ -0,0 +1,815 @@
...
+
+	if (sem_id >= OS_MAX_MUTEXES || !mutexes[sem_id].used) {
+		result = OS_ERR_INVALID_ID;
+		goto exit;
+	}
+
+	sl_lock_take(&mutexes[sem_id].lock);
```
gparmer: A simple solution that doesn't really scale in complexity is to just get rid of the `mutex_data_lock`, and take each mutex's lock separately.  Since we're in a static data-structure, this is likely fine.

src/components/implementation/no_interface/cFE_booter/ostask.c
```diff
@@ -0,0 +1,815 @@
...
+	sl_lock_release(&mutex_data_lock);
+
+	return result;
+}
+
+int32
+OS_MutSemDelete(uint32 sem_id)
+{
+	int32 result = OS_SUCCESS;
+	sl_lock_take(&mutex_data_lock);
```
gparmer: empty line before

src/components/implementation/no_interface/cFE_booter/ostask.c
```diff
@@ -0,0 +1,815 @@
...
+}
+
+int32
+OS_MutSemGetIdByName(uint32 *sem_id, const char *sem_name)
+{
+	int32 result = OS_SUCCESS;
+
+	sl_lock_take(&mutex_data_lock);
+
+	if (sem_id == NULL || sem_name == NULL) {
```
gparmer: move outside of lock along with next error check

src/components/implementation/no_interface/cFE_booter/ostask.c
```diff
@@ -0,0 +1,815 @@
...
+
+	int i;
+	for (i = 0; i < OS_MAX_MUTEXES; i++) {
+		if (mutexes[i].used && (strcmp(mutexes[i].prop.name, (char *)sem_name) == 0)) {
+			*sem_id = i;
+			goto exit;
+		}
+	}
+
+	/* The name was not found in the table,
```
gparmer:
```
/*
 * ...
 */
```

src/components/implementation/no_interface/cFE_booter/ostask.c
```diff
@@ -0,0 +1,815 @@
...
+// Generic semaphore methods
+int32
+OS_SemaphoreCreate(struct semaphore *semaphores, uint32 max_semaphores, uint32 *sem_id, const char *sem_name,
+                   uint32 sem_initial_value, uint32 options)
+{
+	int32 result = OS_SUCCESS;
+
+	sl_lock_take(&semaphore_data_lock);
+
+	if (sem_id == NULL || sem_name == NULL) {
```
gparmer: move these error checks out of the lock

gparmer: since these checks are repetitive, you might want to move them into a function.

src/components/implementation/no_interface/cFE_booter/ostask.c
```diff
@@ -0,0 +1,815 @@
...
+	sl_lock_release(&semaphore_data_lock);
+
+	return result;
+}
+
+int32
+OS_SemaphoreDelete(struct semaphore *semaphores, uint32 max_semaphores, uint32 sem_id)
+{
+	int32 result = OS_SUCCESS;
+	sl_lock_take(&semaphore_data_lock);
```
gparmer: empty line

src/components/implementation/no_interface/cFE_booter/cFE_emu_support.c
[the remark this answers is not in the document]
gparmer: The functions on the client and server side of an invocation through an interface function should be in sstubs.c and cstubs.c in the interface.  I'm not sure if that is what these are doing.

src/components/implementation/no_interface/cFE_booter/osapi.c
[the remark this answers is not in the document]
gparmer: I don't understand either of these justifications.  `*err_name` is a `os_err_name_t`.  If posix did something dumb I don't see a reason to copy it.  There might be other arguments; these seems lazy.  I don't have time to go through all comments to see if they are making good justifications or not, so this isn't heartening.

src/components/implementation/no_interface/cFE_booter/ostask.c
```diff
@@ -0,0 +1,815 @@
...
+	sl_thd_param_set(timer_thd, spperiod.v);
+	sl_thd_param_set(timer_thd, spprio.v);
+
+	sl_sched_loop();
+}
+
+void
+osal_task_entry_wrapper(void *task_entry)
+{
+	((osal_task_entry)task_entry)();
```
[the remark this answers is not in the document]
Others: Unfortunately, this is a cFE definition

src/components/implementation/no_interface/cFE_booter/ostask.c
Others: I really didn't want to spend time fixing sl, when we're not even sure what the design should look like. I switched to this implementation, where the cFE tracks its own task info. It really isn't much of a hack anyway.

If you think I should do it the other way, then I can spend some time to figure that out. At least this new implementation does not involve memory corruption.

src/components/implementation/no_interface/cFE_booter/ostask.c
Others: This is from the cFE. Their comment style is all over the place...

PR #341, opened 2018-03-28
Enable tls management through capmgr component

RobertGiff:

This PR adds tls management to the capmgr component. Upon boot, the capmgr component maps a tls region for every possible thread in the system. Each thread, (identified with thread id), has a virtual address offset from 0x7000... where there will be guaranteed mapped pages. Each component has different mapped pages into this region. There are two interfaces which other components can manage their tls regions. The first simply returns the virtual address at which the component has its TLS range. This lets components like the rumpkernel who already manage their own tls continue to do this with minimal changes. Capmgr component tls management for the RK is debugged and will be coming in a later PR. The second interface included in this PR is for components to set the current running thread tls region. This enables components that need Rust and other tls related necessities to easily set up their tls region when needed. **This second interface has not yet been debugged for use with Rust.** I will be working with @maloneya to debug when he begins using this for his Rust components.

### Intent for your PR

Choose one (Mandatory):

### Reviewers (Mandatory):
@gparmer, @Others, @phanikishoreg 

### Code Quality

### Testing

- rumpkern_boot.sh
- llbooter_pong.sh
- llbooter_test.sh
- micro_boot.sh
- unit_capmgrtest.sh
- unit_hierschedcomps.sh
- unit_schedcomp.sh
- unit_schedtests.sh

### DIFF
--- a/src/components/implementation/capmgr/naive/init.c
+++ b/src/components/implementation/capmgr/naive/init.c
@@ -19,14 +19,15 @@ capmgr_comp_info_iter(void)
 	do {
 		spdid_t spdid = 0, sched_spdid = 0;
 		struct cap_comp_info *rci = NULL;
+		struct cos_compinfo tmp;
 		struct sl_thd *ithd = NULL;
 		u64_t chbits = 0, chschbits = 0;
 		pgtblcap_t pgtslot = 0;
 		captblcap_t captslot = 0;
 		compcap_t ccslot = 0;
-		vaddr_t vasfr = 0;
+		vaddr_t vasfr = 0, src_pg = 0;
 		capid_t capfr = 0;
-		int ret = 0, is_sched = 0;
+		int ret = 0, is_sched = 0, offset = 0;
 		int remain_child = 0;
 		spdid_t childid;
 		comp_flag_t ch_flags;
@@ -71,6 +72,26 @@ capmgr_comp_info_iter(void)
 			cap_info_initthd_init(rci, ithd, 0);
 		} else if (cos_spd_id() == spdid) {
 			cap_info_initthd_init(rci, sl__globals()->sched_thd, 0);
+		}
+
+		/*
+		 * Set tls region for this component
+		 * Expand the 2nd level pte within this component's page table at our new range
+		 */
+		tmp.pgtbl_cap = pgtslot;
+		tmp.memsrc    = ci;
+		ret = (int)cos_pgtbl_intern_alloc(&tmp, pgtslot, TLS_BASE_ADDR,
+					     PAGE_SIZE * TLS_NUM_PAGES);
+		assert(ret);
+
+		/* Place desired number of pages at this new range */
+		offset = 0;
+		for (i = 0 ; i < (signed)TLS_NUM_PAGES ; i++) {
+			src_pg = (vaddr_t)cos_page_bump_alloc(ci);
+			assert(src_pg);
+			ret = cos_mem_alias_at(&tmp, TLS_BASE_ADDR + offset, ci, src_pg);
+			assert(!ret);
+			offset += PAGE_SIZE;
 		}
 	} while (remaining > 0);
 
--- a/src/components/implementation/capmgr/naive/mem_mgr.c
+++ b/src/components/implementation/capmgr/naive/mem_mgr.c
@@ -65,3 +65,35 @@ memmgr_shared_page_map_cserialized(vaddr_t *pgaddr, int *unused, int idx)
 done:
 	return num_pages;
 }
+
+/* These functions return the address at which the gs register is set to */
+void *
+memmgr_tls_alloc(unsigned int dst_tid)
+{
+	/* Just return the vaddr range for this tid */
+	assert(dst_tid < MAX_NUM_THREADS);
+	return TLS_BASE_ADDR + (void *)(TLS_AREA_SIZE * dst_tid);
+}
+
+void *
+_memmgr_tls_alloc_and_set(void *area)
+{
+	spdid_t cur = cos_inv_token();
+	struct cap_comp_info  *cur_rci = cap_info_comp_find(cur);
+	struct cos_compinfo  *rci      = cap_info_ci(cur_rci);
+	struct cos_compinfo  *cur_ci   = cos_compinfo_get(cos_defcompinfo_curr_get());
+	void *addr;
+	unsigned int dst_thdcap;
+	int tid;
+
+	assert(cur_ci && rci);
+
+	dst_thdcap = sl_thd_curr()->aepinfo->thd;
+	tid        = cos_introspect(cur_ci, dst_thdcap, THD_GET_TID);
+	addr       = memmgr_tls_alloc(tid);
+	assert(addr);
+
+	cos_thd_mod(rci, dst_thdcap, addr);
+
+	return addr;
+}
--- a/src/components/interface/capmgr/memmgr.h
+++ b/src/components/interface/capmgr/memmgr.h
@@ -1,6 +1,7 @@
 #ifndef MEMMGR_H
 #define MEMMGR_H
 
+#include <cos_kernel_api.h>
 #include <cos_types.h>
 
 vaddr_t memmgr_heap_page_alloc(void);
--- a/src/components/interface/capmgr/memmgr.h
+++ b/src/components/interface/capmgr/memmgr.h
@@ -10,4 +10,23 @@ int memmgr_shared_page_alloc(vaddr_t *pgaddr);
 int memmgr_shared_page_allocn(int num_pages, vaddr_t *pgaddr);
 int memmgr_shared_page_map(int id, vaddr_t *pgaddr);
 
+/* This magic number is double the tls size defined in RK */
+#define TLS_AREA_SIZE 32
+#define TLS_NUM_PAGES (round_up_to_page(TLS_AREA_SIZE * MAX_NUM_THREADS) / PAGE_SIZE)
+#define TLS_BASE_ADDR 0x70000000
+
+void *memmgr_tls_alloc(unsigned int dst_tid);
+void *_memmgr_tls_alloc_and_set(void *area);
+
+static void *
+memmgr_tls_alloc_and_set(void *area)
+{
+	void *addr = _memmgr_tls_alloc_and_set(area);
+
+	/* Set area within addr for this tid */
+	*(void **)addr = area;
+
+	return addr;
+}
+
 #endif /* MEMMGR_H */
--- a/src/components/lib/posix/posix.c
+++ b/src/components/lib/posix/posix.c
@@ -250,38 +250,14 @@ cos_set_tid_address(int *tidptr)
 	return 0;
 }
 
-/* struct user_desc {
- *     int  		  entry_number; // Ignore
- *     unsigned int  base_addr; // Pass to cos thread mod
- *     unsigned int  limit; // Ignore
- *     unsigned int  seg_32bit:1;
- *     unsigned int  contents:2;
- *     unsigned int  read_exec_only:1;
- *     unsigned int  limit_in_pages:1;
- *     unsigned int  seg_not_present:1;
- *     unsigned int  useable:1;
- * };
+/*
+ * TLS is managed by capmgr component, if something needs tls, it should use capmgr
+ * Stub needed as libc_init expects there to be something to setup tls, although it doesn't
+ * do anything with tls immediately
  */
-
-void* backing_data[SL_MAX_NUM_THDS];
-
-static void
-setup_thread_area(struct sl_thd *thread, void* data)
-{
-	struct cos_compinfo *ci = cos_compinfo_get(cos_defcompinfo_curr_get());
-	thdid_t thdid = sl_thd_thdid(thread);
-
-	backing_data[thdid] = data;
-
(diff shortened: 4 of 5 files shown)


src/components/interface/capmgr/memmgr.h
```diff
@@ -10,4 +11,23 @@ int memmgr_shared_page_alloc(vaddr_t *pgaddr);
...
+void *memmgr_tls_alloc(unsigned int dst_tid);
+void *_memmgr_tls_alloc_and_set(void *area);
+
+static void *
+memmgr_tls_alloc_and_set(void *area)
+{
+	void *addr = _memmgr_tls_alloc_and_set(area);
+
+	/* Set area within addr for this tid */
+	*(void **)addr = area;
```
maloneya: i think im missing something here. You are casting a void * to a  void ** an then dereferencing? is that not the same as just addr = area ?

src/components/interface/capmgr/memmgr.h
```diff
@@ -10,4 +11,23 @@ int memmgr_shared_page_alloc(vaddr_t *pgaddr);
...
+/* This magic number is double the tls size defined in RK */
+#define TLS_AREA_SIZE 32
+#define TLS_NUM_PAGES (round_up_to_page(TLS_AREA_SIZE * MAX_NUM_THREADS) / PAGE_SIZE)
+#define TLS_BASE_ADDR 0x70000000
+
+void *memmgr_tls_alloc(unsigned int dst_tid);
+void *_memmgr_tls_alloc_and_set(void *area);
+
+static void *
+memmgr_tls_alloc_and_set(void *area)
```
maloneya: As far as i can tell area is never used

src/components/implementation/capmgr/naive/init.c
```diff
@@ -19,14 +19,15 @@ capmgr_comp_info_iter(void)
 	do {
 		spdid_t spdid = 0, sched_spdid = 0;
 		struct cap_comp_info *rci = NULL;
+		struct cos_compinfo tmp;
```
phanikishoreg: just tmp? perhaps tmpci?

src/components/implementation/capmgr/naive/init.c
```diff
@@ -72,6 +73,26 @@ capmgr_comp_info_iter(void)
...
 		}
+
+		/*
+		 * Set tls region for this component
+		 * Expand the 2nd level pte within this component's page table at our new range
+		 */
+		tmp.pgtbl_cap = pgtslot;
+		tmp.memsrc    = ci;
+		ret = (int)cos_pgtbl_intern_alloc(&tmp, pgtslot, TLS_BASE_ADDR,
+					     PAGE_SIZE * TLS_NUM_PAGES);
```
phanikishoreg: incorrect spacing here.. should start under `&tmp`.

src/components/implementation/capmgr/naive/init.c
```diff
@@ -72,6 +73,26 @@ capmgr_comp_info_iter(void)
...
+		assert(ret);
+
+		/* Place desired number of pages at this new range */
+		offset = 0;
+		for (i = 0 ; i < (signed)TLS_NUM_PAGES ; i++) {
```
phanikishoreg: `(signed)` isn't telling what the type of `i` or the cast is.. being more explicit, like `(int)`?

src/components/implementation/capmgr/naive/mem_mgr.c
```diff
@@ -65,3 +65,35 @@ memmgr_shared_page_map_cserialized(vaddr_t *pgaddr, int *unused, int idx)
 done:
 	return num_pages;
 }
+
+/* These functions return the address at which the gs register is set to */
+void *
+memmgr_tls_alloc(unsigned int dst_tid)
```
phanikishoreg: Why is it called `alloc()` if it's not doing any `allocation`? 
Why not use the type `thdid_t` here? 

Shouldn't we check if the client (calling) component has been initialized with `capmgr` ??

src/components/implementation/capmgr/naive/mem_mgr.c
```diff
@@ -65,3 +65,35 @@ memmgr_shared_page_map_cserialized(vaddr_t *pgaddr, int *unused, int idx)
...
+{
+	/* Just return the vaddr range for this tid */
+	assert(dst_tid < MAX_NUM_THREADS);
+	return TLS_BASE_ADDR + (void *)(TLS_AREA_SIZE * dst_tid);
+}
+
+void *
+_memmgr_tls_alloc_and_set(void *area)
```
phanikishoreg: Why is the weird naming `_memmgr_xxxx`? 

Passing address `area` between address-spaces? If it's not used, why pass it?

src/components/interface/capmgr/memmgr.h
```diff
@@ -10,4 +11,23 @@ int memmgr_shared_page_alloc(vaddr_t *pgaddr);
 int memmgr_shared_page_allocn(int num_pages, vaddr_t *pgaddr);
 int memmgr_shared_page_map(int id, vaddr_t *pgaddr);
 
+/* This magic number is double the tls size defined in RK */
+#define TLS_AREA_SIZE 32
+#define TLS_NUM_PAGES (round_up_to_page(TLS_AREA_SIZE * MAX_NUM_THREADS) / PAGE_SIZE)
+#define TLS_BASE_ADDR 0x70000000
```
phanikishoreg: None of these should be here. It is perhaps implementation specific, so either have it `capmgr` implementation if it isn't used outside that or have it somewhere like `cos_types.h` if it's required outside of the implementation.

src/components/interface/capmgr/memmgr.h
phanikishoreg: `area` is confusing here.. Do you mean `data` ?? 
@maloneya I think it's setting the passed `data` (here `area`) to the `TLS AREA` that is allocated by the memmgr. 
I don't see a utility of `_set` here.. 
Interface users get the address and they can just set the data there..

src/components/lib/posix/posix.c
```diff
@@ -293,7 +269,9 @@ cos_clone(int (*func)(void *), void *stack, int flags, void *arg, pid_t *ptid, v
 
 	struct sl_thd * thd = sl_thd_alloc((cos_thd_fn_t) func, arg);
 	if (tls) {
-		setup_thread_area(thd, tls);
+		/* Make sure TLS is managed by */
```
phanikishoreg: by ? ;-)

src/components/implementation/capmgr/naive/init.c
```diff
@@ -72,6 +73,26 @@ capmgr_comp_info_iter(void)
...
+			src_pg = (vaddr_t)cos_page_bump_alloc(ci);
+			assert(src_pg);
+			ret = cos_mem_alias_at(&tmp, TLS_BASE_ADDR + offset, ci, src_pg);
+			assert(!ret);
+			offset += PAGE_SIZE;
+		}
```
phanikishoreg: I'd have all of this in `cap_info.h/.c` API like cap_info_comptls_init() or something and call that! There is so much code already in this function, I'd avoid expanding it when there are other alternatives.. 

Better, I'd call that `cap_info_comptls_init()` in `cap_info_comp_init` function..

src/components/interface/capmgr/memmgr.h
RobertGiff: @phanikishoreg I see no reason to leave the responsibility to set the data up to the user of the interface. I agree that data would be a much better name though!
@maloneya casting addr as a void double pointer, and dereferencing that makes addr point to area, it doesn't set the addresses to equal the same thing.

RobertGiff: The real purpose of set however is that cos_thd_mod is called and sets the content of the gs register with addr.

src/components/interface/capmgr/memmgr.h
RobertGiff: It is only used above. As @phanikishoreg said, we could leave it up to the interface user to set what addr points to on their own.

src/components/implementation/capmgr/naive/mem_mgr.c
RobertGiff: "Shouldn't we check if the client (calling) component has been initialized with capmgr ??" - We should, do you have an example of where we do that currently?

The name choice of alloc is bad.

src/components/implementation/capmgr/naive/mem_mgr.c
RobertGiff: It should not be passed in here

src/components/implementation/capmgr/naive/init.c
RobertGiff: Thanks for the advice on this, being unfamiliar with a lot of the changes to the llbooter, I had no idea where it would be ideal to place this initialization code. I selected this location as it worked haha

src/components/interface/capmgr/memmgr.h
phanikishoreg: If the real purpose of this API is to call `cos_thd_mod`, shouldn't we pass in the `thdid`?? Or is this only for the current thread that we intend to use?? 

We'd want to setup the TLS before running a thread right? So, I'd think that we also want to pass in the thdid..

src/components/implementation/capmgr/naive/mem_mgr.c
```diff
@@ -65,3 +65,35 @@ memmgr_shared_page_map_cserialized(vaddr_t *pgaddr, int *unused, int idx)
 done:
 	return num_pages;
 }
+
+/* These functions return the address at which the gs register is set to */
+void *
+memmgr_tls_alloc(unsigned int dst_tid)
+{
+	/* Just return the vaddr range for this tid */
+	assert(dst_tid < MAX_NUM_THREADS);
```
phanikishoreg: We should differentiate from `VALIDATING INPUT PARAMETERS` from the capmgr internal data consistencies.. So if `dst_tid` passed is invalid, we'd not want an assert here, instead return a failure..

src/components/implementation/capmgr/naive/mem_mgr.c
phanikishoreg: `if the client (calling) component has been initialized with capmgr` is checked in every existing interface implementation API in capmgr. 

For ex (from `memmgr_heap_page_allocn()`):
```
        spdid_t cur = cos_inv_token();
        struct cos_compinfo  *cap_ci  = cos_compinfo_get(cos_defcompinfo_curr_get());
        struct cap_comp_info *cur_rci = cap_info_comp_find(cur);
        struct cos_compinfo  *cur_ci  = cap_info_ci(cur_rci);
        vaddr_t src_pg, dst_pg;
        unsigned int off = 0;

        if (!cur_rci || !cap_info_init_check(cur_rci)) return 0;
        if (!cur_ci) return 0;
```

src/components/implementation/capmgr/naive/mem_mgr.c
```diff
@@ -65,3 +65,35 @@ memmgr_shared_page_map_cserialized(vaddr_t *pgaddr, int *unused, int idx)
...
+	struct cap_comp_info  *cur_rci = cap_info_comp_find(cur);
+	struct cos_compinfo  *rci      = cap_info_ci(cur_rci);
+	struct cos_compinfo  *cur_ci   = cos_compinfo_get(cos_defcompinfo_curr_get());
+	void *addr;
+	unsigned int dst_thdcap;
+	int tid;
+
+	assert(cur_ci && rci);
+
+	dst_thdcap = sl_thd_curr()->aepinfo->thd;
```
phanikishoreg: use `sl_thd_thdcap` here..

src/components/implementation/capmgr/naive/mem_mgr.c
```diff
@@ -65,3 +65,35 @@ memmgr_shared_page_map_cserialized(vaddr_t *pgaddr, int *unused, int idx)
...
+	tid        = cos_introspect(cur_ci, dst_thdcap, THD_GET_TID);
```
phanikishoreg: `sl_thd_thdid()` here

src/components/implementation/capmgr/naive/mem_mgr.c
phanikishoreg: I don't see a benefit of having 2 interfaces one making the caller (client) do cos_thd_mod and the other doing it from the `capmgr`. 

The point of the first interface I thought originally was really just get the address for being able to access it in your component. (It's my bad I didn't follow the PR description thoroughly).

We talked about it briefly in person but I still think we should just have a single interface. 
I've a suggestion above, which is to make `cos_thd_mod` be part of the thread creation and this interface is only for getting the address! That would be great. If we were to have a tls manager as a separate component, then this detail of actually cos_thd_mod() could still be hidden in that one interface for tls allocation or set! But I'm not sure if we'd want a standalone tls manager either..

src/components/interface/capmgr/memmgr.h
gparmer: I do *not* want all components with thread capabilities to be able to set `gs` as it has *global* impacts across multiple components.  This is clearly the responsibility of a shared manager component.  In the future, we will allow only a subset of operations on specific thread capabilities, and the only component that should be able to set `gs` is the capmgr (in this case).

PR #372, opened 2018-08-31
Implementation of user-level fault handler

WenyuanShao:

This PR is the implementation of user_level fault handlers. When a fault happens, it will make a sinv call to the user_level fault handler to handle the fault.

1. Modified the invstk and set a bit into the captbl of the component on the top of the stack to indicate whether this is a sinv call to fault handlers or not. Add some additional comments to make this design easier to understand. 
2. Created a new type for sinv cap to the fault handlers, so we can get rid of a branch in the fast path.
3. Expanded the thd_introspect to get the register information in user_level and print these register information when fault happens.
4. Create an interface for fault handlers so different components can implement fault handlers differently.

### Intent for your PR

Choose one (Mandatory):

### Reviewers (Mandatory):
@gparmer @hungry-foolish @phanikishoreg 

### Code Quality

### Testing

-llbooter_test.sh, llbooter_pong.sh, microboot.sh


src/components/implementation/no_interface/llbooter/boot_deps.h
```diff
@@ -113,6 +123,92 @@ boot_spd_initaep_get(spdid_t spdid)
...
+			(unsigned int)fault_regs.ip, (unsigned int)fault_regs.sp, (unsigned int)fault_regs.bp);
+	printc("Indicator->EFLAGS: %x\n", (unsigned int)fault_regs.flags);
+	printc("(Exception Error Code->ORIG_AX: %x)\n", (unsigned int)fault_regs.orig_ax);
+}
+
+/* The fault handlers*/
+void
+fault_div_by_zero(unsigned long sp, unsigned long ip, unsigned long fault_addr, unsigned long fault_type)
+{
+	PRINTLOG(PRINT_DEBUG, "in div by zero error fault handler, fault happens in component:%u\n\n", cos_inv_token());
```
phanikishoreg: You should really use `PRINT_ERROR` here.

src/components/implementation/no_interface/llbooter/boot_deps.h
```diff
@@ -143,6 +239,35 @@ boot_capmgr_mem_alloc(void)
 	cos_meminfo_alloc(capmgr_info, BOOT_MEM_KM_BASE, mem_sz);
 }
 
+/* 
+ * The sinv cap belongs to a different type.
+ * So we can get rid of an extra branch in fast path.
+ */
```
phanikishoreg: Probably place the comment where the `cap` you're referring to is created? 
Or rephrase this to something like: "all `sinv caps` created here are `fault sinv caps`, a different capability type to avoid an extra branch in the invocation fast-path"..

src/kernel/include/shared/cos_types.h
```diff
@@ -131,6 +132,7 @@ typedef enum {
 typedef enum {
 	CAP_FREE = 0,
 	CAP_SINV,       /* synchronous communication -- invoke */
+	CAP_SINVFLT,    /* synchronous communication of fault handler -- invoke*/
```
phanikishoreg: Because it is really not an "INVOCATION" from USER-LEVEL, I think it can just be named `CAP_FAULT` or `CAP_FLT`

src/kernel/include/thd.h
```diff
@@ -18,8 +18,22 @@
 #include "tcap.h"
 #include "list.h"
 
-struct invstk_entry {
+/* 
+ * 1. This data structure is designed to replace the comp_info used in invstk_entry.
+ * The reason for using comp_invstk_info instead of comp_info is we want to set a flag
+ * into the captbl pointer of comp_info to achieve better alignment.
```
phanikishoreg: You could indent the comments for readability.. just a nitpick..

src/kernel/include/thd.h
```diff
@@ -18,8 +18,22 @@
...
+ * 2. We plan to set a bit into the captbl pointer because captbls are perfectly aligned 
+ * which means the last few bits of captbl is always "0". By this design, we can achieve
+ * both better cache line alignment and structure alignment.
```
phanikishoreg: I'm not sure I understand this comment.. 

If we have it "perfectly aligned", why would the last few bits in the captbl be "0"?? "perfectly aligned" I read it as, the size of the captbl is a multiple of the cache line size or is a single cache line.. so there is no padding in which case, why would you have "always 0" in the last few bits??

src/kernel/include/thd.h
```diff
@@ -472,17 +498,30 @@ curr_invstk_top(struct cos_cpu_local_info *cos_info)
 	return cos_info->invstk_top;
 }
 
+/* 
+ * This function aims to output the "in_fault" flag and return comp_info with the bit in captbl reset.
+ * Refer to the comments of comp_invstk_info to see more about this design.
```
phanikishoreg: I don't think you need that comment on "refer to the comments...." for each of the function that is part of the fault-handling mechanism..

phanikishoreg: sorry.. a nitpick again..

src/kernel/include/thd.h
```diff
@@ -472,17 +498,30 @@ curr_invstk_top(struct cos_cpu_local_info *cos_info)
...
+ */
 static inline struct comp_info *
-thd_invstk_current(struct thread *curr_thd, unsigned long *ip, unsigned long *sp, struct cos_cpu_local_info *cos_info)
+thd_invstk_current_fault(struct thread *curr_thd, unsigned long *ip, unsigned long *sp, unsigned long *in_fault, struct cos_cpu_local_info *cos_info)
```
phanikishoreg: Signature is so jumbled.. You'd want to order the "INPUT args" and "OUTPUT args" correctly..

src/kernel/include/thd.h
```diff
@@ -493,34 +532,56 @@ thd_current_pgtbl(struct thread *thd)
 	/* don't use the cached invstk_top here. We need the stack
 	 * pointer of the specified thread. */
 	curr_entry = &thd->invstk[thd->invstk_top];
-	return curr_entry->comp_info.pgtbl;
+	return curr_entry->comp_invstk_info.comp_info.pgtbl;
```
phanikishoreg: blank line before this plz. I know the style used here isn't part of your change but you might as well fix it.

src/kernel/include/thd.h
```diff
@@ -493,34 +532,56 @@ thd_current_pgtbl(struct thread *thd)
...
+/*
+ * This function will be called by thd_invstk_push to modify the current top entry of invstk.
+ * This modification will include the modification of the ip, sp, and in_fault flag.
+ */
+static inline void
+thd_invstk_modify_current(struct thread *thd, unsigned long ip, unsigned long sp, unsigned long in_fault, 
+						  struct cos_cpu_local_info *cos_info)
+{
+	struct invstk_entry *    curr_entry;
+	struct comp_info *       curr_comp_info;
```
phanikishoreg: unnecessary extra spacing within the declarations.

src/kernel/include/thd.h
```diff
@@ -493,34 +532,56 @@ thd_current_pgtbl(struct thread *thd)
...
+	curr_comp_info = &curr_entry->comp_invstk_info.comp_info;
+	curr_entry->ip = ip;
+	curr_entry->sp = sp;
+
+	curr_comp_info->captbl = (struct captbl*)(((unsigned long)curr_comp_info->captbl & (~1u)) | in_fault);
 }
 
+/*
+ * The push operation of invstk has been divided into two steps. First, modify the top of the invstk
+ * by calling the thd_invstk_modify_current fucntion. Then push a new invstk_entry into the invstk.
```
phanikishoreg: nitpicks: 
1. fucntion vs function typo.
2. Do you really need that comment here? Or I think the code is readable to understand that it first modifies the top and then copies the new entry.

src/kernel/include/thd.h
```diff
@@ -493,34 +532,56 @@ thd_current_pgtbl(struct thread *thd)
...
-	memcpy(&top->comp_info, ci, sizeof(struct comp_info));
+	memcpy(&top->comp_invstk_info.comp_info, ci, sizeof(struct comp_info));
 	top->ip = top->sp = 0;
 
 	return 0;
 }
 
 static inline struct comp_info *
-thd_invstk_pop(struct thread *thd, unsigned long *ip, unsigned long *sp, struct cos_cpu_local_info *cos_info)
+thd_invstk_pop(struct thread *thd, unsigned long *ip, unsigned long *sp, unsigned long *in_fault, struct cos_cpu_local_info *cos_info)
```
phanikishoreg: not your fault again.. sorry but the input and output args are just mixed.. 
you probably want input args first and then output args or vice versa..

src/kernel/include/thd.h
```diff
@@ -577,15 +638,42 @@ thd_switch_update(struct thread *thd, struct pt_regs *regs, int issame)
...
+		THD_GET(THD_GET_FAULT_REG8, retval, t->fault_regs.fs); \
+		THD_GET(THD_GET_FAULT_REG9, retval, t->fault_regs.gs); \
+		THD_GET(THD_GET_FAULT_REG10, retval, t->fault_regs.ss); \
+		THD_GET(THD_GET_FAULT_REG11, retval, t->fault_regs.si); \
+		THD_GET(THD_GET_FAULT_REG12, retval, t->fault_regs.di); \
+		THD_GET(THD_GET_FAULT_REG13, retval, t->fault_regs.ip); \
+		THD_GET(THD_GET_FAULT_REG14, retval, t->fault_regs.sp); \
+		THD_GET(THD_GET_FAULT_REG15, retval, t->fault_regs.bp); \
+		THD_GET(THD_GET_FAULT_REG16, retval, t->fault_regs.flags); \
+		THD_GET(THD_GET_FAULT_REG17, retval, t->fault_regs.orig_ax); \
```
phanikishoreg: you love macros.. don't you? :-P

I think we should stick to the standard and not use macros here.. If you're just getting and returning values its fine but if there is a case in the future to do something more in some "given case" and not all, then you'd have to expand the switch case back.. 

Also I don't think there is any benefit of doing this here.. The number of characters in your "case" macro is less (statement could be inaccurate) than the number of characters in the macro call you have. :-)

src/platform/i386/exception.c
```diff
@@ -7,6 +7,7 @@
 #include "chal_cpu.h"
 
 #define PRINTK(format, ...) printk("(CPU%ld:) " format, get_cpuid(), ## __VA_ARGS__)
+#define FLT_SINV(regs, cap) fault_handler_sinv(regs, cap)
```
phanikishoreg: Yeah, I bet.. You love macros.. :P

phanikishoreg:
Oh well.. Didn't notice, you need to pull the latest and fix the conflicts..

phanikishoreg:
Just curious, why haven't you checked:
```
[ ] I've commented appropriately where code is tricky
[ ] I agree that there is no "throw-away" code, and that code in this PR is of high quality
```
If you don't, I think you should mention somewhere in the PR why! Helps reviewers.. ;)

And another nitpick:
```
[ ] This PR is for a code-review and is intended to get feedback, but not to be pulled yet.
[ ] This PR is mature, and ready to be integrated into the repo.
```
You should check one of these boxes only only.. Should really be treated as a radio-button.. :P

src/kernel/include/shared/cos_types.h
```diff
@@ -223,40 +226,55 @@ captbl_idsize(cap_t c)
...
+ * 32-33 = this captbl,
+ * 34-35 = our pgtbl root,
+ * 36-39 = our component,
+ * 40-41 = vm pte for booter
+ * 42-43 = untyped memory pgtbl root,
+ * 44-45 = vm pte for physical memory,
+ * 46-47 = km pte,
+ * 48-49 = comp0 captbl,
+ * 50-51 = comp0 pgtbl root,
+ * 52-55 = comp0 component,
```
phanikishoreg: Just another nitpick: There are some of these caps that can be removed from the table.. And some that only need to be part of the booter.. If we can somehow rearrange the capabilities so the ones needed by the booter are past some "COMP_FREE" offset, then only llbooter will need to set it's captbl to the LLBOOTER_COMP_FREE and others can reuse those..

1. PTE caps are only required by the booter. 
2. COMP0 caps are not used at all in the new system in my understanding.. we could get rid of them.. 
3. Perhaps the fault capabilities are not required in the booter's captbl.. 
We can rearrange them I'm sure.. Can save some capabilities but honestly I'm thinking of it as more of a code cleanup..

src/kernel/include/inv.h
```diff
@@ -303,28 +303,36 @@ sinv_call(struct thread *thd, struct cap_sinv *sinvc, struct pt_regs *regs, stru
...
 		// FIXME: add fault handling here.
 		__userregs_set(regs, -EFAULT, __userregs_getsp(regs), __userregs_getip(regs));
-		return;
+		goto ret;
 	}
 
 	pgtbl_update(ci->pgtbl);
 	/* Set return sp and ip and function return value in eax */
 	__userregs_set(regs, __userregs_getinvret(regs), sp, ip);
+ret:
```
phanikishoreg: Wait.. I thought you said we're saving a branch in the fast-path..? But we're not here.

src/components/implementation/no_interface/llbooter/boot_deps.h
phanikishoreg: There isn't one.
I've only added `PRINT_DEBUG`, `PRINT_WARN` and `PRINT_ERROR`.

src/kernel/include/shared/cos_types.h
phanikishoreg: Not all `>32`. 
Ones with `PTE` in the enums are only populated by the kernel for the booter. I think they're not used by any component.. not even booter **uses** it.
`SELF_COMP`, I guess with capmgr as capability manager, I believe other components may not need it but I don't know if we want to get rid of it. 

I think booter is a special case and I think it is absolutely fine because I think we very well know that booter is a "special case". For ex, that `BOOT_CAPTBL_SINV_CAP` capability is not populated for booter but it just exists and wasting one slot. 

We're not going to have any **abstraction** that special cases booter really so I don't think having BOOTER_XX_CAP and COMP_XX_CAP is really harmful. I think it would self document better, in that it says, booter is indeed special and it has more "low level" capabilities. 

I don't have an argument for worthiness of the time and effort that would be put in but..

src/kernel/include/shared/cos_types.h
```diff
@@ -266,6 +284,19 @@ enum
 	BOOT_CAPTBL_FREE = round_up_to_pow2(BOOT_CAPTBL_LAST_CAP, CAPMAX_ENTRY_SZ)
 };
 
+/*
+ * The slots of the sinv to user-level fault handlers.
+ */
+typedef enum {
+	FAULT_CAPTBL_MEM_ACCESS        = BOOT_CAPTBL_FLT_MEM_ACCESS,
```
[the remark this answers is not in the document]
phanikishoreg: I feel like `COMP_CAPTBL_FAULT_XXXX` would probably be a better name for these fault capabilities.

I think your point about booter having those fault capabilities is somewhat tricky and perhaps will only make sense to have a subset of fault caps.
If there is a page fault in booter and booter is invoked to resolve it, that literally would be a double fault. So, I'm not sure if it's essential for us to do that. 
But in some projects like fswmanager that Yuxin, Vlad were working on, booter is not just a booter there. (I don't know if that's a long-term idea). In those cases, we'd probably have some scope for booter to have some of the fault capabilities.  In general, if booter is overloaded, then having those fault SINVs in booter would be useful and not otherwise IMO.

src/kernel/include/thd.h
```diff
@@ -18,8 +18,22 @@
...
+ * 1. This data structure is designed to replace the comp_info used in invstk_entry.
+ * The reason for using comp_invstk_info instead of comp_info is we want to set a flag
+ * into the captbl pointer of comp_info to achieve better alignment.
+ * 2. We plan to set a bit into the captbl pointer because captbls are perfectly aligned 
+ * which means the last few bits of captbl is always "0". By this design, we can achieve
+ * both better cache line alignment and structure alignment.
+ * 3. Another advantage of using this structure is if comp_info needs to be changed in
+ * the future, we do not have to change comp_invstk_info.
+ */
+struct comp_invstk_info {
```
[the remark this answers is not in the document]
phanikishoreg: It is a lot clearer.
I'd rephrase "this enables the structure to be 16-byte aligned" to "this keeps the structure 16-byte aligned" just emphasizing that it is already 16byte aligned and will continue to be if we were to make use of a bit in the struct. However, the point in @WenyuanShao 's comment is trying to say that the captbl is already aligned and that there is some "reserved" or "free" bits that can be utilized to enable this mechanism.. I think having that additional sentence would make this completely unambiguous.

src/kernel/include/thd.h
phanikishoreg: I think the way we have those lines right now
```'
case THD_GET_XXXX: *retval = thd->xxxxxx; break;
```
is quite terse. I don't know if macros are useful here.
But if we had some configuration (ex: CONFIG_GP_REGS_ONLY kind ) where we'd enable only a subset to be accessible, then this would both help generate less code and get rid of those other register switch cases. But I don't think we'd have such a config or need such.
Also when we have architecture specific "gets" we'd probably move this "introspect" code to `chal` layer and would be per architecture?? I'm not sure if macros are all that useful. 

I'd probably just do the normal thing here for "consistency".

PR #394, opened 2020-02-27
update the exceptions vector number

huachuan:
Non-maskable Interrupt is missing, option one is add it after the IRQ_DEBUG_TRAP; option two is what I did, manually set the vector numbers

**Add description here.**

### Intent for your PR

Choose one (Mandatory):

### Reviewers (Mandatory):

### Code Quality

### Testing

- [ ] micro_booter
- [ ] unit_pingpong
- [ ] unit_schedtests
- [ ] ...(add others here)


gparmer:
This change obviously takes the system closer to correctness, so it is great. Which is the interrupt vector that you're making not be represented in the enum (presumably offset 2)?

huachuan:
> This change obviously takes the system closer to correctness, so it is great. Which is the interrupt vector that you're making not be represented in the enum (presumably offset 2)?

Vector number 2 Non-maskable Interrupt is not in the enum.

gparmer:
I see. I think that the correct change, then, is

1. to add the NMI handler to the enum (for documentation purposes), and 
2. to prevent this bug in the future by explicitly placing a `= 2`, ` = 3`, etc... after every entry. That way we are explicit, and it is easy to check. 

Would these changes be possible?

Sorry for missing your response!!!

huachuan:
I have added the Non-maskable Interrupt and explicitly placed the exception numbers after every entry.

PR #396, opened 2020-03-07
qemu-mp.sh, qemu.sh: Enable KVM when applicable

msdx321:
Signed-off-by: msdx321 <msdx321@gmail.com>

Enable KVM when KVM acceleration is available.

### Intent for your PR

Choose one (Mandatory):

### Reviewers (Mandatory):

### Code Quality

### Testing

- [ ] micro_booter
- [ ] unit_pingpong
- [ ] unit_schedtests
- [ ] ...(add others here)

### DIFF
--- a/src/platform/i386/qemu.sh
+++ b/src/platform/i386/qemu.sh
@@ -9,7 +9,13 @@ if ! [ -r $1 ]; then
   exit 1
 fi
 
+if [ -e "/dev/kvm" ] && [ -r "/dev/kvm" ] && [ -w "/dev/kvm" ]; then
+  KVM_FLAG="-enable-kvm"
+else 
+  KVM_FLAG=""
+fi
+
 MODULES=$(sh $1 | awk '/^Writing image/ { print $3; }' | tr '\n' ' ')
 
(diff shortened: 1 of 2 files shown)


src/platform/i386/qemu.sh
```diff
@@ -9,7 +9,13 @@ if ! [ -r $1 ]; then
   exit 1
 fi
 
+if [ -e "/dev/kvm" ] && [ -r "/dev/kvm" ] && [ -w "/dev/kvm" ]; then
+  KVM_FLAG="-enable-kvm"
```
phanikishoreg: In other branches, the script used for KVM is explicit: https://github.com/gwsystems/composite/blob/part/src/platform/i386/qemu-kvm.sh

I like this approach, perhaps saves us some learning of different runscripts.
On a similar note, perhaps merge multicore script into single? Composite only runs on as many as NUM_CPU, so having -smp 4 while using only 1 core, might still be OK.

Another nitpick, I think, some logging for KVM enabled/disabled would be useful. If someone new, runs this script on 2 different machines, with completely different performance numbers, they'd be worried without it. Because, likely, the rest of the kernel_tests.sh logs don't indicate why this difference I guess. Just an opinion!

PR #405, opened 2020-07-07
lazy fpu and unit tests

huachuan:
remove ekflib

lazy fpu and unit tests

fix some style issues

**Add description here.**

### Intent for your PR

Choose one (Mandatory):

### Reviewers (Mandatory):

### Code Quality

### Testing

- [ ] micro_booter
- [ ] unit_pingpong
- [ ] unit_schedtests
- [ ] ...(add others here)

### DIFF
--- a/src/kernel/include/fpu.h
+++ b/src/kernel/include/fpu.h
@@ -65,7 +65,7 @@ fpu_init(void)
 		return -1;
 	}
 #endif
-	/*printk("fpu_init on core %d\n", get_cpuid());*/
+
 	return 0;
 }
 
@@ -85,6 +85,7 @@ fpu_is_disabled(void)
 {
 	int *disabled = PERCPU_GET(fpu_disabled);
 	assert(fpu_read_cr0() & FPU_DISABLED_MASK ? *disabled : !*disabled);
+
 	return *disabled;
 }
 
@@ -93,6 +94,7 @@ fpu_read_cr0(void)
 {
 	unsigned long val;
 	asm volatile("mov %%cr0, %0" : "=r"(val));
+
 	return val;
 }
 
@@ -100,10 +102,12 @@ static inline void
 fpu_set(int status)
 {
 	unsigned long val, cr0;
+
 	cr0 = fpu_read_cr0();
 	val = status ? (cr0 & ~FPU_DISABLED_MASK)
 	             : (cr0 | FPU_DISABLED_MASK); // ENABLE(status == 1) : DISABLE(status == 0)
 	asm volatile("mov %0, %%cr0" : : "r"(val));
+
 	return;
 }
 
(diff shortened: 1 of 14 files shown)


src/kernel/include/fpu.h
```diff
@@ -84,41 +151,11 @@ fpu_check_sse(void)
...
-	/* printk("fpu_init on core %d\n", get_cpuid()); */
-
-	return 0;
-}
-
 static inline int
 fpu_disabled_exception_handler(void)
 {
 	struct thread *curr_thd;
+	/*printk("COS KERNEL: SUCCESS fpu enabled, now in exception handler \n");*/
```
gparmer, line 158: Remove comment.

src/kernel/include/fpu.h
```diff
@@ -39,6 +38,74 @@ static inline void          fpu_set(int);
...
+	unsigned long val;
+	asm volatile("mov %%cr0, %0" : "=r"(val));
+	return val;
+}
+
+static inline void
+fpu_set(int status)
+{
+	unsigned long val, cr0;
+	cr0 = fpu_read_cr0();
```
gparmer, line 103: Empty line after local variables.

src/kernel/include/fpu.h
```diff
@@ -39,6 +38,74 @@ static inline void          fpu_set(int);
...
+	val = status ? (cr0 & ~FPU_DISABLED_MASK)
+	             : (cr0 | FPU_DISABLED_MASK); // ENABLE(status == 1) : DISABLE(status == 0)
+	asm volatile("mov %0, %%cr0" : : "r"(val));
+	return;
```
gparmer, line 107: empty line before return.

src/kernel/include/fpu.h
```diff
@@ -39,6 +38,74 @@ static inline void          fpu_set(int);
...
+	fpu_set(FPU_DISABLE);
+	*PERCPU_GET(fpu_disabled) = 1;
+
+	return;
+}
+
+static inline int
+fpu_is_disabled(void)
+{
+	int *disabled = PERCPU_GET(fpu_disabled);
```
gparmer, line 86: See empty line comments in `fpu_set`. Apply them here and in `fpu_read_cr0`.

src/platform/i386/exception.c
```diff
@@ -11,6 +11,7 @@
 void
 print_regs_state(struct pt_regs *regs)
 {
+/*
```
gparmer, line 14: Why remove these?

huachuan: > Why remove these?

I remove this part because I use qemu for the test, when the device not available interruption is triggered, the print out of this part will have bad influence on the result, e.g. thread cannot switch. I will undo this removal.

gparmer: Thanks for the context here. I think that we want it uncommented (as with your updated PR), but you can **remove** the call to print out the registers in the co-processor exception handler. The goal is that 1. the system has to run correctly (so we can't have the  prints in the co-processor exception), and 2. we need to print out the registers in other fault handlers to help debug. So we activate the register print outs in all other exception handlers, but not in yours.

huachuan: Thank you for your comments, I've (1) undone the removal of the print out, (2) removed the calling of print_regs_state() for device not available exception.

PR #412, opened 2021-02-27
PCI library

ldierksheide:
### PCI library

**Port PCI library from eos to composite.**

### Create PCI library within new composite organization, allowing us to use PCI when porting dpdk to composite for MPK work

Choose one (Mandatory):

### Reviewers (Mandatory):
@gparmer @WenyuanShao 

### Code Quality

- [X] I have run the auto formatter on my code before submitting this PR (see doc/auto_formatter.md for instructions)

### Testing

- [ ] micro_booter
- [ ] unit_pingpong
- [ ] unit_schedtests
- [X] tests/pci

### DIFF
--- a/src/components/implementation/tests/pci/doc.md
+++ b/src/components/implementation/tests/pci/doc.md
@@ -1,0 +1,9 @@
+## tests
+
+This is the skeleton interface used by the `mkcomponent.sh` script to aid in the creation of a new component.
+It provides no functionality, and should never be depended on!
+This documentation should be *replaced* in the documentation for a new component.
+
+### Description
+
+### Usage and Assumptions
--- a/src/components/implementation/tests/pci/doc.md
+++ b/src/components/implementation/tests/pci/doc.md
@@ -1,9 +1,10 @@
-## tests
-
-This is the skeleton interface used by the `mkcomponent.sh` script to aid in the creation of a new component.
-It provides no functionality, and should never be depended on!
-This documentation should be *replaced* in the documentation for a new component.
+## tests - pci
 
 ### Description
 
+This component tests the PCI library implemented in `src/components/lib/pci`
+
 ### Usage and Assumptions
+IMPORTANT: We are assuming QEMU version 1:2.11+dfsg-1ubuntu7.36 for these tests and hard-coding the expected values for the devices when running that version. 
+
+If the tests are failing, check again that you are running with this QEMU version, or update `hw_profile.h` with the devices expected for your version.
--- a/src/components/implementation/tests/pci/pci.c
+++ b/src/components/implementation/tests/pci/pci.c
@@ -1,0 +1,159 @@
+#include <cos_component.h>
+#include <llprint.h>
+#include <pci.h>
+
+struct pci_dev all_dev[6] = {
+		{
+			.bus = 0,	
+			.dev = 0,	
+			.func = 0,	
+			.vendor = 0x8086,	
+			.device = 0x1237,	
+			.classcode = 6,	
+			.subclass = 0,	
+			.progIF = 0,	
+			.header = 0,	
+			.data = {},	
+			.index = 0,	
+			.drvdata = 0
+		},
+		{
+			.bus = 0,	
+			.dev = 1,	
+			.func = 0,	
+			.vendor = 0x8086,	
+			.device = 0x7000,	
+			.classcode = 6,	
+			.subclass = 1,	
+			.progIF = 0,	
+			.header = 0x80,	
+			.data = {},	
+			.index = 0,	
+			.drvdata = 0
+		},
+		{
+			.bus = 0,	
+			.dev = 1,	
+			.func = 1,	
+			.vendor = 0x8086,	
+			.device = 0x7010,	
+			.classcode = 1,	
+			.subclass = 1,	
+			.progIF = 0x80,	
+			.header = 0,	
+			.data = {},	
+			.index = 0,	
+			.drvdata = 0
+		},
+		{
+			.bus = 0,	
+			.dev = 1,	
+			.func = 3,	
+			.vendor = 0x8086,	
+			.device = 0x7113,	
+			.classcode = 6,	
+			.subclass = 0x80,	
+			.progIF = 0,	
+			.header = 0,	
+			.data = {},	
+			.index = 0,	
+			.drvdata = 0
+		},
+		{
+			.bus = 0,	
+			.dev = 2,	
+			.func = 0,	
+			.vendor = 0x1234,	
+			.device = 0x1111,	
+			.classcode = 3,	
+			.subclass = 0,	
+			.progIF = 0,	
+			.header = 0,	
+			.data = {},	
+			.index = 0,	
+			.drvdata = 0
+		},
+		{
+			.bus = 0,	
+			.dev = 3,	
+			.func = 0,	
+			.vendor = 0x8086,	
+			.device = 0x100e,	
+			.classcode = 2,	
+			.subclass = 0,	
+			.progIF = 0,	
+			.header = 0,	
+			.data = {},	
+			.index = 0,	
+			.drvdata = 0
+		}
+	};
+
+void
+cos_init(void)
+{
+
+}
+
+int
+main(void)
+{
+	int size;
+	int i;
+	size = pci_num();
+
+	if(size != 6) {
+		printc("Failure: pci_num returned %d instead of 6 (QEMU version 1:2.11+dfsg-1ubuntu7.36)\n", size);
+		return 0;
+	}
+
+	struct pci_dev my_devices[size];
+	
+	if(pci_scan(my_devices, size) != 0) {
+		printc("Failure: pci_scan returned non-0 value\n");
+		return 0;
+	}
+	for(i = 0 ; i < size ; i++) {
+		if(my_devices[i].bus != all_dev[i].bus) {
+			printc("Failure: got %x instead of %x for devices[%d] bus (QEMU version 1:2.11+dfsg-1ubuntu7.36)\n", my_devices[i].bus, all_dev[i].bus, i);
+			return 0;
+		}
+		if(my_devices[i].dev != all_dev[i].dev) {
+			printc("Failure: got %x instead of %x for devices[%d] dev (QEMU version 1:2.11+dfsg-1ubuntu7.36)\n", my_devices[i].dev, all_dev[i].dev, i);
+			return 0;
+
+		}
+		else if(my_devices[i].func != all_dev[i].func) {
+			printc("Failure: got %x instead of %x for devices[%d] func (QEMU version 1:2.11+dfsg-1ubuntu7.36)\n", my_devices[i].func, all_dev[i].func, i);
+			return 0;		
+		}
+		else if(my_devices[i].vendor != all_dev[i].vendor) {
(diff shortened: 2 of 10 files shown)


WenyuanShao (commented):
This is a great PR. But I still have some concerns, most of them are style issues and improvements you can make.

src/components/implementation/tests/pci/doc.md
```diff
@@ -0,0 +1,9 @@
+## tests
+
+This is the skeleton interface used by the `mkcomponent.sh` script to aid in the creation of a new component.
```
gparmer, line 3: Add some documentation here. Doesn't have to be much. But you can at least add a Warning about the hard-coded checking for the devices. Can repeat the qemu version etc... here.

Particularly relevant for "assumptions" below.

src/components/implementation/tests/pci/pci.c
```diff
@@ -0,0 +1,159 @@
+#include <cos_component.h>
+#include <llprint.h>
+#include <pci.h>
+
+struct pci_dev all_dev[6] = {
```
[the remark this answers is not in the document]
gparmer: I agree that a big comment here saying what this is is important.

You can get rid of the `6`  here and let the compiler infer the size.

src/components/implementation/tests/pci/pci.c
```diff
@@ -0,0 +1,159 @@
...
+			.device = 0x100e,	
+			.classcode = 2,	
+			.subclass = 0,	
+			.progIF = 0,	
+			.header = 0,	
+			.data = {},	
+			.index = 0,	
+			.drvdata = 0
+		}
+	};
```
gparmer, line 90: It is a large enough structure, that having it in a separate file might make sense (as @WenyuanShao recommended above). Would give us an opportunity to choose a name to self-document slightly as well. `hw_profile.h` might be appropriate? The .h file would still be in this directory, FYI.

src/components/implementation/tests/pci/pci.c
```diff
@@ -0,0 +1,159 @@
...
+
+void
+cos_init(void)
```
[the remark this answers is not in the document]
gparmer: I think that you can just avoid having a `cos_init`. (you should be able to do that, unless the initialization had a bug creep in.) That might be the best solution?

src/components/implementation/tests/pci/pci.c
```diff
@@ -0,0 +1,159 @@
...
+{
+
+}
+
+int
+main(void)
+{
+	int size;
+	int i;
+	size = pci_num();
```
gparmer, line 103: Either initialize `size` in the definition above, or put a blank line after the definitions and code. The only exception to this (in the style) is assertions. We always think of definitions as "paragraphs".

src/components/implementation/tests/pci/pci.c
```diff
@@ -0,0 +1,159 @@
...
+
+	if(size != 6) {
```
gparmer, line 105: space after `if`. Keywords and control flow operations aren't function calls.

src/components/implementation/tests/pci/pci.c
```diff
@@ -0,0 +1,159 @@
...
+		printc("Failure: pci_num returned %d instead of 6 (QEMU version 1:2.11+dfsg-1ubuntu7.36)\n", size);
+		return 0;
+	}
+
+	struct pci_dev my_devices[size];
```
gparmer, line 110: All definitions at the top of the block (in this case at the top of the function).

src/components/implementation/tests/pci/pci.c
gparmer: I'm not sure I like the name of this function. `num` can mean many different things. Perhaps `pci_dev_count`? Maybe something else that is better?

src/components/implementation/tests/pci/pci.c
gparmer: This is implemented with an `alloca`, thus it effectively dynamically grows the stack based on input. That might result in massive stack overflows in some systems which will be very hard to debug.

I'd prefer to have the size of the array be statically known here since you're already checking that the size is `6` anyway.

src/components/implementation/tests/pci/pci.c
gparmer: This is a recurring issue below. I won't redundantly post this request to lower the # of comments.

src/components/lib/pci/doc.md
```diff
@@ -0,0 +1,9 @@
+## pci
```
gparmer, line 1: Same here. I think that you can add a brief description, and state that an assumption is that we're only using, what, pci 1.0 here? You can add that a todo is to add support for pcie etc...

src/components/lib/pci/io.h
```diff
@@ -0,0 +1,45 @@
+#ifndef IO_H
+#define IO_H
+
+#include <cos_types.h>
+
```
gparmer, line 5: This is all going to break on non-x86. Any preferences on how to handle it?

@WenyuanShao @msdx321 @betahxy

gparmer: Maybe add a way to do conditional compilation of libraries/components into the build system? I don't think that we want this component, nor the test component in non-x86.

src/components/lib/pci/pci.c
```diff
@@ -0,0 +1,131 @@
+/* Simple PCI driver, only works for IVSHMEM device */
```
gparmer, line 1: Comment style is:

```c
/*
 * this is a typical multipline comment
 * that keeps on going.
 */
```
or 
```c
/***
 * for module-level documentation (like in your case).
 */
```

a double star at the start is for function-level documentation (e.g. `/** ... */`).

src/components/lib/pci/pci.c
```diff
@@ -0,0 +1,131 @@
...
+/* PCI background information can be obtained from OSDev http://wiki.osdev.org/PCI */
+/* IVSHMEM device information can be found from qemu documentation */
```
gparmer, line 3: I don't know what this comment is referring to. Up to you if that is OK or not.

src/components/lib/pci/pci.c
```diff
@@ -0,0 +1,131 @@
...
+
+#include <consts.h>
+#include <assert.h>
+#include "io.h"
+#include "../../../kernel/include/chal.h"
```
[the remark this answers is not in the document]
gparmer: Not strange: incorrect.

We *cannot* include kernel files, in general, in user. This needs to change, so @ldierksheide we can discuss the bugs when you remove this, and how to address them.

src/components/lib/pci/pci.c
```diff
@@ -0,0 +1,131 @@
...
+
+	if(sz <= 0) return -1;
+
+	for(i = 0 ; i < PCI_BUS_MAX ; i++) {
+		for(j = 0 ; j < PCI_DEVICE_MAX ; j++) {
+			for(f = 0 ; f < PCI_FUNC_MAX ; f++) {
+				reg = pci_config_read(i, j, f, 0x0);
+				if (reg == PCI_BITMASK_32) continue;
+				for(k = 0 ; k < PCI_DATA_NUM ; k++) devices[dev_num].data[k] = pci_config_read(i, j, f, k << 2);
+				devices[dev_num].bus       = (u32_t)i;
```
gparmer, line 62: Might be less busy to use the static initialization syntax here?

```c
devices[dev_num] = (struct pci_dev) {
        .bus = (u32_t)i,
        .dev = (u32_t)j,
        ...
};
```

src/components/lib/pci/pci.c
```diff
@@ -0,0 +1,131 @@
...
+				devices[dev_num].func      = (u32_t)f;
+				devices[dev_num].vendor    = (u16_t)PCI_VENDOR_ID(devices[dev_num].data[0]);
+				devices[dev_num].device    = (u16_t)PCI_DEVICE_ID(devices[dev_num].data[0]);
+				devices[dev_num].classcode = (u8_t)PCI_CLASS_ID(devices[dev_num].data[2]);
+				devices[dev_num].subclass  = (u8_t)PCI_SUBCLASS_ID(devices[dev_num].data[2]);
+				devices[dev_num].progIF    = (u8_t)PCI_PROG_IF(devices[dev_num].data[2]);
+				devices[dev_num].header    = (u8_t)PCI_HEADER(devices[dev_num].data[3]);
+				for(k = 0 ; k < PCI_BAR_NUM ; k++) {
+					devices[dev_num].bar[k].raw = devices[dev_num].data[4+k];
+					/* LED: add paddr */
```
gparmer, line 73: We use LED and the like to identify TODOs, usually. As you aren't doing that here, and are just providing a comment, I don't htink you need it.

However, you likely want to provide a more informative answer here. Something like "The least significant 8 bits hold X data, which we need to mask out to get the physical address."

src/components/lib/pci/pci.c
```diff
@@ -0,0 +1,131 @@
...
+					devices[dev_num].bar[k].paddr = devices[dev_num].bar[j].raw & 0xFFFFFFF0;
+
+				}
+				
+				dev_num++;
+				if(dev_num >= sz || dev_num >= PCI_DEVICE_MAX) {
+					
+                    return 0;
```
gparmer, line 81: The spacing is wonky here. Likely a mismatch between how github renders and your editor? I'm guessing your editor is hiding a tab vs. space mismatch.

src/components/lib/pci/pci.c
```diff
@@ -0,0 +1,131 @@
...
+	struct pci_bar *bar;
+
+	dev_num = 0;
+
+	for(i = 0 ; i < PCI_BUS_MAX ; i++) {
+		for(j = 0 ; j < PCI_DEVICE_MAX ; j++) {
+			for(f = 0 ; f < PCI_FUNC_MAX ; f++) {
+				reg = pci_config_read(i, j, f, 0x0);
+				if (reg == PCI_BITMASK_32) continue;
+					dev_num++;
```
gparmer, line 123: irregular indentation.

src/components/lib/pci/pci.h
```diff
@@ -0,0 +1,78 @@
...
+	u8_t  subclass;
+	u8_t  progIF;
+	u8_t  header;
+	struct pci_bar bar[PCI_BAR_NUM];
+	u32_t data[PCI_DATA_NUM];
+	unsigned int index;
+	void *drvdata;
+} __attribute__((packed));
+
+/* API */
```
gparmer, line 71: Worth providing full documentation for all of this. See the documentation section of the dev_manual. The documentation scripts should pull out the `/**...*/` comments ahead of functions. This is your main documentation for the API.

I also don't think that we need to have all of the struct detail (above) in this file. All of that is fine in the .c file as it is "implementation guts".

gparmer (changes requested):
Looks really good! Only minor changes requested. Feel free to ask about any of them.

src/components/lib/pci/pci.c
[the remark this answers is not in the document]
ldierksheide: That must have been a relic from the previous eos code. But removing that include doesn't seem to mess anything up

src/components/lib/pci/pci.h
ldierksheide: I don't think I'm able to remove the structs from here, since setting up the hardcoded pci_dev array includes pci.h and requires knowledge of the data structures. unless there's a smarter way to be able to have both? I could try moving them into a different .h and just include that

gparmer (commented):
I'm happy with this. @WenyuanShao ?

WenyuanShao:
Looks good! I have no other concern @gparmer.

PR #417, opened 2021-03-18
sqlite: Initial port 

msdx321:

**Add description here.**

### Intent for your PR

Choose one (Mandatory):

### Reviewers (Mandatory):

### Code Quality

### Testing

- [ ] micro_booter
- [ ] unit_pingpong
- [ ] unit_schedtests
- [ ] ...(add others here)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "src/components/lib/ck/ck"]
 	path = src/components/lib/ck/ck
 	url = https://github.com/gwsystems/ck.git
+[submodule "src/components/lib/sqlite/sqlite"]
+	path = src/components/lib/sqlite/sqlite
+	url = https://github.com/msdx321/sqlite.git
--- a/src/components/implementation/tests/sqlite_tests/doc.md
+++ b/src/components/implementation/tests/sqlite_tests/doc.md
@@ -1,0 +1,9 @@
+## tests
+
+This is the skeleton interface used by the `mkcomponent.sh` script to aid in the creation of a new component.
+It provides no functionality, and should never be depended on!
+This documentation should be *replaced* in the documentation for a new component.
+
+### Description
+
+### Usage and Assumptions
--- a/src/components/implementation/tests/sqlite_tests/sqlite_tests.c
+++ b/src/components/implementation/tests/sqlite_tests/sqlite_tests.c
@@ -1,0 +1,44 @@
+#include <cos_component.h>
+#include <llprint.h>
+#include <sqlite3.h>
+#include <cos_defkernel_api.h>
+
+void
+cos_init(void)
+{
+}
+
+int result_print(void *v, int argc, char **argv, 
+                    char **col_name) {
+    for (int i = 0; i < argc; i++) {
+        printc("%s = %s\n", col_name[i], argv[i] ? argv[i] : "NULL");
+    }
+    
+    return 0;
+}
+
+int
+main(void)
+{
+	printc("Calling sqlite functions\n");
+
+    sqlite3 *db;
+    char *err_msg = 0;
+    char *sql = "CREATE TABLE TestTable(Id INTEGER PRIMARY KEY, Value TEXT);"
+    "INSERT INTO TestTable(Value) VALUES ('TestItem1');"
+    "INSERT INTO TestTable(Value) VALUES ('TestItem2');"
+    "INSERT INTO TestTable(Value) VALUES ('TestItem3');"
+    "INSERT INTO TestTable(Value) VALUES ('TestItem4');"
+    "INSERT INTO TestTable(Value) VALUES ('TestItem5');";
+
+    sqlite3_open(":memory:", &db);
+
+    sqlite3_exec(db, sql, 0, 0, &err_msg);
+    
+	sql = "SELECT * FROM TestTable WHERE Id == 3";
+	sqlite3_exec(db, sql, result_print, 0, &err_msg);
+
+    sqlite3_close(db);
+
+	while(1);
+}
--- a/src/components/lib/libc/Makefile
+++ b/src/components/lib/libc/Makefile
@@ -7,7 +7,7 @@
 #
 # NOTE: libc is special-cased and the compilation is orchestrated
 # directly. It should have no dependencies.
-LIBRARY_OUTPUT = c crypt m thread resolv rt
+LIBRARY_OUTPUT = c crypt m pthread resolv rt
 # The .o files that are mandatorily linked into dependents. This is
 # rarely used, and only when normal .a linking rules will avoid
 # linking some necessary objects. This list is of names (for example,
--- a/src/components/lib/posix/posix.c
+++ b/src/components/lib/posix/posix.c
@@ -12,9 +12,12 @@
 #include <cos_component.h>
 #include <cos_defkernel_api.h>
 #include <llprint.h>
+#include <memmgr.h>
+/*
 #include <sl.h>
 #include <sl_lock.h>
 #include <sl_thd.h>
+*/
 
 volatile int* null_ptr = NULL;
 #define ABORT() do {int i = *null_ptr;} while(0)
@@ -31,7 +34,7 @@ libc_syscall_override(cos_syscall_t fn, int syscall_num)
 	cos_syscalls[syscall_num] = fn;
 }
 
-struct sl_lock stdout_lock = SL_LOCK_STATIC_INIT();
+//struct sl_lock stdout_lock = SL_LOCK_STATIC_INIT();
 
 ssize_t
 write_bytes_to_stdout(const char *buf, size_t count)
@@ -46,9 +49,9 @@ cos_write(int fd, const void *buf, size_t count)
 {
 	/* You shouldn't write to stdin anyway, so don't bother special casing it */
 	if (fd == 1 || fd == 2) {
-		sl_lock_take(&stdout_lock);
+		//sl_lock_take(&stdout_lock);
 		write_bytes_to_stdout((const char *) buf, count);
-		sl_lock_release(&stdout_lock);
+		//sl_lock_release(&stdout_lock);
 		return count;
 	} else {
 		printc("fd: %d not supported!\n", fd);
@@ -60,13 +63,13 @@ ssize_t
 cos_writev(int fd, const struct iovec *iov, int iovcnt)
 {
 	if (fd == 1 || fd == 2) {
-		sl_lock_take(&stdout_lock);
+		//sl_lock_take(&stdout_lock);
 		int i;
 		ssize_t ret = 0;
 		for(i=0; i<iovcnt; i++) {
 			ret += write_bytes_to_stdout((const void *)iov[i].iov_base, iov[i].iov_len);
 		}
-		sl_lock_release(&stdout_lock);
+		//sl_lock_release(&stdout_lock);
 		return ret;
 	} else {
 		printc("fd: %d not supported!\n", fd);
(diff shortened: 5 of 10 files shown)


src/components/lib/posix/posix.c
```diff
@@ -31,7 +34,7 @@ libc_syscall_override(cos_syscall_t fn, int syscall_num)
 	cos_syscalls[syscall_num] = fn;
 }
 
-struct sl_lock stdout_lock = SL_LOCK_STATIC_INIT();
+//struct sl_lock stdout_lock = SL_LOCK_STATIC_INIT();
```
Others, line 37: Is there demand for a single threaded version of this? Removing all this synchronization logic is going to lead to weird bugs as soon as you have more than one thread. Why doesn't it work as-is? Too tight coupling to SL?

gparmer: Great question, and good intuition. We revamping libposix right now to use a generic interface that can either be co-located (via library) with sl, *or* accessed in another component. Given this, most of the synchronization facilities (and sl dependencies) in posix need to change backing APIs.

As such, Bite is trying to get sqlite working for a comparison case, and is not trying to solve other problems for now. As we update posix, I'm hoping we can add back in those features.

A lot of the cruft assuming single-component runtimes has been removed, but we have more to do, and libposix is one of the big places more love needs to be applied.

.gitmodules
```diff
@@ -4,3 +4,6 @@
 [submodule "src/components/lib/ck/ck"]
 	path = src/components/lib/ck/ck
 	url = https://github.com/gwsystems/ck.git
+[submodule "src/components/lib/sqlite/sqlite"]
```
gparmer, line 7: You'll want to move this into gwsystems. Remember, we want the main branch of the repo be unmodifed, and have a `cos` branch that we pull in as a submodule. That way we have a path for updating the code as upstream changes.

src/components/implementation/tests/sqlite_tests/doc.md
```diff
@@ -0,0 +1,9 @@
+## tests
+
+This is the skeleton interface used by the `mkcomponent.sh` script to aid in the creation of a new component.
```
gparmer, line 3: Add documentation here, or remove the file.

src/components/implementation/tests/sqlite_tests/sqlite_tests.c
```diff
@@ -0,0 +1,44 @@
...
+#include <llprint.h>
+#include <sqlite3.h>
+#include <cos_defkernel_api.h>
+
+void
+cos_init(void)
+{
+}
+
+int result_print(void *v, int argc, char **argv,
```
gparmer, line 11: your line width wraparound seems too aggressive. I don't think that the argument needs to be on another line.

src/components/implementation/tests/sqlite_tests/sqlite_tests.c
```diff
@@ -0,0 +1,44 @@
+#include <cos_component.h>
+#include <llprint.h>
+#include <sqlite3.h>
+#include <cos_defkernel_api.h>
+
+void
+cos_init(void)
```
gparmer, line 7: If you remove this function, it should still work. If you don't have code in it, remove it!

src/components/implementation/tests/sqlite_tests/sqlite_tests.c
```diff
@@ -0,0 +1,44 @@
...
+        printc("%s = %s\n", col_name[i], argv[i] ? argv[i] : "NULL");
+    }
+    
+    return 0;
+}
+
+int
+main(void)
+{
+	printc("Calling sqlite functions\n");
```
gparmer, line 23: inconsistent indentation styules. likely mixing spaces and tabs.

gparmer: variable definitions should be at the top of the block.

src/components/implementation/tests/sqlite_tests/sqlite_tests.c
```diff
@@ -0,0 +1,44 @@
...
+
+    sqlite3 *db;
+    char *err_msg = 0;
+    char *sql = "CREATE TABLE TestTable(Id INTEGER PRIMARY KEY, Value TEXT);"
+    "INSERT INTO TestTable(Value) VALUES ('TestItem1');"
```
gparmer, line 28: Personal opinion (ignore if you'd like): all strings would look better if the starting "s were aligned.

src/components/implementation/tests/sqlite_tests/sqlite_tests.c
```diff
@@ -0,0 +1,44 @@
...
+    "INSERT INTO TestTable(Value) VALUES ('TestItem2');"
+    "INSERT INTO TestTable(Value) VALUES ('TestItem3');"
+    "INSERT INTO TestTable(Value) VALUES ('TestItem4');"
+    "INSERT INTO TestTable(Value) VALUES ('TestItem5');";
+
+    sqlite3_open(":memory:", &db);
+
+    sqlite3_exec(db, sql, 0, 0, &err_msg);
+    
+	sql = "SELECT * FROM TestTable WHERE Id == 3";
```
gparmer, line 38: indentation again.

src/components/implementation/tests/sqlite_tests/sqlite_tests.c
```diff
@@ -0,0 +1,44 @@
...
+	sqlite3_exec(db, sql, result_print, 0, &err_msg);
+
+    sqlite3_close(db);
+
+	while(1);
```
gparmer, line 43: space after keywords

src/components/lib/libc/Makefile
```diff
@@ -7,7 +7,7 @@
 #
 # NOTE: libc is special-cased and the compilation is orchestrated
 # directly. It should have no dependencies.
-LIBRARY_OUTPUT = c crypt m thread resolv rt
+LIBRARY_OUTPUT = c crypt m pthread resolv rt
```
gparmer, line 10: Thanks for this!!!

src/components/lib/posix/posix.c
```diff
@@ -12,9 +12,12 @@
 #include <cos_component.h>
 #include <cos_defkernel_api.h>
 #include <llprint.h>
+#include <memmgr.h>
+/*
```
gparmer, line 16: No commented out code in commits. That's why we have commit histories.

gparmer: This comment applies below as well.

src/components/lib/posix/posix.c
```diff
@@ -46,9 +49,9 @@ cos_write(int fd, const void *buf, size_t count)
 {
 	/* You shouldn't write to stdin anyway, so don't bother special casing it */
 	if (fd == 1 || fd == 2) {
-		sl_lock_take(&stdout_lock);
+		//sl_lock_take(&stdout_lock);
```
gparmer, line 52: I'm not sure that I want to remove all of these. Certainly we want to get rid of sl, but I don't think that we want to get rid of the code for lock take/release. We want to add it back in, so it is odd ot remove it all. How about this:

you define some #defines for lock_take lock_release and other similar functions. The #defines can be empty for now (thus not generating any executable logic).

src/components/lib/posix/posix.c
```diff
@@ -170,7 +174,8 @@ cos_mprotect(void *addr, size_t len, int prot)
 pid_t
 cos_gettid(void)
 {
-	return (pid_t) sl_thdid();
+	//return (pid_t) sl_thdid();
```
gparmer, line 177: Replace with `cos_thdid()`

src/components/lib/posix/posix.c
```diff
@@ -221,16 +226,20 @@ cos_nanosleep(const struct timespec *req, struct timespec *rem)
 		return -1;
 	}
 
-	wakeup_deadline        = sl_now() + sl_usec2cyc(time_to_microsec(req));
-	completed_successfully = sl_thd_block_timeout(0, wakeup_deadline);
-	wakeup_time   = sl_now();
+	//wakeup_deadline        = sl_now() + sl_usec2cyc(time_to_microsec(req));
+	wakeup_deadline = 0;
```
gparmer, line 230: I don't think that silently failing is the best policy here. We should probably fail loudly if a function that we don't support is called. BUG(), assert(0), after a printc to explain. etc...

I don't trust client checking for `ENOTSUP` which would be the normal solution. Failure is more appropriate if it isn't supported here.

src/components/lib/posix/posix.c
```diff
@@ -263,23 +272,27 @@ cos_set_tid_address(int *tidptr)
  * };
  */
 
-void* backing_data[SL_MAX_NUM_THDS];
+//void* backing_data[SL_MAX_NUM_THDS];
+void* backing_data[1];
```
gparmer, line 276: `COS_NUM_THREADS`, I think???

src/components/lib/posix/posix.c
```diff
@@ -291,11 +304,13 @@ cos_clone(int (*func)(void *), void *stack, int flags, void *arg, pid_t *ptid, v
 		return -1;
 	}
 
-	struct sl_thd * thd = sl_thd_alloc((cos_thd_fn_t) func, arg);
+	//struct sl_thd * thd = sl_thd_alloc((cos_thd_fn_t) func, arg);
+	struct sl_thd * thd = 0;
```
gparmer, line 308: This shouldn't compile, right? Where is the defn of `sl_thd` coming from here?

src/components/lib/posix/posix.c
```diff
@@ -331,124 +346,33 @@ struct futex_data futexes[FUTEX_COUNT];
...
 
-struct sl_lock futex_lock = SL_LOCK_STATIC_INIT();
+//struct sl_lock futex_lock = SL_LOCK_STATIC_INIT();
 
 /*
  * precondition: futex_lock is taken
  */
 int
 cos_futex_wait(struct futex_data *futex, int *uaddr, int val, const struct timespec *timeout)
 {
```
gparmer, line 361: This is all pretty challenging logic that Gregor spent a lot of time on. What I'd like: create an issue with a link to the commit in which this is removed, so that we can know how to easily bring back this code when hook up posix to the rest of the system.

src/components/lib/sqlite/doc.md
```diff
@@ -0,0 +1,9 @@
+## sqlite
+
+This is the skeleton library used by the `mklib.sh` script to aid in the creation of a new library.
```
gparmer, line 3: documentation or remove.

gparmer (changes requested):
I think that a lot of your changes are in the sqlite repo. How do you want to proceed with that. See my comment about having a main branch that tracks sqlite's upstream.

Some comments about how to avoid losing all of the sync and futex code.

src/components/lib/posix/posix.c
Others: I think there have to be ways to do this that don't sacrifice correctness. (ie that fault instead of being unsafe concurrently) I'm always nervous leaving things in a subtly broken state--it's easy to forget the exact brokenness and put yourself in debugging hell for no reason later

gparmer: Great point.

Perhaps we replace the current lock implementation with something that simply errors out on contention:

```c
struct lock {
        unsigned long owner;
};
struct lock global_lock = (struct lock) { .owner = 0 };
void
lock_take(void) {
        // failure here indicates that you're using concurrency/parallelism, which posix doesn't yet support!
        if (!ps_cas(&global_lock.owner, 0, cos_thdid())) assert(0);
}
void
lock_release(void) {
        // failure here indicates unpaired take and release.
        assert(global_lock.owner == cos_thdid());
        ps_store(&global_lock.owner, 0);
}
```
Replace normal lock behavior which requires contention management with simple "error out".

PR #419, opened 2021-05-13
Initial Checkpoint 

ldierksheide:

**Checkpoint creation and execution**

### Intent for your PR

Choose one (Mandatory):

### Reviewers (Mandatory):
@gparmer @WenyuanShao @msdx321 @betahxy

### Code Quality

### Testing

- [ ] micro_booter
- [x] unit_pingpong
- [x] unit_schedtests
- [x] chkpt


src/components/implementation/no_interface/llbooter/llbooter.c
```diff
@@ -299,6 +320,64 @@ comps_init(void)
...
+#ifdef ENABLE_CHKPT
+	/* create the component */
+	void *elf_hdr;
+	int   keylen;
+	compid_t id = crt_ncomp() + 1;
+	const char *root = "binaries/";
+	int   len  = strlen(root);
+	char  path[INITARGS_MAX_PATHNAME];
+
+	assert(id < MAX_NUM_COMPS && id > 0 && name);
```
msdx321, line 335: Personally I would prefer do an assert right after the assign. For example
``` c
compid_t id;
char  path[INITARGS_MAX_PATHNAME];

id = crt_ncomp() + 1;
assert(id < MAX_NUM_COMPS && id > 0 && name);

memset(path, 0, INITARGS_MAX_PATHNAME);
strncat(path, root, len);
assert(path[len] == '\0');
```

src/components/implementation/no_interface/llbooter/llbooter.c
```diff
@@ -34,13 +34,23 @@
...
 static const  compid_t sched_root_id  = 2;
 static        long     boot_id_offset = -1;
 
-SS_STATIC_SLAB(sinv, struct crt_sinv, BOOTER_MAX_SINV);
-SS_STATIC_SLAB(thd,  struct crt_thd,  BOOTER_MAX_INITTHD);
-SS_STATIC_SLAB(rcv,  struct crt_rcv,  BOOTER_MAX_SCHED);
+SS_STATIC_SLAB(sinv, 	struct crt_sinv, 	BOOTER_MAX_SINV);
+SS_STATIC_SLAB(thd,  	struct crt_thd,  	BOOTER_MAX_INITTHD);
+SS_STATIC_SLAB(rcv,  	struct crt_rcv,  	BOOTER_MAX_SCHED);
+SS_STATIC_SLAB(chkpt, 	struct crt_chkpt, 	BOOTER_MAX_CHKPT);
```
gparmer, line 53: tabs vs. spaces here? I suggest spaces if you want everything to line up.

src/components/implementation/no_interface/llbooter/llbooter.c
```diff
@@ -270,7 +282,16 @@ comps_init(void)
 		ss_sinv_activate(sinv);
 		printc("\t%s (%lu->%lu):\tclient_fn @ 0x%lx, client_ucap @ 0x%lx, server_fn @ 0x%lx\n",
 		       sinv->name, sinv->client->id, sinv->server->id, sinv->c_fn_addr, sinv->c_ucap_addr, sinv->s_fn_addr);
+	#ifdef ENABLE_CHKPT
+		serv->sinvs[serv->n_sinvs] = sinv;
```
gparmer, line 286: checks that `serv->n_sinvs` is within bounds? can just be asserts.

src/components/implementation/no_interface/llbooter/llbooter.c
```diff
@@ -270,7 +282,16 @@ comps_init(void)
...
+		serv->n_sinvs++;
+
+		cli->sinvs[cli->n_sinvs] = sinv;
```
gparmer, line 289: same

src/components/implementation/no_interface/llbooter/llbooter.c
```diff
@@ -270,7 +282,16 @@ comps_init(void)
...
+		cli->n_sinvs++;
+	#endif /* ENABLE_CHKPT */
 	}
+	
+	args_iter(&comps, &i, &curr);
```
gparmer, line 294: It seems quite odd to have to add this. Are you sure this is necessary? I cannot see above or below why you'd need this.

src/components/implementation/no_interface/llbooter/llbooter.c
```diff
@@ -299,6 +320,64 @@ comps_init(void)
...
 }
 
+void 
+chkpt_comp_init(struct crt_comp *comp, struct crt_chkpt *chkpt, char *name)
+{
+#ifdef ENABLE_CHKPT
+	/* create the component */
+	void *elf_hdr;
+	int   keylen;
+	compid_t id = crt_ncomp() + 1;
```
gparmer, line 330: comment:

`/* We only support a single checkpoint directly above the existing components */`

src/components/implementation/no_interface/llbooter/llbooter.c
gparmer: Yeah, this looks cleaner to me as well.

src/components/implementation/no_interface/llbooter/llbooter.c
```diff
@@ -299,6 +320,64 @@ comps_init(void)
...
+	const char *root = "binaries/";
+	int   len  = strlen(root);
+	char  path[INITARGS_MAX_PATHNAME];
+
+	assert(id < MAX_NUM_COMPS && id > 0 && name);
+	memset(path, 0, INITARGS_MAX_PATHNAME);
+	strncat(path, root, len);
```
gparmer, line 337: before this: `assert(len < INITARGS_MAX_PATHNAME)`? I believe this is productive. Could be wrong and it mioght be redundant with some check somewhere.

src/components/implementation/no_interface/llbooter/llbooter.c
```diff
@@ -299,6 +320,64 @@ comps_init(void)
...
+		if (crt_comp_create_from(comp, name, id, chkpt)) {
+			printc("Error constructing the resource tables and image of component %s.\n", comp->name);
+			BUG();
+		}
+	}
+	assert(comp->refcnt != 0);
+
+	/* TODO: create the thread/execution context */
+	struct crt_comp_exec_context ctxt = { 0 };
+	/* TODO: assume only chkpt components that can be executed, but aren't schedulers */
```
gparmer, line 355: I don't really parse this comment. Can it be reworded? Am I just being obtuse?

src/components/implementation/no_interface/llbooter/llbooter.c
```diff
@@ -348,14 +427,50 @@ void
 init_done(int parallel_init, init_main_t main_type)
 {
 	compid_t client = (compid_t)cos_inv_token();
-	struct crt_comp *c;
+	struct crt_comp    *c;
+	struct crt_chkpt   *chkpt;
+	struct crt_comp    *new_comp  = boot_comp_get(crt_ncomp() + 1);
+	thdcap_t 		thdcap;
```
gparmer, line 433: tabs are throwing off alignment here, and the next 2 lines.

src/components/implementation/no_interface/llbooter/llbooter.c
```diff
@@ -348,14 +427,50 @@ void
...
+	int 			ret;
+	char			*name;
 
 	assert(client > 0 && client <= MAX_NUM_COMPS);
 	c = boot_comp_get(client);
 
 	crt_compinit_done(c, parallel_init, main_type);
 
+#ifdef ENABLE_CHKPT
+	if (c->id == cos_compid()) {
```
gparmer, line 443: Can we throw all of this `ENABLE_CHKPT` block in a separate function? It isn't great for it to make the main path more complex, and it would be good to have most of the checkpoint logic together.

src/components/implementation/no_interface/llbooter/llbooter.c
```diff
@@ -348,14 +427,50 @@ void
...
+	 	/* don't allow chkpnts of the booter */
+	 	BUG();
+	}
+
+	name = "chkpt_";
+	strncat(name, c->name, strlen(c->name));
```
gparmer, line 449: Isn't this a bug? name doesn't have any space allocated after the static string for `c->name`. So you're just writing into RO memory indiscriminately.

When you do properly allocate `name` (statically in the function is fine), you'll want to add an assertion that `strlen(c->name) + strlen("chkpt_") < whatever the size of name is`?

src/components/implementation/no_interface/llbooter/llbooter.c
```diff
@@ -348,14 +427,50 @@ void
...
+	/* completed all initialization */
+	if (c->init_state >= CRT_COMP_INIT_MAIN) {
+		if (crt_nchkpt() > BOOTER_MAX_CHKPT) {
+			BUG();
+		}
+		chkpt = ss_chkpt_alloc();
+		if (crt_chkpt_create(chkpt, c) != 0) {
+			BUG();
+		}
+		ss_chkpt_activate(chkpt);
```
gparmer, line 459: no return value for this? If not, that's fine.

src/components/lib/crt/crt.c
```diff
@@ -55,6 +55,54 @@
 
 #define CRT_REFCNT_INITVAL 1
 
+static int nchkpt = 0;
+static int ncomp = 1;
+
+int crt_ncomp()
```
gparmer, line 61:
```c
int
crt_ncomp(void) {
...
}
```

I know, pedantic. Sorry!

src/components/lib/crt/crt.c
```diff
@@ -55,6 +55,54 @@
...
+{
+	return ncomp;
+}
+
+int crt_nchkpt()
```
gparmer, line 66: same structuring

src/components/lib/crt/crt.c
```diff
@@ -140,6 +188,8 @@ crt_comp_init(struct crt_comp *c, char *name, compid_t id, void *elf_hdr, vaddr_
 	assert(!elf_hdr || c->entry_addr != 0); /* same as `if elf_hdr then c->entry_addr` */
 	simple_barrier_init(&c->barrier, init_parallelism());
 
+	ncomp++;
```
gparmer, line 191: This should be atomic. I believe  `ps_faa(ncomp, 1)`.

src/components/lib/crt/crt.c
```diff
@@ -167,6 +217,74 @@ crt_comp_create_with(struct crt_comp *c, char *name, compid_t id, struct crt_com
...
 			  r->ptc, r->ctc, r->compc, r->heap_ptr, r->captbl_frontier,
 			  cos_compinfo_get(cos_defcompinfo_curr_get()));
+	return 0;
+}
+
+/** 
+ * Create the component from the checkpoint
+ * c is the new component
+ * chkpt is the checkpoint used to create c
+ * name is c's name
```
gparmer, line 227: This comment doesn't follow the structure of this type of comment. See `crt_comp_create` below. Just copy and paste that comment and change the sections. Use that style of argument documentation, use of empty lines, etc...

src/components/lib/crt/crt.c
```diff
@@ -167,6 +217,74 @@ crt_comp_create_with(struct crt_comp *c, char *name, compid_t id, struct crt_com
...
+	int     ret;
+	vaddr_t	info = chkpt->c->info;
+
+	assert(c && name);
+
+	if (crt_comp_init(c, name, id, NULL, chkpt->c->info)) BUG();
+	ci      = cos_compinfo_get(c->comp_res);
+	root_ci = cos_compinfo_get(cos_defcompinfo_curr_get());
+
+	// overwrite
```
gparmer, line 247: Use `/* */` comments

src/components/lib/crt/crt.c
```diff
@@ -167,6 +217,74 @@ crt_comp_create_with(struct crt_comp *c, char *name, compid_t id, struct crt_com
...
+	c->entry_addr = chkpt->c->entry_addr;
+	c->ro_addr = chkpt->c->ro_addr;
+	c->rw_addr = chkpt->c->rw_addr;
+
+	//re-work the sinvs with the right IDs
```
gparmer, line 252: same

src/components/lib/crt/crt.c
```diff
@@ -167,6 +217,74 @@ crt_comp_create_with(struct crt_comp *c, char *name, compid_t id, struct crt_com
...
+	memcpy(c->sinvs, chkpt->c->sinvs, sizeof(c->sinvs));
+	c->n_sinvs = chkpt->c->n_sinvs;
+	for(u32_t i = 0; i < c->n_sinvs; i++) {
+		//if the client was the old component, update id in new
```
gparmer, line 256: comment styule

src/components/lib/crt/crt.c
```diff
@@ -167,6 +217,74 @@ crt_comp_create_with(struct crt_comp *c, char *name, compid_t id, struct crt_com
...
+		if(chkpt->c->sinvs[i]->client->id == chkpt->c->id) {
+			c->sinvs[i]->client = c;
```
gparmer, line 258: sinvs from a component should *only have* `->client->id == chkpt->c->id`. And none of the servers should have its id. If all of this cannot be replaced with the following, there might be something very strange going on:

```c
struct crt_sinv inv = &chkpt->c->sinvs[i];

assert(inv->client->id == chkpt->c->id);
c->sinvs[i]->client = c;
assert(inv->server->id != chkpt->c->id);
```
If I'm wrong, then either 1. the sinv saving code is doing something strange, or 2. I'm missing something here.

src/components/lib/crt/crt.c
```diff
@@ -263,7 +386,9 @@ crt_booter_create(struct crt_comp *c, char *name, compid_t id, vaddr_t info)
 		.init_state = CRT_COMP_INIT_PREINIT,
 		.main_type  = INIT_MAIN_NONE,
 		.init_core  = cos_cpuid(),
-		.barrier    = SIMPLE_BARRIER_INITVAL
+		.barrier    = SIMPLE_BARRIER_INITVAL,
+		
+		.n_sinvs	= 0,
```
gparmer, line 391: tab throwing off spacing?

No reason to havea n empty line before this???

src/components/lib/crt/crt.c
```diff
@@ -167,6 +217,74 @@ crt_comp_create_with(struct crt_comp *c, char *name, compid_t id, struct crt_com
...
+	c->n_sinvs = chkpt->c->n_sinvs;
+	for(u32_t i = 0; i < c->n_sinvs; i++) {
+		//if the client was the old component, update id in new
+		if(chkpt->c->sinvs[i]->client->id == chkpt->c->id) {
+			c->sinvs[i]->client = c;
+		}
+		if(chkpt->c->sinvs[i]->server->id == chkpt->c->id) {
+			c->sinvs[i]->server = c;
+		}
+	}
```
gparmer, line 263: Isn't all of this creating pointers to the same set of component sinvs? Since the sinvs array is an array of pointers, aren't all checkpoints of the same component going to point to the same actual `crt_sinv`s? Thus any updates we make to the ids will be overwritten by each checkpoint?

gparmer: If this is true, the alternative is to just have the sinvs array in each chkpoint/component be of actual `crt_sinv`s rather than pointers.

src/components/lib/crt/crt.h
```diff
@@ -69,6 +73,12 @@ struct crt_comp {
 	coreid_t init_core;
 
 	crt_refcnt_t refcnt;
+
+	size_t tot_sz_mem;
+	size_t ro_sz;
+	struct crt_sinv *sinvs[CRT_COMP_SINVS_LEN];
```
gparmer, line 79: see comment above. Perhaps this shouldn't be pointers.

src/components/lib/crt/crt.h
```diff
@@ -80,6 +90,12 @@ struct crt_comp_resources {
 	vaddr_t     info;
 };
 
+struct crt_chkpt {
+    struct crt_comp *c;
```
gparmer, line 94: you're mixing spaces and tabs here. See the mismatch on github to make it obvious.

src/components/lib/crt/crt.h
```diff
@@ -80,6 +90,12 @@ struct crt_comp_resources {
...
+	char			*mem;
+	size_t			 tot_sz_mem;
```
gparmer, line 96: I don't think that `mem` etc need to be so offset here.

gparmer (changes requested):
Looks great! Only a couple of easy to fix (potential) bugs, and mainly style updates.

src/components/implementation/no_interface/llbooter/llbooter.c
ldierksheide: I removed the first one because agreed that it doesn't make sense. 

I will re-phrase the second one -- it's left from a discussion about not checkpointing schedulers/capability managers, so we aren't copying the capability tables (but yeah, the way it's phrased with that code doesn't make sense)

src/components/lib/crt/crt.c
ldierksheide: ok, the server part is never getting called in the test I have, so I will replace with the above

src/components/lib/crt/crt.h
ldierksheide: ok, makes sense. I moved this to above the definition for crt_comp because of the issue with circular definition and I think that works

gparmer: Does that mean that it is no longer allocated within the component?

src/components/implementation/no_interface/llbooter/llbooter.c
```diff
@@ -344,18 +429,64 @@ execute(void)
...
 }
 
+void
+init_done_chkpt(struct crt_comp *c)
+{
+	struct crt_comp  *new_comp = boot_comp_get(crt_ncomp() + 1);
+	struct crt_chkpt *chkpt;
+	thdcap_t          thdcap;
+	int               ret;
+	char              name[32];
```
gparmer, line 439: Previously, component names were bounded by `INITARGS_MAX_PATHNAME`. Should keep that.

src/components/implementation/no_interface/llbooter/llbooter.c
```diff
@@ -344,18 +429,64 @@ execute(void)
...
+
+	if (c->id == cos_compid()) {
+	 	/* don't allow chkpnts of the booter */
+	 	BUG();
+	}
+
+	memcpy(name, "chkpt_", sizeof("chkpt_"));
+	assert(strlen(name) < strlen("chkpt_") + strlen(c->name));
```
gparmer, line 447: I don't think that all of the logic around the strings makes sense here.

1. I'm not sure if line 446 is copying the `\0` at the end of the string.
2. this assertion is always true, right? strlen(name) is identical to strlen("chkpt_").
3. the strcat might as well do the right thing despite the assertion check above. If you remove the assertion then 448 can overrun `name`. Assertions are checks, and we shouldn't really rely on their correctness.

All of this can be replaced with

```c
char *prefix = "chkpt_";
int prefix_sz = strlen("chkpt_");
...
assert(INITARGS_MAX_PATHNAME > prefix_sz + strlen(c->name));
memcpy(name, prefix, prefix_sz + 1);  //check my logic, but I think we want to copy the \0
strncat(name, c->name, INITARGS_MAX_PATHNAME - prefix_sz - 1);
c->name[INITARGS_MAX_PATHNAME - 1] = '\0';
```

gparmer: Most important, just make sure that you test it, but you might want to write a small test on Linux. Not sure.

src/components/lib/crt/crt.c
```diff
@@ -55,6 +55,56 @@
 
 #define CRT_REFCNT_INITVAL 1
 
+static unsigned long nchkpt = 0;
+static unsigned long ncomp = 1;
+
+int
+crt_ncomp()
```
gparmer, line 62: see previous comment

(void)

Just consistency, that's all.

src/components/lib/crt/crt.c
```diff
@@ -167,6 +219,76 @@ crt_comp_create_with(struct crt_comp *c, char *name, compid_t id, struct crt_com
...
+
+	/* re-work the sinvs with the right IDs */
+	memcpy(c->sinvs, chkpt->c->sinvs, sizeof(c->sinvs));
+	c->n_sinvs = chkpt->c->n_sinvs;
+	for(u32_t i = 0; i < c->n_sinvs; i++) {
+		struct crt_sinv inv = chkpt->c->sinvs[i];
+		assert(inv.client->id == chkpt->c->id);
+		c->sinvs[i].client = c;
+		assert(inv.server->id != chkpt->c->id);
+
```
gparmer, line 266: empty line here?

src/components/lib/crt/crt.c
```diff
@@ -167,6 +219,76 @@ crt_comp_create_with(struct crt_comp *c, char *name, compid_t id, struct crt_com
...
+	c->entry_addr = chkpt->c->entry_addr;
+	c->ro_addr = chkpt->c->ro_addr;
+	c->rw_addr = chkpt->c->rw_addr;
+
+	/* re-work the sinvs with the right IDs */
+	memcpy(c->sinvs, chkpt->c->sinvs, sizeof(c->sinvs));
+	c->n_sinvs = chkpt->c->n_sinvs;
+	for(u32_t i = 0; i < c->n_sinvs; i++) {
+		struct crt_sinv inv = chkpt->c->sinvs[i];
+		assert(inv.client->id == chkpt->c->id);
```
gparmer, line 263: should be an empty line after variable definition.

gparmer (changes requested):
Very nice. Almost there!

PR #446, opened 2022-03-09
x86-64 PCID and PMU Support

evanstella:

This PR is for an initial review of PCID support. This is not ready to be pulled but will need to be integrated with the namespace implementation that is being developed right now.

This PR adds support for PCID, a kernel API for using PCID for an ASID implementation, as well as PMU programming support for Intel's hardware PMU.

Particularly, I'd like your opinion on:
-  the kernel ASID API and whether it makes sense
- The updated context switch code (`chal_pgtbl_update()`)
- (less important) The PMU programming API and whether this is actually an addition we would want to add to the kernel API

### Intent for your PR

Choose one (Mandatory):

### Reviewers (Mandatory):
@gparmer 

### Code Quality

### Testing

- [ ] micro_booter
- [x] unit_pingpong
- [x] unit_schedtests
- [x] All test outlined in the x86-64 patch


src/platform/i386/chal/chal_proto.h
```diff
@@ -41,21 +45,43 @@ struct cap_pgtbl {
 	struct cap_header h;
 	u32_t             refcnt_flags; /* includes refcnt and flags */
 	pgtbl_t           pgtbl;
-	u32_t             lvl;       /* what level are the pgtbl nodes at? */
-	struct cap_pgtbl *parent;    /* if !null, points to parent cap */
-	u64_t             frozen_ts; /* timestamp when frozen is set. */
+	u32_t             lvl;          /* what level are the pgtbl nodes at? */
+	asid_t            asid;         /* hardware context identifier */
```
gparmer, line 49: This should only be !0 if the lvl == 0. I might forget this fact later when the node is constructed.

gparmer: This fact does point conceptually for the *right place* for this being in the component creation APIs, eventually.

src/platform/i386/chal/chal_proto.h
```diff
@@ -41,21 +45,43 @@ struct cap_pgtbl {
...
 /* Update the page table */
 static inline void
 chal_pgtbl_update(struct pgtbl_info *pt)
-{
-	asm volatile("mov %0, %%cr3" : : "r"(pt->pgtbl));
+{	
+	pgtbl_t *curr_cached = &tlb_asid_active[get_cpuid()][pt->asid];
+
+	/* currently cached pgtbl is this pgtbl */
+	if (*curr_cached == pt->pgtbl) {
```
gparmer, line 63: `likely(...)` This is the fast path.

src/platform/i386/chal/chal_proto.h
```diff
@@ -41,21 +45,43 @@ struct cap_pgtbl {
...
+		__writecr3((unsigned long)pt->pgtbl | pt->asid | CR3_NO_FLUSH);
+		return;
+	}
+
+	/* no pgtbl cached for this asid */
+	if (*curr_cached == 0) {
```
gparmer, line 69: Might not be worth having this case, TBH. Without this case, we'll have an additional TLB flush only for the first time we switch to a new asid. That will happen max 4096 (or whatever the # of PCIDs is) times in the system. Likely not a worthwhile optimization.

src/platform/i386/chal_cpu.h
```diff
@@ -120,7 +121,7 @@ chal_cpu_init(void)
 #if defined(__x86_64__)
 	u32_t low = 0, high = 0;
 
-	chal_cpu_cr4_set(cr4 | CR4_PSE | CR4_PGE);
+	chal_cpu_cr4_set(cr4 | CR4_PSE | CR4_PGE | CR4_PCE | CR4_PCIDE);
```
gparmer, line 124: Check in CPUID if the feature exists on the processor? Bomb out if it doesnt?

src/platform/i386/chal_pgtbl.c
```diff
@@ -660,7 +661,11 @@ chal_pgtbl_pgtblactivate(struct captbl *ct, capid_t cap, capid_t pt_entry, capid
 	ret = cap_kmem_activate(ct, pgtbl_cap, kmem_cap, (unsigned long *)&kmem_addr, &pte);
 	if (unlikely(ret)) return ret;
 	assert(kmem_addr && pte);
- 
+#if defined(__i386__)	
+	/* x86-32 does not support PCID :( */
+	if (asid != 0) return -EINVAL;
```
gparmer, line 666: maybe it is better to just proceed. x86-32 can pass in any asids it wants. The system will still be functionally correct. So it might be better to simply have this code for 32 be the same. In the kernel, we just ignore the value.

We don't want to have to maintain differnet client code for 32 bit vs. 64.

src/platform/x86_64/chal_pmu.c
```diff
@@ -0,0 +1,149 @@
...
+#define IA32_FIXED_CTR_CTRL_USR_EN (1 << 0)
+#define IA32_FIXED_CTR_CTRL_OS_EN  (1 << 1)
+
+#define PMU_NUM_FIXED_PMC    3
+
+/* set during chal initialization */
+extern u8_t processor_num_pmc;
+
+static inline void
+wrmsr(u32_t msr_addr, u32_t in_lo, u32_t in_hi)
```
gparmer, line 48: Are these already in chal_cpu?

src/platform/x86_64/chal_pmu.c
```diff
@@ -0,0 +1,149 @@
...
+
+    if (cntr >= PMU_NUM_FIXED_PMC) return -EINVAL;
+
+    rdmsr(MSR_PERF_GLOBAL_CTRL, &perf_global_ctrl_lo, &perf_global_ctrl_hi);
+    rdmsr(MSR_FIXED_CTR_CTRL, &perf_fixed_ctrl_lo, &perf_fixed_ctrl_hi);
+
+    /* fixed counter enable is in the high dword */
+    perf_global_ctrl_hi |= 1 << cntr;
+    /* enable OS and USR mode counting of event, maybe this should be made optional? */
+    perf_fixed_ctrl_lo  |= (IA32_FIXED_CTR_CTRL_USR_EN | IA32_FIXED_CTR_CTRL_OS_EN) << (cntr * 4);
```
gparmer, line 73: Would be useful to be able to watch only kernel mode events for microoptimization in the kernel. But I don't like having to be considered in the API. What you have here is probably sufficient for now.

src/platform/i386/chal/chal_plat.h
```diff
@@ -7,24 +7,76 @@ int chal_tlbstall(void);
...
+	 * invpcid takes a 128 bit value from memory as: | address | 000...0 | pcid | 
+	 *                                              127        63        11     0
+	 * This is the case in both 32 and 64 bit execution modes
+	 */ 
+	struct { u64_t pcid; u64_t addr; } desc = { .pcid = pcid, .addr = addr };
+
+	asm volatile("invpcid %0, %1" : : "m"(desc), "r"(type) : "memory");
+}
+
+static inline unsigned long
```
gparmer, line 34: I don't see where this and __write_cr3 are used. __write* I can see being used to switch page-tables. but __readcr3?

gparmer:
You do need to add logic into pgtbl_deactivate to remove the asid from the per-cpu table. We need to discuss this as we might be deactivating on a different core than that on which the component executed.

src/components/implementation/tests/unit_pingshmem/ping.c
```diff
@@ -19,6 +19,16 @@ char *pong_test_strings[] = {
 
 shm_bm_t shm;
 
+static unsigned long 
+rdpmc (unsigned long cntr)
```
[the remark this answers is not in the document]
evanstella: Yes, need to figure out what a higher level abstraction would look like

src/components/lib/kernel/cos_defkernel_api.c
```diff
@@ -135,7 +135,7 @@ cos_defcompinfo_child_alloc(struct cos_defcompinfo *child_defci, vaddr_t entry,
 	struct cos_aep_info    *child_aep = cos_sched_aep_get(child_defci);
 
 	assert(curr_defci_init_status == INITIALIZED);
-	ret = cos_compinfo_alloc(child_ci, heap_ptr, cap_frontier, entry, ci);
+	ret = cos_compinfo_alloc(child_ci, heap_ptr, cap_frontier, entry, (asid_t)0, ci);
```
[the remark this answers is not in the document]
evanstella: Not necessary. Wanted more self documenting code because just sticking another 0 in there can be a little confusing. I can remove this though

src/components/lib/kernel/cos_kernel_api.h
```diff
@@ -201,6 +201,9 @@ int     cos_hw_tlbstall(hwcap_t hwc);
 int     cos_hw_tlbstall_recount(hwcap_t hwc);
 void    cos_hw_shutdown(hwcap_t hwc);
 
+int     cos_pmu_enable_fixed_counter(hwcap_t hwc, u8_t cntr);
```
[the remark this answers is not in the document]
evanstella: That's hardware level. Most PMU implementations have two different counters: a set of fixed counters that count pre-programmed events (Intel has 3: hw instruction counter and two separate cycle counters), and programmable counters that can be set to count a different event. The hardware level interface for enabling/programming them is different so I separated them at the kernel level.

src/platform/armv7a/chal/chal_plat.h
```diff
@@ -22,6 +22,11 @@ chal_flush_tlb(void)
 {
 }
 
+static inline void
+chal_flush_tlb_asid(asid_t asid)
+{
+}
```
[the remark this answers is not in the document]
evanstella: Agreed

src/platform/i386/chal/chal_proto.h
evanstella: I agree; The issue is just how to add it to the API with the limited number of input arguments we have right now.

PR #452, opened 2022-05-24
For Review: User-level synchronous invocations and kernel support.

evanstella:

This PR encompasses all my recent work off of @ldierksheide's namespace implementation including an callgate using MPK for user-level synchronous invocations and kernel integration for detecting and handling invocations that are made from userspace. This is **only for review** at this stage; this PR covers a large portion of the system from low-level hardware support to higher-level library abstractions so I'm sure there is a lot that could be changed/improved. 

### Intent for your PR

Choose one (Mandatory):

### Reviewers (Mandatory):
@gparmer @WenyuanShao 

### Code Quality

### Testing

- [x] ping_pong_pang_shared_vas.toml


src/kernel/include/isb.h
```diff
@@ -0,0 +1,46 @@
...
+{
+    struct cap_isb   *isbcap;
+    struct cap_pgtbl *ptcap;
+    paddr_t           pframe = chal_va2pa((void *)kaddr);
+    int               ret;
+
+    ptcap = (struct cap_pgtbl *)captbl_lkup(t, ptcid);
+    if (!ptcap || ptcap->h.type != CAP_PGTBL) return -EINVAL;
+    if (pgtbl_mapping_add(ptcap->pgtbl, uaddr, pframe, PGTBL_USER_DEF, 12)) return -EINVAL;
+
```
gparmer, line 25: This might be a problem with DCB/SCBs as well, but there is no retyping happening here. So that means that the kernel will believe the memory for this is virtual memory, not kernel memory. I'm not sure what problems could come of this, but I'd expect there would be some.

src/kernel/include/shared/cos_types.h
```diff
@@ -370,6 +378,16 @@ typedef unsigned long      invtoken_t;
 #define THDCLOSURE_INIT
 typedef int                thdclosure_index_t;
 
+/* This is an attempt decouple hardware specific code from
+   parts of the kernel interface that are kernel agnostic.
```
gparmer, line 382: Multiline comments:

```
/*
 * comment here...
 * continuing here...
 */
```

gparmer: This enables the use of `/**` and `/***` to emphasize function/high-level comments, and module-level comments.

src/kernel/include/shared/cos_types.h
```diff
@@ -414,6 +432,20 @@ struct cos_stack_freelists {
  * COMP_INFO_TMEM_STK_RELINQ != 0.  Change the defines, or change the assembly" */
 /* #endif */
 
+/* Maybe this should be somewhere else... */
```
gparmer, line 435: Yeah, we can discuss this. Not sure, but certainly feels like an arch-specific abstraction.

src/kernel/include/shared/cos_types.h
```diff
@@ -414,6 +432,20 @@ struct cos_stack_freelists {
...
+struct cos_ulinvstk_entry {
+	capid_t sinv_cap;
+	vaddr_t sp;
+} __attribute__((packed));
+
+/* 1 thread's stack = 2 cache lines */
+#define COS_ULK_INVSTK_SZ 7
```
gparmer, line 442: Not sure how `7` maps to the comment here.

src/kernel/include/shared/cos_types.h
```diff
@@ -414,6 +432,20 @@ struct cos_stack_freelists {
...
+
+struct cos_ulinvstk {
+	u64_t top, pad;
+	struct cos_ulinvstk_entry s[COS_ULK_INVSTK_SZ];
```
gparmer, line 446: s -> stk maybe?

src/kernel/include/shared/cos_types.h
```diff
@@ -414,6 +432,20 @@ struct cos_stack_freelists {
  * COMP_INFO_TMEM_STK_RELINQ != 0.  Change the defines, or change the assembly" */
 /* #endif */
 
+/* Maybe this should be somewhere else... */
+struct cos_ulinvstk_entry {
+	capid_t sinv_cap;
+	vaddr_t sp;
+} __attribute__((packed));
```
gparmer, line 439: Packing this doesn't result the alignments getting wonky?

gparmer: Note that we really want to make sure that the 64 bit value is in memory that is 64 bit aligned. x86 will generate slow code for non-aligned accesses.

src/kernel/include/thd.h
```diff
@@ -13,68 +13,16 @@
 #include "fpu_regs.h"
 #include "chal/cpuid.h"
 #include "chal/call_convention.h"
+#include "chal/chal_thd.h"
```
gparmer, line 16: Removing the `struct thd` is a little extreme here. I would have assumed that removing only parts of the thread that are arch-specific would be possible.

src/kernel/include/thd.h
```diff
@@ -443,73 +394,29 @@ thd_init(void)
...
 {
-	/* curr_thd should be the current thread! We are using cached invstk_top. */
-	struct invstk_entry *curr;
-
-	curr = &curr_thd->invstk[curr_invstk_top(cos_info)];
-	*ip  = curr->ip;
-	*sp  = curr->sp;
-
-	return &curr->comp_info;
+	return chal_invstk_current(curr_thd, ip, sp, cos_info);
```
gparmer, line 400: If a shadow entry is added on, I'm missing why this is required. My mental model (likely missing something) is that if you make the shadow entry on the stack, the most code paths should work without modifications.

src/platform/i386/chal/chal_proto.h
```diff
@@ -46,11 +52,38 @@ struct cap_pgtbl {
...
+	 */
+	wrpkru(~(0b11 << (2 * mpk_key)) & ~0b11);
+}
+
 /* Update the page table */
 static inline void
 chal_pgtbl_update(struct pgtbl_info *pt)
 {
+	/* TODO: PCID SUPPORT */
+	u16_t asid = PROT_DOMAIN_PCID(pt->protdom);
```
gparmer, line 82: Want to understand why the bit packing is the answer here.

src/platform/x86_64/chal/chal_thd.h
```diff
@@ -0,0 +1,186 @@
+#ifndef CHAL_INVSTK
+#define CHAL_INVSTK
```
gparmer, line 2: This variable doesn't match the file name.

CHAL_THD_H

src/platform/x86_64/chal/chal_thd.h
```diff
@@ -0,0 +1,186 @@
...
+	u8_t             ulinvstk_off;
+} HALF_CACHE_ALIGNED;
+
+/**
+ * The thread descriptor.  Contains all information pertaining to a
+ * thread including its registers, id, rcvcap information, and, most
+ * importantly, the kernel invocation stack of execution through
+ * components.
+ */
+struct thread {
```
gparmer, line 44: Everything we put in architecture-specific code, we have to *separately maintain* for each architecture. We want to avoid, where possible, replication between architectures for this reason. We can talk about ways to avoid most of this code in the architecture-specific code.

src/platform/x86_64/chal/chal_thd.h
```diff
@@ -0,0 +1,186 @@
...
+static inline struct comp_info *
+chal_ulinvstk_current(struct cos_ulinvstk *stk, struct comp_info *origin)
+{
+	struct cos_ulinvstk_entry *ent = &stk->s[0];
+	struct comp_info          *ci = origin;
+	struct cap_sinv           *sinvcap;
+	unsigned int               stk_iter;
+
+	/* upper-bounded by ulinvstk sz */
+	for (stk_iter = 0; stk_iter < COS_ULK_INVSTK_SZ && stk_iter < stk->top; stk_iter++) {
```
gparmer, line 115: Generally, when walking through an array, stick to iterating through *indices* as they are easier for the compiler to reason about, in general, than pointer arithmetic.

gparmer: (this will, of course, involve converting `stk->top` into an offset, which is likely the reason you chose this approach here.)

src/platform/x86_64/chal/chal_thd.h
```diff
@@ -0,0 +1,186 @@
...
+		ent = &stk->s[stk_iter];
+		sinvcap = (struct cap_sinv *)captbl_lkup(ci->captbl, ent->sinv_cap);
+		assert(sinvcap->h.type == CAP_SINV);
```
gparmer, line 118: We need actual error checking here, if possible. We obviously need to trust the user-level IS to some degree, but hopefully not for kernel integrity, or for trust between address spaces.

gparmer (commented):
Sorry, hard for me to build up the context on this. This is great work, and we should talk about it in detail on Monday if not earlier @betahxy @WenyuanShao

src/kernel/include/isb.h
evanstella: I believe that `cap_kmem_activate` (in capinv.c) performs the retyping when the memory is activated.

src/components/lib/jitutils/jitutils.h
```diff
@@ -0,0 +1,68 @@
...
+	u64_t inv_placeholder = JIT_INV_PLACEHOLDER;
+	u32_t mpk_placeholder = JIT_MPK_PLACEHOLDER;
+	u32_t pkru_server = ~(0b11 << (2 * srv_pkey)) & ~0b11;
+	u32_t pkru_client = ~(0b11 << (2 * cli_pkey)) & ~0b11;
+
+    /* 
+     * FIXME: 
+	        1) Error handling. There are a few ways this could go wrong
+     *         if one of these values is in the compiled binary already.
+	 *      2) Statically determine these offsets to avoid searching which
```
gparmer, line 57: As this isn't a common-case functionality, I'm less concerned with this optimization. Most important, at this point, is to have a function (this function) that we can replace with whatever we need to fix up the implementation in the future.

src/components/lib/kernel/cos_kernel_api.c
```diff
@@ -742,45 +734,167 @@ __alloc_mem_cap(struct cos_compinfo *ci, cap_t ct, vaddr_t *kmem, capid_t *cap)
 	return 0;
 }
 
-static thdcap_t
-__cos_thd_alloc(struct cos_compinfo *ci, compcap_t comp, thdclosure_index_t init_data)
+/*
```
gparmer, line 737: I can't really really follow the next functions, what they're doing, and if they can be replaced by the composer vas functionality in conjunction with the *general* APIs with checks behind them to do the "right thing". I'm really just ignorant here, and can't reverse engineer this here.

src/components/lib/kernel/cos_kernel_api.c
```diff
@@ -742,45 +734,167 @@ __alloc_mem_cap(struct cos_compinfo *ci, cap_t ct, vaddr_t *kmem, capid_t *cap)
...
+ * allocates N threads, one per core, and uses the coreid + 1 as the
+ * thread's id. Thus, we want to start our thread ids at NUM_CPU + 2.
+ */
+unsigned long __thdid_alloc = NUM_CPU + 2;
+
+
+thdid_t
+cos_thd_id_alloc(void)
+{
+  unsigned long id = ps_faa(&__thdid_alloc, 1);
```
gparmer, line 869: Indentation here is 2 spaces, needs to be a tab.

src/components/lib/kernel/cos_kernel_api.c
```diff
@@ -869,30 +985,62 @@ cos_comp_alloc_shared(struct cos_compinfo *ci, pgtblcap_t ptc, vaddr_t entry, st
...
+static pgtblcap_t
+__alloc_pgtbl_secondlvl(struct cos_compinfo *ci, vaddr_t heap_ptr)
+{
+	pgtblcap_t pte_cap;
+
+	pte_cap = __bump_mem_expand_intern(__compinfo_metacap(ci), ci->pgtbl_cap, heap_ptr, 0, 0);
+	assert(pte_cap);
+
+	/* 
+	 * this should only be called during component initialization
```
gparmer, line 997: I don't understand this comment. Is the argument that this line is, in fact, not needed?

src/components/lib/kernel/cos_kernel_api.c
```diff
@@ -869,30 +985,62 @@ cos_comp_alloc_shared(struct cos_compinfo *ci, pgtblcap_t ptc, vaddr_t entry, st
...
-	compc = cos_comp_alloc(ci_resources, ctc, ptc, entry);
+	compc = cos_comp_alloc(ci_resources, ctc, ptc, entry, protdom);
 	assert(compc);
 
 	cos_compinfo_init(ci, ptc, ctc, compc, heap_ptr, cap_frontier, ci_resources);
 
-	/* This is to make sure that "the address below vas_frontier has been allocated, follow the assumption we put in cos_vasfrontier_init()"*/
-	for (pgtbl_lvl = 0; pgtbl_lvl < COS_PGTBL_DEPTH - 1; pgtbl_lvl++) {
-		__bump_mem_expand_intern(ci, ptc, round_to_page(ci->vas_frontier - 1), 0, pgtbl_lvl);
+	ci->mi.second_lvl_pgtbl_cap  = __bump_mem_expand_intern(__compinfo_metacap(ci), ci->pgtbl_cap, heap_ptr, 0, 0);
```
gparmer, line 1026: The change throughout this file to use the meta instead of the ci is HUGE. Perfectly fantastic if it is fixing a big bug, and is OK, but it is a big risk. Could break existing stuff. Just make sure to test it with enough setups that utilize the code.

src/platform/i386/chal_cpu.h
```diff
@@ -120,7 +121,7 @@ chal_cpu_init(void)
 #if defined(__x86_64__)
 	u32_t low = 0, high = 0;
 
-	chal_cpu_cr4_set(cr4 | CR4_PSE | CR4_PGE);
+	chal_cpu_cr4_set(cr4 | CR4_PSE | CR4_PGE | CR4_PKE);
```
gparmer, line 124: Should we check to make sure that MPK exists before setting this? Will it cause systems to crash that don't define the bit? Generally, we can see if it is supported in `cpuid`. This might be a TODO for the future.

src/kernel/include/ulk.h
```diff
@@ -0,0 +1,48 @@
...
+ulk_activate(struct captbl *t, capid_t ctcap, capid_t ulkcapid, livenessid_t lid, vaddr_t kaddr, capid_t ptcid, vaddr_t uaddr)
+{
+    struct cap_ulk   *ulkcap;
+    struct cap_pgtbl *ptcap;
+    paddr_t           pframe = chal_va2pa((void *)kaddr);
+    int               ret;
+
+    ptcap = (struct cap_pgtbl *)captbl_lkup(t, ptcid);
+    if (!ptcap || ptcap->h.type != CAP_PGTBL) return -EINVAL;
+    if (pgtbl_mapping_add(ptcap->pgtbl, uaddr, pframe, PGTBL_USER_DEF | ULK_PGTBL_FLAG, 12)) return -EINVAL;
```
gparmer, line 26: I think that we have a PAGE_ORD or some such macro instead of `12`.

src/kernel/include/ulk.h
```diff
@@ -0,0 +1,48 @@
...
+
+    ulkcap = (struct cap_ulk *)__cap_capactivate_pre(t, ctcap, ulkcapid, CAP_ULK, &ret);
+    if (!ulkcap) return -EINVAL;
+
+    ltbl_get(lid, &ulkcap->liveness);
+    ulkcap->kern_addr = kaddr;
```
gparmer, line 32: Should also store the uaddr (for the eventual removal).

gparmer: There's a challenging design issue here. We'll need to support *multiple* pages as there are more threads than their ulis can fit into a single page. So we'll need to change the thread creation path as well, and store the kernel address in the thread structure.

The code you have here is likely perfectly fine for that. But it is yet another capability (to this page) to pass into the thread creation. Worth discussing to make sure that we have a decent plan.

Note, you might have already solved this, and I simply haven't gotten to it in your review. Ignore if that's the case ;-)

src/platform/armv7a/boot_comp.c
```diff
@@ -19,6 +19,8 @@ struct cap_asnd hw_asnd_caps[HW_IRQ_TOTAL];
 void *thd_mem[NUM_CPU], *tcap_mem[NUM_CPU];
 struct captbl *glb_boot_ct;
 
+thdid_t tid = 0;
```
gparmer, line 22: Comment with a `TODO` to make additions to this atomic. This feels like it should *really* be `static`.

src/platform/i386/chal/chal_proto.h
```diff
@@ -46,11 +46,34 @@ struct cap_pgtbl {
 	u64_t             frozen_ts; /* timestamp when frozen is set. */
 } __attribute__((packed));
 
+static inline void
```
gparmer, line 49: Can these be removed now that you have the callgates?

gparmer (commented):
This looks pretty great! small tweaks, and integration with the new composer.

I want to go through the kernel update with a fresh brain, so will follow up with that in 24 hours.

evanstella:
we can talk about the kernel update in our meeting; i remember not being completely convinced my solution was the best we could come up with> This looks pretty great! small tweaks, and integration with the new composer.
> 
> 
> 
> I want to go through the kernel update with a fresh brain, so will follow up with that in 24 hours.

src/kernel/include/thd.h
```diff
@@ -500,16 +497,106 @@ thd_invstk_push(struct thread *thd, struct comp_info *ci, unsigned long ip, unsi
...
+	struct invstk_entry *curr;
+	struct comp_info    *ci;
+
 	if (unlikely(curr_invstk_top(cos_info) == 0)) return NULL;
 	curr_invstk_dec(cos_info);
-	return thd_invstk_current(thd, ip, sp, cos_info);
+	curr = &thd->invstk[curr_invstk_top(cos_info)];
+
+	/* if we pushed over a UL inv entry, need to pop that too */
+	if (curr->ulk_stkoff == THD_UL_INV_ORIGIN) {
```
gparmer, line 547: `unlikely`?

src/kernel/include/thd.h
```diff
@@ -500,16 +497,106 @@ thd_invstk_push(struct thread *thd, struct comp_info *ci, unsigned long ip, unsi
...
+
+	return 0;
+}
+
+static inline void
+thd_invstk_pop_ulk(struct thread *thd, struct cos_cpu_local_info *cos_info)
+{
+	struct invstk_entry *curr;
+
+	if (unlikely(curr_invstk_top(cos_info) == 0)) return;
```
gparmer, line 527: If the top is already calculated in teh calling `syscall` function, then just pass it in. Redundant computations on the fastpath here aren't great, and it would be good if this function came down to just the conditional below. If not that, then maybe conditionally call this if there was actually a `push` (which we can track in a varaible in the calling function).

src/kernel/include/thd.h
```diff
@@ -500,16 +497,106 @@ thd_invstk_push(struct thread *thd, struct comp_info *ci, unsigned long ip, unsi
...
+	curr = &thd->invstk[curr_invstk_top(cos_info)-1];
+
+	if (curr->ulk_stkoff == THD_UL_INV_ORIGIN) {
```
gparmer, line 530: unlikely.

src/kernel/include/thd.h
```diff
@@ -500,16 +497,106 @@ thd_invstk_push(struct thread *thd, struct comp_info *ci, unsigned long ip, unsi
...
 
 	return 0;
 }
 
+static inline int
+thd_invstk_push_ulk(struct thread *thd, struct comp_info *ci, struct cos_cpu_local_info *cos_info)
+{
+	struct invstk_entry *top;
+
+	if (unlikely(curr_invstk_top(cos_info) >= THD_INVSTK_MAXSZ)) return -1;
```
gparmer, line 511: WE're recomputing the top many times. Verify it is fast to compute.

src/kernel/include/thd.h
```diff
@@ -500,16 +497,106 @@ thd_invstk_push(struct thread *thd, struct comp_info *ci, unsigned long ip, unsi
...
+	return 0;
+}
+
+static inline void
+thd_invstk_pop_ulk(struct thread *thd, struct cos_cpu_local_info *cos_info)
+{
+	struct invstk_entry *curr;
+
+	if (unlikely(curr_invstk_top(cos_info) == 0)) return;
+	curr = &thd->invstk[curr_invstk_top(cos_info)-1];
```
gparmer, line 528: `-1` -> ` - 1` Remember that binary operations use spaces.

src/kernel/include/thd.h
```diff
@@ -500,16 +497,106 @@ thd_invstk_push(struct thread *thd, struct comp_info *ci, unsigned long ip, unsi
...
+		*ip = curr->ip;
+		*sp = curr->sp;
+		return &curr->comp_info;
+	}
+
+	/* 
+	 * if there are new entries on the UL stack, 
+	 * we must be coming from a comp on the UL stack 
+	 */
+	if (ulk_invstk->top > curr->ulk_stkoff) {
```
gparmer, line 583: I'm missing how `ulk_stkoff` is passed to subsequent stack entries. Would need it if we had ULINV -> KINV -> *KINV* -> ULINV -> KINV. In the highlighted KINV, we'd need to continue tracking were we were in the ULINV from the past calculation, right?

gparmer (changes requested):
I'm not confident I can understand the current code's interactions between the invocation stack and UL invstk. I think that adding the push to the start of the syscall will clarify that. I'm hoping, also, that there are more optimizations we can apply here. This needs to be very intentionally designed around the normal sinv fast-path.

I think that from an optimization point of view, we want to make sure that the code paths with no ULinvs is optimized for in terms of performance and readability. I think that we also need a better name for `THD_UL_INV_ORIGIN`.

PR #457, opened 2022-07-21
Simple TCP echo server component  + LWIP component + DPDK component implementation, for discussion

betahxy:

Adds a simple tcp server component and test the full path of receiving & transmitting of the system. There are three components that need to be read:

+ The DPDK/NIC component: it uses a single thread to process rx/tx packets, with the help of ck_ring buffer. It has 2 ck_ring buffers, one for receiving packets, and the other for transmitting packets. Notice that it does not have a hash table to redirect connections because currently we only have one single thread tcp server to process requests. Another small idea there is that I want to support the ARP protocol, so that the connection does not need to configure MAC addresses. That's why there is a switch condition within the receiving loop.

+ The LWIP component: it implements a `netmgr` interface, which is similar to linux style socket API. For example, it implements functions like `netmgr_tcp_bind`, `netmgr_tcp_listen` and `netmgr_tcp_accept`. Notice that the `netmgr_tcp_accept` is a blocking API. Current implementation only uses three gloabl variables `return_flag`, `g_pbuf` and `g_objid` to indicate whether the thread should return from `tcp_recv`. The LWIP sending logic `cos_interface_output` is a little opeaque because: 1. It needs to know whether the pbuf comes from a shared memory or not. 2. If it is not a shared memory case, it needs to copy the full packet content to the shared memory for sending. 3. If it is a shared memory case, the pbuf is chained actually 2 pieces. The first piece is the packet header created by the lwip, the second one is the shared memory referenced by a `PBUF_ROM` pbuf. Thus, we need to copy the headers to our shared memory for DPDK to send it out. Normally, the headers consist of a 20-bytes TCP header, 20-bytes IP header and a 14-byte ether header, totally 54 bytes that needs to be copied when sending out shared memory packets.

+ The TCP application component: It uses interfaces provided by the shared memory and `netmgr` to process connections. Each time when it receives a message (a packet/a shared memory), it allocates a sending shared memory block and copy the original receiving data to the sending buffer. And finally send it out.

There is also a simple python based tcp client (`tcp-test.py`) to test the tcp echo server. 

There are still some other infrastructures needed, like the DPDK connection hashing function, a more user-friendly share memory API. Thus,  this PR is only for a code review to see if there are some major problems of the implementation of our network stack facilities. 

Thank you very much for the review.

### Intent for your PR

Choose one (Mandatory):

### Reviewers (Mandatory):
@gparmer @WenyuanShao @evanstella 

### Code Quality

### Testing

- [ ] micro_booter
- [ ] unit_pingpong
- [ ] unit_schedtests
- [ ] ...(add others here)

### DIFF
--- a/src/components/implementation/netmgr/lwip/interface_impl.c
+++ b/src/components/implementation/netmgr/lwip/interface_impl.c
@@ -1,0 +1,232 @@
+#include <cos_types.h>
+#include <cos_component.h>
+#include <netshmem.h>
+#include <shm_bm.h>
+#include <string.h>
+#include <nic.h>
+#include <contigmem.h>
+
+#include <lwip/init.h>
+#include <lwip/netif.h>
+#include <lwip/tcp.h>
+#include <lwip/stats.h>
+#include <lwip/prot/tcp.h>
+#include <netif/ethernet.h>
+#include <lwip/etharp.h>
+
+#include <netmgr.h>
+
+int return_flag = 0;
+shm_bm_objid_t           g_objid;
+struct pbuf *g_pbuf = NULL;
+
+#define TCP_MAX_SERVER (16)
+
+extern struct netif net_interface;
+
+struct tcp_server
+{
+	struct tcp_pcb *tp;
+};
+
+struct tcp_server tcp_servers[TCP_MAX_SERVER];
+
+static err_t
+cos_lwip_tcp_sent(void *arg, struct tcp_pcb *tp, u16_t len)
+{
+	// printc("tcp sent\n");
+	return ERR_OK;
+}
+
+
+static err_t
+cos_lwip_tcp_recv(void *arg, struct tcp_pcb *tp, struct pbuf *p, err_t err)
+{
+	err_t ret_err;
+
+	if(p != NULL) {
+		return_flag = 1;
+		g_pbuf = p;
+		// tcp_recved(tp, p->tot_len);
+	}
+
+	return ERR_OK;
+
+}
+
+static void
+cos_lwip_tcp_err(void *arg,  err_t err)
+{
+	assert(0);
+	return;
+}
+
+static err_t
+cos_lwip_tcp_accept(void *arg, struct tcp_pcb *tp, err_t err)
+{
+	err_t ret_err;
+	compid_t client = (compid_t)cos_inv_token();
+
+	tcp_arg(tp, 0);
+	tcp_err(tp, cos_lwip_tcp_err);
+	tcp_recv(tp, cos_lwip_tcp_recv);
+	tcp_sent(tp, cos_lwip_tcp_sent);
+
+	tcp_servers[client].tp = tp;
+
+	tcp_nagle_disable(tp);
+
+	ret_err = ERR_OK;
+	return ret_err;
+}
+
+void netmgr_shmem_init(cbuf_t rx_shm_id, cbuf_t tx_shm_id)
+{
+	nic_shmem_init(rx_shm_id, tx_shm_id);
+	netshmem_map_shmem(rx_shm_id, tx_shm_id);
+}
+
+int
+netmgr_tcp_bind(u32_t ip_addr, u16_t port)
+{
+
+	unsigned long npages;
+	void         *mem;
+	shm_bm_t shm;
+	struct netshmem_pkt_buf *obj;
+
+	struct tcp_pcb *tp;
+	struct ip4_addr ipa = *(struct ip4_addr*)&ip_addr;
+	err_t ret;
+
+	compid_t client = (compid_t)cos_inv_token();
+	unsigned long paddr = 0;
+
+	nic_bind_port(ip_addr, htons(port));
+
+	tp = tcp_new();
+	assert(tp != NULL);
+
+	tcp_servers[client].tp = tp;
+
+	ret = tcp_bind(tp, &ipa, port);
+
+	return ret;
+}
+
+int
+netmgr_tcp_listen(u8_t backlog)
+{
+	struct tcp_pcb *new_tp = NULL;
+	compid_t client = (compid_t)cos_inv_token();
+
+	new_tp = tcp_listen_with_backlog(tcp_servers[client].tp, backlog);
+	assert(new_tp);
+
+	tcp_servers[client].tp = new_tp;
+
+	return ERR_OK;
+}
+
+static void
+net_interface_input(void * pkt, int len)
+{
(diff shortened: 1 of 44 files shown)


src/components/implementation/netmgr/lwip/interface_impl.c
```diff
@@ -0,0 +1,232 @@
...
+
+	tcp_servers[client].tp = tp;
+
+	ret = tcp_bind(tp, &ipa, port);
+
+	return ret;
+}
+
+int
+netmgr_tcp_listen(u8_t backlog)
```
WenyuanShao, line 118: Really good job getting listen and bind working. We used to don't have these functions. It should make port applications much easier.

src/components/implementation/netmgr/lwip/interface_impl.c
```diff
@@ -0,0 +1,232 @@
...
+	p = pbuf_alloc(PBUF_LINK, len, PBUF_ROM);
+
+	assert(p);
+
+	p->payload = pl;
+	if (net_interface.input(p, &net_interface) != ERR_OK) {
+		assert(0);
+	}
+
+	if (p->ref != 0) {
```
WenyuanShao, line 148: Is this an error condition? If it is, instead of lwip reporting error elsewhere, I'd prefer adding an assert here.

src/components/implementation/nicmgr/dpdk/init.c
```diff
@@ -1,20 +1,187 @@
...
+
+	g_tx_ring    = buf_addr;
+	g_tx_ringbuf = buf_addr + sizeof(struct ck_ring);
+}
+
+static struct client_session *
+find_session(uint32_t dst_ip, uint16_t dst_port)
 {
 	int i;
+	for (i = 0; i < NIC_MAX_SESSION; i++) {
```
WenyuanShao, line 101: Seems you iterate every session while doing a search. It is good for now but may result in additional overheads. Using hash map may be another solution.

src/components/implementation/nicmgr/dpdk/interface_impl.c
```diff
@@ -0,0 +1,203 @@
+#include <cos_types.h>
```
WenyuanShao, line 1: I don't like the name of this file. iterface_impl makes very less sense to the user. Maybe rename it as nic_mgr_coordination.c or something like that.

WenyuanShao (commented):
Great PR for lwip. My comments mainly on like small places I don't understand. Really appreciate the effort.

src/components/implementation/tests/bench_dpdk/tcp-test.py
```diff
@@ -0,0 +1,29 @@
...
+time.sleep(5)
+
+data = sock.recv(1000)
+print(data.decode())
+
+
+sock.send('second hello from client\n'.encode())
+
+data = sock.recv(1000)
+print(data.decode())
```
gparmer, line 27: I'm not sure if you have a test that sends/receives quite a bit more, but I'm having trouble validating that you're using the reference counting on the shared memory correctly. We can test that by sending/receving a lot of data to ensure that we don't run out of shared memory.

src/components/interface/netshmem/lib.c
```diff
@@ -0,0 +1,75 @@
+#include "netshmem.h"
```
gparmer, line 1: Some documentation as to what this library actually does would be useful. I read through most of the code, and still don't really understand its purpose. It likely has a good one, I just don't get it ;-)

src/components/lib/posix_legacy/posix_legacy.c
```diff
@@ -291,7 +291,7 @@ cos_clone(int (*func)(void *), void *stack, int flags, void *arg, pid_t *ptid, v
 		return -1;
 	}
 
-	struct sl_thd * thd = sl_thd_alloc((cos_thd_fn_t)func, arg);
+	struct sl_thd * thd = sl_thd_alloc((cos_thd_fn_t)(void *)func, arg);
```
gparmer, line 294: Thanks for killing warnings!!!

gparmer (commented):
Sorry for too many comments here. This is *fantastic code*, and you've done a great job! Most of this is very small things. Many are small comments about style which you do want to nail down so that we have consistent code. Not many comments of huge substance. My larger concern is about properly using the shared memory refcnts (which you might be doing OK!), and storing sensitive information in the shared memory.

src/components/implementation/netmgr/lwip/interface_impl.c
```diff
@@ -0,0 +1,232 @@
...
+}
+
+
+static err_t
+cos_lwip_tcp_recv(void *arg, struct tcp_pcb *tp, struct pbuf *p, err_t err)
+{
+	err_t ret_err;
+
+	if(p != NULL) {
+		return_flag = 1;
```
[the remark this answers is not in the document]
betahxy: @WenyuanShao Each time when LWIP receives a packet, it needs to figure out if this packet is a data packet to applications or it is just a state packet (eg, tcp states) that don't need to send to applications. We use a return flag to indicate if LWIP should return and send data to applications.

src/components/interface/nic/nic.h
```diff
@@ -4,8 +4,17 @@
 #include <cos_types.h>
 #include <cos_component.h>
 #include <cos_stubs.h>
+#include <shm_bm.h>
 
-int nic_send_packet(char* pkt, size_t pkt_size);
-int nic_bind_port(u32_t ip_addr, u16_t port, void* share_mem, size_t share_mem_sz);
+void nic_shmem_map(cbuf_t shm_id);
 
+int nic_send_packet(shm_bm_objid_t pktid, u16_t pkt_offset, u16_t pkt_len);
```
gparmer, line 11: This requires the client and server to manually do all of the conversions themselves for the shared memory, and means that *cannot* replace this shared memory implementation with another for this interface.

On Monday we need to talk about how to hide all of the mechanics behind the shared memory of the API, and make the interface simply pass pointers.

I'm fine with this PR currently, but this mistake cannot continue to pervade the system. The amount of refactoring in the future to fix it will be untenable.

gparmer (approved):
I really cannot properly review this. The list of what you've done is not something I can use to go through the code as I don't have a good understanding of which files are impacted. However, it looks good for now, and is a TON of brilliant work. I think we can go back and fix anything that needs fixing later.

gparmer:
Thanks for all the work on this! Looks great.

PR #465, opened 2023-01-24
composer: remove dead pipers package

pskrgag:
pipers crate seems to be archived (https://github.com/mgattozzi/pipers), so composer fails to build in new environment.

Removed it from the dependencies and added bare minimal interface to replace it.

Tested locally on top of loader branch

Signed-off-by: Pavel Skripkin <paskripkin@gmail.com>


gparmer:
This is brilliant! We're in "push mode" for a deadline, and have been using/sharing the an old Cargo.lock to delay fixing the pipers dependency. Then you do this PR! Thank you so much: you've saved us a lot of hastle and time.

Looking at your github: your buddy allocator is fantastic. We've done a lot of work on lock and wait-free algorithms (a few of which are the foundation for parallelism in Composite.

If you're ever interested in doing a PhD, feel free to reach out. We're doing a lot of exciting work toward virtualization and verification, including using rust more broadly in the system.

pskrgag:
@gparmer Thanks for merging!

lock free buddy master thesis is not my work, I just found it on the internet and reimplemented it in Rust for my purposes

Just wanted to make it clear :) 

I'd love to take a PhD one day, but only after I finish my MS.

Thanks once again! :)

PR #466, opened 2023-01-25
composer: skip empty arguemnts

pskrgag:
I've messed up in previous patch. .split(' ') splits by exactly one whitespace, which causes compose failure when no optional make arguments provided. Fix it by using proper whitespace parsing API 

I just hope nobody noticed

Signed-off-by: Pavel Skripkin <paskripkin@gmail.com>


gparmer:
Appreciate the follow-up! We haven't pulled yet, so no harm done. Our fault for still not having ci set up.

PR #468, opened 2023-02-11
PR for supporting cross core wakeup for review.

WenyuanShao:

**Add description here.**

### Intent for your PR

Choose one (Mandatory):

### Reviewers (Mandatory):

### Code Quality

### Testing

- [ ] micro_booter
- [ ] unit_pingpong
- [x] unit_schedtests
- [ ] ...(add others here)

### DIFF
--- a/src/components/implementation/sched/pfprr_quantum_static/main.c
+++ b/src/components/implementation/sched/pfprr_quantum_static/main.c
@@ -44,7 +44,7 @@ struct slm_notify_buf {
 	struct ck_ring *ring;
 	struct slm_notify *ringbuf;
 };
-struct slm_notify_buf slm_xcpu_notify;
+struct slm_notify_buf slm_xcpu_notify[NUM_CPU];
 CK_RING_PROTOTYPE(slm_notify_buf, slm_notify);
 
 struct slm_thd *slm_thd_static_cm_lookup(thdid_t id);
@@ -244,8 +244,8 @@ thd_wakeup(struct slm_thd *t)
 	} else {
 		struct slm_notify notify;
 		notify.tid = t->tid;
-		ret = slm_notify_queue_enqueue(&slm_xcpu_notify, &notify);
-		assert(!slm_notify_queue_empty(&slm_xcpu_notify));
+		ret = slm_notify_queue_enqueue(&slm_xcpu_notify[t->cpuid], &notify);
+		assert(!slm_notify_queue_empty(&slm_xcpu_notify[t->cpuid]));
 		cos_asnd(ipithd->asnd, 1);
 	}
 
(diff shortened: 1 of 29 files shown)


gparmer (commented):
A thought that I didn't fit into a specific comment below: should most of the IPI stuff be in slm? It is a little awkward in main. Likely something we should consider in the future.

Generally pretty great!!! I think that some refactoring of the ipi data-structures (see my suggestion) would make this all quite a bit simpler.

src/components/lib/slm/slm.c
```diff
@@ -312,9 +347,17 @@ int
 slm_thd_wakeup(struct slm_thd *t, int redundant)
 {
 	assert(t);
+	if (unlikely(t->cpuid != cos_cpuid())) {
+		struct slm_ipi_percore *ipi_data = slm_ipi_percore_get(t->cpuid);
+		struct slm_ipi_event    event    = { 0 };
+		event.tid = t->tid;
+		int ret = slm_ipi_event_enqueue(&event, t->cpuid);
+		assert(!slm_ipi_event_empty(t->cpuid));
```
gparmer, line 355: This can fail if the dequeue happens quickly on the other core. Good assert for debugging, but not good for the PR, I don't think.

src/components/implementation/sched/pfprr_quantum_static/main.c
```diff
@@ -630,48 +574,31 @@ void
...
+		assert(!slm_ipi_event_empty(cos_cpuid()));
 
-		slm_cs_enter(current, SLM_CS_NONE);
-		while (!slm_notify_queue_empty(&slm_xcpu_notify[cos_cpuid()])) {
-			slm_notify_queue_dequeue(&slm_xcpu_notify[cos_cpuid()], &notify);
-			thd = slm_thd_static_cm_lookup(notify.tid);
+		while (!slm_ipi_event_empty(cos_cpuid())) {
+			slm_ipi_event_dequeue(&event, cos_cpuid());
+			thd = slm_thd_static_cm_lookup(event.tid);
+			slm_cs_enter(current, SLM_CS_NONE);
```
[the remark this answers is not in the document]
gparmer: Lets look at the options, ranked from best to worst. I could be wrong on the ranking, but I figured you might want to see optiosn and why they are good/bad:

1. assert out in the error case. We are assuming that `slm_thd_wakeup` should *not* return an error as we derived `thd` from scheduler-internal data-structures.  This is not strictly correct in all cases as the thread might have been deleted before this call, but then we have larger restructuring issues. Given the sassumptions of the current system, it feels correct to just assert here.
2. Log the error, drop the notification and move on to the next iteration. I don't like this option as having to look at a log to discern a logic error is not good. I think it is better than silently failing.
3. return from this function silently. This is bad as it means developers will see errors that are unrelated to this loop (wakeups stop firing, e.g. "when I take a lock, it never wakes up me...thus the error must be in the lock"), so it will be hard for them to zero-in on this code.

gparmer (changes requested):
A few small/fast fixes needed here. Great job with the restructuring. Everything looks much simpler now.

src/components/implementation/tests/unit_ipitest/unit_ipi.c
```diff
@@ -64,7 +66,10 @@ parallel_main(coreid_t cid, int init_core, int ncores)
 {
 	int i = 0;
 
-	if (NUM_CPU < 2) return;
+	if (NUM_CPU < 2) {
+		printc("ERROR: This test requires more than 2 cores.\n");
```
gparmer, line 70: Why does this require MORE THAN 2 cores? That isn't consistent with the logic.

PR #474, opened 2023-09-11
PR before the MPK and Slite merge

betahxy:

This PR to the main branch is for the MPK and Slite merge as suggested by @WenyuanShao . This PR only contains the application changes without MPK and Slite. 

### Intent for your PR

Choose one (Mandatory):

### Reviewers (Mandatory):
@gparmer @WenyuanShao


src/components/implementation/capmgr/simple/capmgr.c
```diff
@@ -842,11 +853,17 @@ cos_init(void)
 	 * after we've sealed in all initargs and sinvs. Regardless,
 	 * that is the *correct* approach.
 	 */
+	// printc("addr_get(cos_compid(), ADDR_CAPTBL_FRONTIER):%d\n", addr_get(cos_compid(), ADDR_CAPTBL_FRONTIER));
```
WenyuanShao, line 856: remove unused code.

src/components/implementation/nicmgr/dpdk/init.c
```diff
@@ -112,7 +138,7 @@ process_tx_packets(void)
 	char *tx_packets[MAX_PKT_BURST];
 	struct pkt_ring_buf *per_thd_tx_ring;
 	int i = 0, j = 0;
-
+	// cycles_t    before, after;
```
WenyuanShao, line 141: Again remove unused code in the main branch.

src/components/implementation/nicmgr/dpdk/nicmgr.c
```diff
@@ -129,12 +153,32 @@ nic_send_packet(shm_bm_objid_t pktid, u16_t pkt_offset, u16_t pkt_len)
 	
 	buf.paddr   = data_paddr;
 	buf.pkt_len = pkt_len;
-
+#if 0
```
WenyuanShao, line 156: What's the point of this check? Remove the code if we don't need or change the macro to something make sense.

src/components/implementation/sched/pfprr_quantum_static/init.c
```diff
@@ -239,7 +239,7 @@ slm_comp_init_loop(void)
 
 		/* Create the thread for initialization of the next component */
 		extern struct slm_thd *thd_alloc_in(compid_t id, thdclosure_index_t idx, sched_param_t *parameters, int reschedule);
-		t = thd_alloc_in(client, 0, param, 0);
+		t = thd_alloc_in(client, 0, param, 1);
```
WenyuanShao, line 242: I don't understand, why change it to reschedule?

src/components/lib/component/arch/x86_64/cos_asm_simple_stacks.h
```diff
@@ -2,24 +2,15 @@
...
-	/* restore the tid in order to save it on the stack*/			\
-	shr $MAX_STACK_SZ_BYTE_ORDER, %rax;					\
-	/* get the cpuid by right shifting the lower 16 bits*/			\
-	shr $16, %rdx;								\
-	/* save the cpuid on the top of the thread stack */			\
-	pushq %rdx;								\
-	/* save the tid followed by the cpuid */				\
+#define COS_ASM_GET_STACK_BASIC									\
+	/* we save the return position to rcx for custom stack acquisition code to return */ 	\
+	movabs $1f, %rcx;									\
```
WenyuanShao, line 7: This is where the bug happens forMPK callgates. just change rcx to r13.

src/components/lib/sync/sync_lock.h
```diff
@@ -73,6 +73,9 @@ sync_lock_teardown(struct sync_lock *l)
 static inline void
 sync_lock_take(struct sync_lock *l)
 {
+#if 1
```
WenyuanShao, line 76: What is this? We are not suppose to have this code in the main line. Either change it to some macro which makes sense or delete the #else part.

src/components/lib/sync/sync_lock.h
```diff
@@ -105,11 +109,15 @@ sync_lock_take(struct sync_lock *l)
 static inline int
 sync_lock_try_take(struct sync_lock *l)
 {
+#if 1
```
WenyuanShao, line 112: Again

src/components/lib/sync/sync_lock.h
```diff
@@ -122,6 +130,9 @@ sync_lock_try_take(struct sync_lock *l)
 static inline void
 sync_lock_release(struct sync_lock *l)
 {
+#if 1
```
WenyuanShao, line 133: Here as well.

src/components/lib/udp_stack/simple_udp_stack.c
```diff
@@ -13,6 +13,7 @@ static u32_t host_ip;
 static u16_t host_port;
 
 static struct ether_addr nic_mac;
+#if 1
```
WenyuanShao, line 16: This as well.

WenyuanShao (changes requested):

src/components/lib/sync/sync_lock.h
WenyuanShao: Also, this function generates a ton of warnings during the compilation.

src/components/lib/kernel/cos_kernel_api.c
```diff
@@ -717,6 +703,104 @@ __page_bump_alloc(struct cos_compinfo *ci, size_t sz, size_t align)
 	return heap_vaddr;
 }
 
+void
+missing_captbl_node_expand(struct cos_compinfo *ci)
```
WenyuanShao, line 707: What is the problem this function trying to solve? Is it solving the problem when CAPTBL_FREE_CAP > CAPTBL_EXPAND_SZ the second captabl has not been allocated?

WenyuanShao: This is too complicated as far as I'm concerned.

src/components/lib/udp_stack/simple_udp_stack.c
betahxy: @WenyuanShao @gparmer We cannot remove these #define or #if (this also makes sense for the previous similar comments @WenyuanShao makde, I reply all of them here) for two reasons:1. We need the code work for both on Qemu and the hardware. 2. We have potential bugs and perf problems, we need code to enable us detect them quickly. Of course there could be better ways to do this, but we are limited by time. Based on these, I think we should not remove them. If you think the reasons don't make sense, then I'll remove them.

src/components/implementation/sched/pfprr_quantum_static/init.c
betahxy: @WenyuanShao This code change was made by you, I merged it from you commit. I have no idea about it.

src/components/lib/kernel/cos_kernel_api.c
betahxy: I forgot the full background here. But generally it relates to that I couldn't allocate sinv cap in the past with more cores, thus I added some checking facilities in the kernel and user space to make sure I have cap for the sinv cap. This is needed I think.

src/components/implementation/sched/pfprr_quantum_static/main.c
```diff
@@ -582,10 +589,11 @@ slm_ipi_process(void *d)
 
 	while (1) {
 		cos_rcv(r->rcv, RCV_ALL_PENDING, &rcvd);
-		assert(!slm_ipi_event_empty(cos_cpuid()));
 
+		// printc("$\n");
```
gparmer, line 593: commented code to remove.

src/components/implementation/simple_mc_udp_server/simple_mc_udp_server_no_lwip/main.c
```diff
@@ -4,18 +4,79 @@
...
 {
 	shm_bm_objid_t objid;
 	struct netshmem_pkt_buf *obj;
-
+	init_thd = cos_thdid();
 	/* create current component's shmem */
 	netshmem_create();
+}
+void
+cos_parallel_init(coreid_t cid, int init_core, int ncores)
```
gparmer, line 22: This function looks quite odd. Should we discuss? Seems to assume a very specific composition script, with a very specific setup.

src/components/implementation/simple_mc_udp_server/simple_mc_udp_server_no_lwip/main.c
```diff
@@ -4,18 +4,79 @@
...
+	}
+	if (cos_compid() == 20 && cid != 15) {
+		return;
+	}
+	if (cos_compid() == 21 && cid != 16) {
+		return;
+	}
 
+	if (init_thd != cos_thdid()) {
+		netshemem_move(init_thd, cos_thdid());
```
gparmer, line 78: First I noticed this name. `netshemem` looks like "net she memory", which isn't very informative. If it is supposed to be for shared memory, then the `e` is throwing off the name.

src/components/lib/kernel/cos_kernel_api.c
gparmer: I'm worried that this looks like it might be solving a bug for your specific scenario, but not for the general case. And it might make that general case more challenging to debug when problems do happen.

I certainly am not questioning if this is solving a bug. I'm sure it is. But I worry that it is just hiding the symptom of a bug in a specific instance. Do you have any recollection if this is a "final solution" to a bug, or just a "workaround"?

src/components/lib/slm/fprr.c
```diff
@@ -20,13 +22,45 @@ void
 slm_sched_fprr_execution(struct slm_thd *t, cycles_t cycles)
 { return; }
 
+// simply dump a core's task queue for debug usage
```
gparmer, line 25: I think that this is much more specific than this. shouldn't the comment be closer to:

```c
/*
 * Dump core 0's runqueue information periodically (and only from core 0). Will only do so after this function has been called a configurable number of times.
 */
```

Also, recall that comments should use `/* */`.

src/components/lib/sync/sync_lock.h
gparmer: We really cannot add warnings. No code we add (aside from e.g. external libraries) should add warnings.

src/components/lib/udp_stack/simple_udp_stack.c
```diff
@@ -63,12 +75,13 @@ udp_stack_eth_hdr_set(struct eth_hdr *eth_hdr, struct ether_addr* src_mac, struc
...
 udp_stack_ip_hdr_set(struct ip_hdr *ip_hdr, u16_t data_len, u32_t src_host, u32_t dst_host)
 {
+	static u16_t ip_id = 0;
 	/* We don't support complex IP options */
 	ip_hdr->ihl = IP_STD_LEN / 4;
 	ip_hdr->version = IPv4;
 	ip_hdr->tos = 0;
 	ip_hdr->total_len = htons(IP_STD_LEN + data_len);
-	ip_hdr->id = 0;
+	ip_hdr->id = ++ip_id;
```
gparmer, line 84: I think you're obviously making a lot of strong assumptions about what we want here. That's fine. Please document them. This just seems like magic for now. What is the ip_id? Why are you setting it to increasing values, etc... Comment!

src/kernel/capinv.c
```diff
@@ -1307,6 +1345,21 @@ static int __attribute__((noinline)) composite_syscall_slowpath(struct pt_regs *
 
 			break;
 		}
+		case CAPTBL_OP_CAPCHECK: {
```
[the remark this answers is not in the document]
gparmer: Any reason this type of functionality can't be addded there?

gparmer (changes requested):
This is all quite fantastic.  A ton of work.

I'm concerned about a lot of hacks being added in without documentation nor an understanding (at least by me) of how to avoid the hacks in the longer term. I'm not asking for too many changes that remove hacks, but am asking for better documentation.

PR #479, opened 2024-06-05
implementation of define constant value

mbai1010:

**customize a #define constant for a component's compilation**

### DIFF
--- a/src/composer/src/build.rs
+++ b/src/composer/src/build.rs
@@ -4,6 +4,8 @@ use std::env;
 use std::fs::File;
 use syshelpers::{dir_exists, emit_file, exec_pipeline, reset_dir};
 use tar::Builder;
+use std::process::Command;
+use std::io::{self, BufRead, BufReader};
 
 // Interact with the composite build system to "seal" the components.
 // This requires linking them with all dependencies, and with libc,
@@ -248,12 +250,58 @@ fn constructor_serialize_args(
     Ok(args_file_path)
 }
 
+fn comp_gen_make_cmd_scan_lib_constant(constants: &[String]) -> Result<Vec<String>, String> {
+    println!("Scanning libraries for constants: {:?}", constants); // Print the constants being scanned for
+    let constants_str = constants.join(" ");
+    let output = Command::new("make")
+        .arg("-C")
+        .arg("src")
+        .arg("--no-print-directory")
+        .arg("scan_libs_for_constants")
+        .env("CONSTANTS", constants_str)
+        .output()
+        .expect("Failed to execute make command");
+    
+    if !output.status.success() {
+        return Err(String::from_utf8_lossy(&output.stderr).to_string());
+    }
+
+    let rebuild_libs_file = File::open("src/components/lib/rebuild_libs.txt").expect("Failed to open rebuild_libs.txt");
+    let reader = BufReader::new(rebuild_libs_file);
+    let libs_to_rebuild: Vec<String> = reader
+        .lines()
+        .map(|line| line.expect("Could not read line"))
+        .collect();
+
+    Ok(libs_to_rebuild)
+}
+
+fn comp_gen_make_cmd_rebuild_lib(libs_to_rebuild: Vec<String>, header_file_path: &str) -> Result<(), String> {
+    println!("Libraries to rebuild: {:?}", libs_to_rebuild); // Print the list of libraries to rebuild
+    for lib in libs_to_rebuild {
+        println!("Rebuilding library: {}", lib);
+        let lib_output = Command::new("make")
+            .arg("-C")
+            .arg("src")
+            .arg("single_lib")
+            .arg(format!("LIBRARY={}", lib))
+            .arg(format!("COMP_CONSTANTS_HEADER={}", header_file_path))
+            .output()
+            .expect("Failed to execute make command for single library");
+
+        if !lib_output.status.success() {
+            return Err(String::from_utf8_lossy(&lib_output.stderr).to_string());
+        }
+    }
+    Ok(())
+}
 fn comp_gen_make_cmd(
     output_name: &String,
     args_file: &String,
     tar_file: &Option<String>,
     id: &ComponentId,
     s: &SystemState,
+    header_file_path: Option<String>,
 ) -> String {
     let c = component(&s, id);
     let ds = deps(&s, id);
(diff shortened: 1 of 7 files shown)


src/composer/src/build.rs
```diff
@@ -248,12 +250,58 @@ fn constructor_serialize_args(
     Ok(args_file_path)
 }
 
+fn comp_gen_make_cmd_scan_lib_constant(constants: &[String]) -> Result<Vec<String>, String> {
+    println!("Scanning libraries for constants: {:?}", constants); // Print the constants being scanned for
+    let constants_str = constants.join(" ");
+    let output = Command::new("make")
```
gparmer, line 256: We're using the `exec_pipeline` in the composer, so I don't see a reason to use two means of constructing pipelines. Either we get rid of the `exec_pipeline`, and use `Command` directly, or we use `exec_pipeline` here.

I might be missing something, and if so, I apologize!

src/composer/src/build.rs
```diff
@@ -248,12 +250,58 @@ fn constructor_serialize_args(
     Ok(args_file_path)
 }
 
+fn comp_gen_make_cmd_scan_lib_constant(constants: &[String]) -> Result<Vec<String>, String> {
+    println!("Scanning libraries for constants: {:?}", constants); // Print the constants being scanned for
```
gparmer, line 254: The composer really focuses on not printing much out, so I don't like this.

We might want a verbose mode, but we don't have that yet. Don't think that the comment adds anything over the print.

src/composer/src/build.rs
```diff
@@ -248,12 +250,58 @@ fn constructor_serialize_args(
...
+        .arg("scan_libs_for_constants")
+        .env("CONSTANTS", constants_str)
+        .output()
+        .expect("Failed to execute make command");
+    
+    if !output.status.success() {
+        return Err(String::from_utf8_lossy(&output.stderr).to_string());
+    }
+
+    let rebuild_libs_file = File::open("src/components/lib/rebuild_libs.txt").expect("Failed to open rebuild_libs.txt");
```
gparmer, line 269: We discussed this previously, but I thought I'd write it here for the record. We certainly do not want to use files to store values that should be passed through data-structures. If you want some input on designing this, I'm happy to help.

Again, I very much could be missing something here, as I can see this file is recording information for all libraries.

I'm not sure if the composer can be executed validly from another directly other than the Composite source root. If so, this hard-coded relative directory is not appropriate. I honestly don't remember if that is an assumption we make elsewhere.

src/composer/src/build.rs
```diff
@@ -284,11 +332,31 @@ fn comp_gen_make_cmd(
...
 
+
     let mut optional_cmds = String::from("");
     optional_cmds.push_str(&format!("COMP_INITARGS_FILE={} ", args_file));
     if let Some(s) = tar_file {
         optional_cmds.push_str(&format!("COMP_TAR_FILE={} ", s));
     }
+
+    if !c.constants.is_empty() {
+        let mut header_content = String::from("#ifndef COMPONENT_CONSTANTS_H\n#define COMPONENT_CONSTANTS_H\n\n");
```
gparmer, line 343: The construction of the file contents should be in the following `if`, right? No reason to do all of this if a header path isn't passed in.

src/composer/src/build.rs
```diff
@@ -413,7 +482,7 @@ impl BuildState for DefaultBuilder {
         let binary = self.comp_obj_path(&c, &s)?;
         let argsfile = constructor_serialize_args(&c, &s, self)?;
         let tarfile = constructor_tarball_create(&c, &s, self)?;
-        let cmd = comp_gen_make_cmd(&binary, &argsfile, &tarfile, &c, &s);
+        let cmd = comp_gen_make_cmd(&binary, &argsfile, &tarfile, &c, &s, None);
```
gparmer, line 485: I would think that we'd want recompile for configured constants for the booter as well.

1. it enables us to be uniform, and 
2. it will enable us to customize things like # of CPUs entirely (eventually) in the configuration script.

src/composer/src/cossystem.rs
```diff
@@ -28,13 +28,20 @@ pub struct InterfaceVariant {
     pub variant: Option<String>,
 }
 
+#[derive(Debug, Deserialize,Clone)]
```
gparmer, line 31: Please make sure to run `rustfmt` on all of this. You should never do a commit without running `rustfmt`.

src/composer/src/cossystem.rs
```diff
@@ -28,13 +28,20 @@ pub struct InterfaceVariant {
...
+}
+
 #[derive(Debug, Deserialize)]
 pub struct TomlComponent {
     name: String,
     img: String,
     baseaddr: Option<String>,
     deps: Option<Vec<Dep>>,
     params: Option<Vec<Parameters>>,
+    constants: Option<Vec<ConstantVal>>, // field for constants
```
gparmer, line 44: Not sure if the comment is adding anything here. If you want it to be useful, add more context that the reader cannot get from the names. "Set of configurable constants used to specialize a component's compilation." for example.

gparmer (commented):
This seems like it is basically untested, so it is hard to know what state it will need to be in eventually.

That said, you're on the right track, I think. Some comments in this review about the code, but the most important comment is what I made in person: lets forget right now about checking if a library uses a constant. Instead, lets just add a specification for the library for it to declare that it uses a constant.

You'll also need to find where the default values of the constants are defined and surround them with 

```c
#ifndef EXAMPLE_CONSTANT
#define EXAMPLE_CONSTANT 16
#endif EXAMPLE_CONSTANT
```

...whereas before it might have just been defined as:

```c
#define EXAMPLE_CONSTANT 16
```

Without this, we'll just be multiply defining constants.

gparmer:
I'm going to close this PR, as you just wanted feedback.

src/composer/src/build.rs
mbai1010: we can use exec_pipeline for sure. But I have faced some issue with using exec_pipeline( the command generated from exec_pipeline has some format issue and can not be recognized by make). I need to find a solution for that.

src/composer/src/build.rs
mbai1010: sounds good, now it is only for my debug purpose, will remove them later

src/composer/src/build.rs
mbai1010: make sense, will change it

PR #481, opened 2024-07-09
Implementation of generating the header file of constants value & Updating the exec_pipeline function

mbai1010:

1. To solve the exec_pipeline issue I faced, I imported rust crate shell_word into pipe.rs. The original design used  cmd.split_whitespace()  to split the make command. In that case, "-include /path/constants.h" will be split into two args. Using shell_words to keep it as a single arg. 
2. Interpret constants values in the TOML file into Rust code.
3. Generate the constants header file.

### Intent for your PR

Choose one (Mandatory):

### Reviewers (Mandatory):

### Code Quality

### Testing

- [ ] micro_booter
- [ ] unit_pingpong
- [ ] unit_schedtests
- [ ] ...(add others here)


src/composer/src/build.rs
```diff
@@ -371,14 +377,46 @@ impl BuildState for DefaultBuilder {
...
+        }
+
+        if !c.constants.is_empty() {    
+            let mut header_content = String::from("#ifndef COMPONENT_CONSTANTS_H\n#define COMPONENT_CONSTANTS_H\n\n");
+            
+            for constant in &c.constants {
+                header_content.push_str(&format!("#define {} {}\n", constant.variable, constant.value));
+            }
+    
+            header_content.push_str("\n#endif // COMPONENT_CONSTANTS_H\n");
```
gparmer, line 400: In C code, we always use /* */

Sigh, sorry for how confusing and frustrating this is!

src/composer/src/build.rs
```diff
@@ -371,14 +377,46 @@ impl BuildState for DefaultBuilder {
...
+    
+    fn comp_const_h(
+        &self, id: &ComponentId, s: &SystemState, header_file: &mut Option<String>
+    ) -> Result<(), String> {
+        let c = component(&s, id);
+        let header_file_path = self.comp_file_path(id, &"component_constants.h".to_string(), &s)?;
+
+        if std::path::Path::new(&header_file_path).exists() {
+            //check the existing header file for constructor_build()
+            *header_file = Some(header_file_path);
```
gparmer, line 389: This is not the way to return multiple values in Rust.

Use the actual return value. You'll change `-> Result<(), String>` to `-> Result<String, String>` with the Ok branch returning the header file name as the `String`.

src/composer/src/build.rs
```diff
@@ -371,14 +377,46 @@ impl BuildState for DefaultBuilder {
...
+    }
 
     fn comp_build(&self, id: &ComponentId, state: &SystemState) -> Result<String, String> {
         let comp_dir = self.comp_dir_path(&id, &state)?;
         compdir_check_build(&comp_dir)?;
         let p = state.get_param_id(&id);
         let output_path = self.comp_obj_path(&id, &state)?;
 
-        let cmd = comp_gen_make_cmd(&output_path, p.param_prog(), p.param_fs(), &id, &state);
+        let mut header_file: Option<String> = None;
```
gparmer, line 416: Given the change above to the return value, you'll be able to remove the `mut` and make this a String, not an option. It can directly get the return value from the function call.

src/composer/src/build.rs
```diff
@@ -254,6 +254,7 @@ fn comp_gen_make_cmd(
     tar_file: &Option<String>,
     id: &ComponentId,
     s: &SystemState,
+    header_file: Option<&String>,
```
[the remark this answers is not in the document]
gparmer: I think you'll find it easier if the header_file is an Option<String> (possible with the changes below).

It is much cleaner not having to do the as_refs and such.

src/composer/src/build.rs
```diff
@@ -371,14 +377,46 @@ impl BuildState for DefaultBuilder {
...
         let comp_dir = self.comp_dir_path(&id, &state)?;
         compdir_check_build(&comp_dir)?;
         let p = state.get_param_id(&id);
         let output_path = self.comp_obj_path(&id, &state)?;
 
-        let cmd = comp_gen_make_cmd(&output_path, p.param_prog(), p.param_fs(), &id, &state);
+        let mut header_file: Option<String> = None;
+        self.comp_const_h(&id, &state, &mut header_file)?;
+
+        let cmd = comp_gen_make_cmd(&output_path, p.param_prog(), p.param_fs(), &id, &state, header_file.as_ref());
```
gparmer, line 419: If you make the return value and other changes for the header file above, the last argument will simply be `Some(header_file)`.

src/composer/src/build.rs
```diff
@@ -413,7 +451,11 @@ impl BuildState for DefaultBuilder {
         let binary = self.comp_obj_path(&c, &s)?;
         let argsfile = constructor_serialize_args(&c, &s, self)?;
         let tarfile = constructor_tarball_create(&c, &s, self)?;
-        let cmd = comp_gen_make_cmd(&binary, &argsfile, &tarfile, &c, &s);
+
+        let mut header_file: Option<String> = None;
```
gparmer, line 455: same as above.

src/composer/src/cossystem.rs
```diff
@@ -28,13 +28,20 @@ pub struct InterfaceVariant {
     pub variant: Option<String>,
 }
 
+#[derive(Debug, Deserialize,Clone)]
```
gparmer, line 31: space before Clone. Did you run rustfmt? You should never do a commit without running that on the project.

gparmer (changes requested):
This is fantastic! Just a few very very small tweaks, mainly around keeping a consistent style. I'm very sorry we don't have great style docs.

src/composer/src/build.rs
```diff
@@ -371,14 +377,46 @@ impl BuildState for DefaultBuilder {
...
+    fn comp_const_h(
+        &self, id: &ComponentId, s: &SystemState, header_file: &mut Option<String>
+    ) -> Result<(), String> {
+        let c = component(&s, id);
+        let header_file_path = self.comp_file_path(id, &"component_constants.h".to_string(), &s)?;
+
+        if std::path::Path::new(&header_file_path).exists() {
+            //check the existing header file for constructor_build()
+            *header_file = Some(header_file_path);
+            return Ok(());
```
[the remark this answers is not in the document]
mbai1010: you are right, this is a potential bug, thank you for pointing that out!  will change it!

src/components/lib/Makefile
```diff
@@ -125,3 +125,18 @@ distclean:
...
+LIB_DIRS := $(shell echo "$(REBUILD_DIRS)" | tr ' ' '\n' | grep '/lib/')
+component_rebuild:
+	$(info **************************************************)
+	$(info **********[ Rebuilding Related Libraries]*********)
+	$(info **************************************************)
+	$(info LIB_DIRS: $(LIB_DIRS))
+	@for dir in $(LIB_DIRS) ; do \
+		echo "*************************[TASK START]*************************";\
+		echo "                    [ Building $$dir ]";\
+		$(MAKE) PLATFORM=$(PLATFORM) $(MAKEOPTIONS) -C $$dir clean || exit $$?; \
```
gparmer, line 138: Ideally a `clean` wouldn't be required, as the dependencies should be properly updated, thus triggering a recompile only of the files that depend on the new header.

But I don't think that our makefiles are bug-free enough to rely on that ;-( Your invocation of `clean` here is likely required, even though it points to deep bugs in the build system.

gparmer (commented):
This looks ready to apply to me. Any reason to wait?

You didn't mark it as ready for merging, but it looks good.

PR #482, opened 2024-07-12
Removed function pointers from heap implementation

esmakokten:

In order to do stack analysis for components, function pointers are removed from heap impl.
Added macro definition 'DECLARE_HEAP' to define heap with given update and compare functions.
Added unit tests to test heap functionality.

### Intent for your PR

Choose one (Mandatory):

### Reviewers (Mandatory):
@gparmer 
@mbai1010 
@spadek67424

### Code Quality

### Testing

- [ ] micro_booter
- [ ] unit_pingpong
- [x] sched_ping_pong
- [x] unit_heap
- [ ] unit_schedtests
- [ ] ...(add others here)


spadek67424:
Hi @esmakokten @gparmer @mbai1010 , From my point, it is good to me. It fixs the function pointer of heap for me. I do not see the function pointer to heap anymore. I am not sure about the file structure and coding style. Maybe somebody could have a look. But I have an off-topic question is that do you miss a commit "7a0fe09d614f5ec7159fc2ecb72aec380172266e" from your branch? Maybe you squash the commit? I do not use PR a lot. Just want to make sure.

src/components/implementation/tests/unit_heap/heap_user.c
```diff
@@ -0,0 +1,155 @@
...
+        assert(!max_heap_highest(h));
+        assert(!max_heap_verify(h, 1));
+    } 
+    assert(heap_size(h) == 0);
+
+    heap_destroy(h);
+    free(es);
+}
+
+#define ITER 10
```
gparmer, line 134: The rest of the code uses tabs. This uses spaces.

Please use tabs. I believe we have an autoformat script lying around that you might apply to this to fix it. Don't remember where, sorry!!!!

src/components/implementation/tmrmgr/simple/tmrmgr.c
```diff
@@ -29,11 +29,25 @@ struct tmr_info {
 };
 
 SS_STATIC_SLAB(timer, struct tmr_info, MAX_NUM_TMR);
-struct heap* timer_active;
```
gparmer, line 32: We only use typedefs and `*_t` names if they are value types. We use `struct blah` as the type when they cannot be passed by value reasonably. See the style guide. This is to make it easier to know how to pass around the type without having to look it up for correctness concerns.

If there was a reason to typedef it, maybe I'm wrong here, but this is not conventional.

src/components/lib/util/heap.h
```diff
@@ -1,34 +1,234 @@
...
+	static inline int NAME##_heap_verify(heap_t *h, int c)                 \
+	{                                                                      \
+		int left, right;                                                   \
+                                                                           \
+		left  = c * 2;                                                     \
+		right = c * 2 + 1;                                                 \
+		if (left < h->e) {                                                 \
+			assert(((hentry_t *)h->data[left])->index == left);            \
+			if (!CMP_FN(h->data[c], h->data[left]) || NAME##_heap_verify(h, left)) {        \
+				printc("Left data %d @ %d < %d @ %d\n", ((hentry_t *)h->data[c])->value, c, \
```
gparmer, line 110: Do you have prints in the code? We shouldn't commit code with debugging prints, if that's what this is (and the one below).

gparmer (changes requested):
Thanks for this! I think that there are a few annoying style and debug things here, so sorry!!!

src/components/implementation/tests/unit_heap/heap_user.c
esmakokten: Oh, I think copy paste lead to this. Thanks, fixed now.

src/components/implementation/tmrmgr/simple/tmrmgr.c
esmakokten: I read the related section in the style guide explains when not to most of the cases, thank you. Removed typedefs.

src/components/lib/util/heap.h
esmakokten: This is the function that the unit test (heap_user.c) uses to verify, it only prints and asserts in case of failure. 
Would it be better to move this under heap_user.c function and hard-code the CMP_FN?

gparmer: Ooops, that's my bad. I forgot the context of these prints.

I'm sorry, they are fine! Ignore my comment.

gparmer (approved):

